PR #2, opened 2025-01-08
updated to 0.7.8

rlmattax:
I couldn't get it to run until I updated this. there might be a better way, but I'm not sure what it is. :)


aaddrick:
Apologies about the late reply, approved. Thanks much!

PR #3, opened 2025-01-13
Fix missing Dock Icon by adding StartupWMClass to .desktop file

drewnolan:
Using Ubuntu 24.02:

1. The default Gear icon was showing up for the Claude Desktop app when running.
image

2. When the app was pinned to the Dock/Dash, launching the app would create a new slot in the Dock/Dash which used the default Gear icon.
image

Setting StartupWMClass in the .desktop file fixes this

ref: https://ubuntuhandbook.org/index.php/2024/04/missing-icon-dock-ubuntu-2404/

### DIFF
--- a/build-deb.sh
+++ b/build-deb.sh
@@ -284,6 +284,7 @@ Type=Application
 Terminal=false
 Categories=Office;Utility;
 MimeType=x-scheme-handler/claude;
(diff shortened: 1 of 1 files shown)


aaddrick:
Thanks for this!

PR #5, opened 2025-01-17
Added location of mcp config file.

freiit:


aaddrick:
Thanks for the callout!

PR #37, opened 2025-03-19
local electron first, nupackage version detection, proper workdir

aar-rafi:
1. Dynamically find the correct nupkg file in the directory
2. Modify the electron check section (around line 73) to prioritize locally installed Electron in node_modules before trying to install it globally
3. The script is trying to access the nupkg file too early, before the previous extraction has completed and created the file. The download and initial extraction of the exe needs to happen before we try to find and process the nupkg file. 

  _issues_:
  1. The WORK_DIR is defined on line 132, but it's used earlier on line 112 before it's defined.
  2. There are multiple directory creation and cd commands that could be consolidated.
  3. The nupkg file detection and version extraction (lines 111-126) happens before the initial exe extraction, but it should be moved after the initial extraction.

  _fix_:
  1. Moved the WORK_DIR definition to the beginning of the script (before its first use)
  2. Moved the nupkg file detection and version extraction section after the first 7z extraction (after line 166)
  3. Consolidated directory creation and cd commands
  5. Ensured proper working directory is maintained throughout the script

The directories are not being created at the right time, which is causing the desktop entry file creation to fail.

_issues_:
1. Initial WORK_DIR creation at lines 121-122
2. Icon directories at line 194
6. electron-app directory at line 204
7. app.asar.contents/resources at lines 256-257
8. claude-native module directory at line 266
9. Various package directories at lines 366-370

_fix_:
moved all the necessary directory creation to the top of the script, right after the directories are defined at line 119. removed the duplicate directory creation commands from later in the script.

### DIFF
--- a/build-deb.sh
+++ b/build-deb.sh
@@ -70,25 +70,48 @@ if [ ! -z "$DEPS_TO_INSTALL" ]; then
     echo "System dependencies installed successfully"
 fi
 
-# Install electron globally via npm if not present
-if ! check_command "electron"; then
+# Check for electron - first local, then global
+# Check for local electron in node_modules
+if [ -f "$(pwd)/node_modules/.bin/electron" ]; then
+    echo "✓ local electron found in node_modules"
+    LOCAL_ELECTRON="$(pwd)/node_modules/.bin/electron"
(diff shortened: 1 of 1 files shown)


aaddrick:
Thanks for the help!

PR #58, opened 2025-04-06
Make build.sh shellcheck happy which is typically a good thing to do

yarikoptic:
Although I did add a few ignores.  Included removal of a few no longer used variables (like all that garbage DNA ;) )


aaddrick:
I'll give it a shot!

PR #62, opened 2025-04-07
chore(readme): update command to build .deb file to not require sudo …

IamGianluca:
…privileges

I just updated the readme file to reflect the correct command to build the `.deb` file. If the user runs the old command, she will be prompted not to use `sudo.` 

BTW, thank you for this project!! :pray:


aaddrick:
Thanks for the catch! Appreciate it!

PR #65, opened 2025-04-07
chore: check for presence of dpkg-deb command, instead of dpkg-dev wh…

IamGianluca:
…ich is a package

The program currently warns that `dpkg-dev` is not installed. I checked, and in my case, I have the package installed. 

I believe the correct behavior should be to check for the presence of the `dpkg-deb` program in the OS, which is installed through the `dpkg-dev` package.


aaddrick:
Thanks again!

PR #76, opened 2025-04-14
bug: it can't type language in ibus, fcitx5 and add text input version Electron argments

ing03201:
i can't type korean ibus, fctix5 ime in ubuntu
there are same problems in [Electron Issue](https://github.com/electron/electron/issues/33662#issuecomment-2299180561)
the argument `--ozone-platform-hint=auto --enable-wayland-ime --wayland-text-input-version=3`  is well done


aaddrick:
Thanks for the help!

PR #137, opened 2025-11-08
Move log from ~/ to ~/.cache/

imaginalnika:
Follows XDG base directory specification


aaddrick:
Thanks @imaginalnika!

PR #143, opened 2025-11-28
Support claude desktop 1.0.1307 with code preview

jacobfrantz1:
This pull request updates the `build.sh` script to support the new versions of claude desktop, adds support for claude code Linux binaries, and updates the handling of the `claude-native` module to use a new package namespace.

It works on my machine. Take or leave any of the changes as you wish. I don't like the large curl command, but was unable to get past cloudflare 403 responses on the redirect endpoint otherwise.

**Installer Download Improvements:**
- Changed the download URLs for the Claude Desktop installer to use the official redirect endpoints from `claude.ai` for both `amd64` and `arm64` architectures, ensuring the latest versions are always fetched.
- Switched from using `wget` to `curl` for downloading the installer, adding detailed HTTP headers to better mimic a browser request and get around cloudflare bot verification.

**Platform Support:**
- Added logic to support installing Linux-specific Claude code binaries, by patching `app.asar.contents/.vite/build/index.js` to recognize both `linux-x64` and `linux-arm64` architectures.

**Module Namespace Updates:**
- Updated the stub implementation of the `claude-native` module for Linux to use the new `@ant/claude-native` namespace, both in the packed and unpacked application directories.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -10,12 +10,12 @@ cat /etc/os-release && uname -m && dpkg --print-architecture
 
 # Set variables based on detected architecture
 if [ "$HOST_ARCH" = "amd64" ]; then
-    CLAUDE_DOWNLOAD_URL="https://claude.ai/api/desktop/win32/x64/exe/latest/redirect"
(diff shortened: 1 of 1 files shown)


aaddrick:
Thanks again!

PR #145, opened 2025-12-03
Update claude download links to newest (1.0.2339)

jacobfrantz1:
This doesn't use the redirect links from the previous pull request, but instead the link behind the redirect it, so I'm hoping we won't have to deal with 403 errors, and it should get us the most up to date exe.

Downside is that the exe will not be updated unless the link is updated. However, the existing google download link refers to 0.14.10, and also doesn't seem to update automatically. (At least not currently)

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -10,12 +10,12 @@ cat /etc/os-release && uname -m && dpkg --print-architecture
 
 # Set variables based on detected architecture
 if [ "$HOST_ARCH" = "amd64" ]; then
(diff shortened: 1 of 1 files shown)


aaddrick:
I see. My bad, didn't realise the link wasn't in use anymore. 

I've hijacked @leobuskin's method using playwrite and am using that to check the version and update the script automatically with the latest direct link via gh action. PR https://github.com/aaddrick/claude-desktop-debian/pull/151

I'm so far behind, this will probably break a bunch of stuff, so I'll start working through it.

Thanks to both of you!

PR #147, opened 2025-12-13
Fix quick window submit issue

### LINKED ISSUE #144: UI Quick Search Bar not working
Hello,

I'm using the AppImage on Kubuntu 24.04 (with Gear Lever), and when clicking on the tray icon (or using the hotkey) the following appears - I can enter some text, but when pressing enter, the window silently disappears. When clicking on "View app" in the tray icon, the app starts normally.

Image

Beside that, I also tried building the DEB on my machine, but although it detects my nvm installation during build, apt still wants to install nodejs/npm and dozens of other node modules. 

```
Running as user: manuel (Home: /home/manuel)
Found NVM installation for user manuel, checking for Node.js 20+...
Adding NVM Node bin path to PATH: /home/manuel/.nvm/versions/node/v24.1.0/bin
System Information:
Distribution: Ubuntu 24.04.3 LTS
Debian version: trixie/sid
Target Architecture: amd64
--- Argument Parsing ---
Selected build format: appimage
Cleanup intermediate files: yes
--- End Argument Parsing ---
Checking dependencies...
✓ p7zip found
✓ wget found
✓ wrestool found
✓ icotool found
✓ convert found
--- Node.js Setup ---
Checking Node.js version...
System Node.js version: v24.1.0
✓ System Node.js version is adequate (v24.1.0)
--- End Node.js Setup ---
```

Why does it first mention "NVM", and then talks about "System Node.js"?

jacobfrantz1:
Before this fix, when submitting a prompt on the quick window, it would simply disappear and the user would have to manually open the main window (right click tray -> show app) to see the response.

This fix gets around an Electron Linux issue where isFocused returns true after hiding the window if blur was not called.

Fixes #144

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -615,6 +615,12 @@ fi
 echo "✓ Tray menu handler patched: function=${TRAY_FUNC}, tray_var=${TRAY_VAR}, check_var=${FIRST_CONST}"
 echo "##############################################################"
 
+# Fix quick window submit issue by adding blur() call before hide()
+if ! grep -q 'e.blur(),e.hide()' app.asar.contents/.vite/build/index.js; then
+    sed -i 's/e.hide()/e.blur(),e.hide()/' app.asar.contents/.vite/build/index.js
+    echo "✓ Added blur() call to fix quick window submit issue"
+fi
+
 
 # Allow claude code installation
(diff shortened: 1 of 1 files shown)


aaddrick:
Thanks again!

PR #230, opened 2026-02-15
Allow passwordless sudo in build script

jarrodcolburn:
This PR updates the build script to support passwordless sudo environments (like containers or CI). It uses `sudo -n true` to check for privileges before falling back to a password prompt.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -129,7 +129,7 @@ check_system_requirements() {
 			echo 'Running as root in CI/container environment (allowed)'
 		else
 			echo 'This script should not be run using sudo or as the root user.' >&2
-			echo 'It will prompt for sudo password when needed for specific actions.' >&2
+			echo 'It will use sudo when needed for specific actions (may prompt for password).' >&2
 			echo 'Please run as a normal user.' >&2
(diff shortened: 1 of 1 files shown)


jarrodcolburn:
I build on my Chromebook. The Linux VM (crostini) and container (penguin) use passwordless sudo.

aaddrick (commented):
Hey! This is a reasonable change.

The logic makes sense. You're checking for passwordless sudo with `sudo -n true` before falling back to `sudo -v`. This helps in container or CI environments where NOPASSWD is configured but EUID isn't 0.

I like the focused scope. The flow is correct: try passwordless first, then fall back to interactive if needed.

One thing I'll point out: `sudo -v` already handles passwordless sudo fine on its own. It succeeds silently when no password is needed. The main behavioral change here is really the "Passwordless sudo detected" message for visibility.

The message change on line 132 is a nice touch. "May prompt for password" is more accurate than "will prompt."

Overall, this is a harmless quality-of-life improvement. It shows thoughtfulness about container environments. I'm good with accepting it.

---

PR #287, opened 2026-03-05
fix(doctor): detect virtiofsd installed outside PATH on Debian

jarrodcolburn:
## Problem

`claude-desktop --doctor` reports `[WARN] virtiofsd: not found` even when virtiofsd is properly installed. The runtime spawn in `cowork-vm-service.js` has the same blind spot and would fail at launch time.

## Root Cause

The Debian `virtiofsd` package **intentionally** installs its binary outside of `$PATH`:

- **`/usr/libexec/virtiofsd`** — primary binary (`/usr/libexec/` is for executables invoked by programs, not users directly, per FHS)
- **`/usr/lib/qemu/virtiofsd`** — symlink for QEMU compatibility

This is not an accident. The package's own vhost-user registration file (`/usr/share/qemu/vhost-user/50-virtiofsd.json`) explicitly hard-codes the binary path so QEMU can locate it:

```json
{
  "description": "virtiofsd vhost-user-fs",
  "type": "fs",
  "binary": "/usr/libexec/virtiofsd",
  ...
}
```

The previous `command -v virtiofsd` check only searches `$PATH`, so it always misses this installation.

Replaces the generic `command -v` loop entry for virtiofsd with a dedicated check that:
1. Tries `$PATH` first (covers distros that do put it in PATH)
2. Falls back to probing `/usr/libexec/virtiofsd` and `/usr/lib/qemu/virtiofsd`
3. Reports the resolved path in the `[PASS]` message for clarity

### `scripts/cowork-vm-service.js`

Adds a `findVirtiofsd()` helper with the same probe order, and guards the virtiofsd spawn behind a found-path check so it fails cleanly rather than erroring on a bad command name.

## Verification

On a system with `virtiofsd` installed via `apt install virtiofsd`:

**Before:** `[WARN] virtiofsd: not found`
**After:** `[PASS] virtiofsd: found (/usr/libexec/virtiofsd)`

---
85% AI / 15% Human
Claude: research, implementation, PR
Human: problem identification, direction, review

### DIFF
--- a/scripts/cowork-vm-service.js
+++ b/scripts/cowork-vm-service.js
@@ -52,6 +52,12 @@ const LOG_FILE = path.join(
     process.env.HOME || '/tmp',
     '.config', 'Claude', 'logs', 'cowork_vm_daemon.log'
 );
+
+/**
+ * Locate virtiofsd binary. Debian installs it to /usr/libexec/virtiofsd which
+ * is outside PATH, so we check well-known locations after checking PATH.
+ * @returns {string|null} Absolute path to virtiofsd, or null if not found.
+ */
 function findVirtiofsd() {
     const candidates = [
         '/usr/libexec/virtiofsd',


sebazic92:
Temporary workaround while waiting for this PR to be merged:

`sudo ln -s /usr/libexec/virtiofsd /usr/local/bin/virtiofsd`

 To revert after updating to a release that includes this PR:

  `sudo rm /usr/local/bin/virtiofsd`

Issue #306, opened 2026-03-19
CoWork KVM: virtiofsd exposes entire $HOME to guest VM — isolate with bind mounts


cbonnissent:
## Problem

The KVM backend in `cowork-vm-service.js` passes `$HOME` as the virtiofsd source directory (`-o source=$HOME`). This exposes the **entire home directory** to the guest VM, including sensitive files like `.ssh/`, `.gnupg/`, `.config/`, browser profiles, etc.

Additionally, the guest sdk-daemon writes files under a `shared/` prefix on the virtiofs mount. Files intended for `$HOME/<subpath>/` end up at `$HOME/shared/<subpath>/` instead. The app then looks for files at the unprefixed path and fails with `INVALID_PATH` or `file is outside allowed folders`.

## Root cause

1. **Security**: virtiofsd is given unrestricted access to `$HOME` — any code running in the VM can read/write any file in the user's home directory.
2. **Path mismatch**: The guest sdk-daemon prepends `shared/` to all paths on the virtiofs mount, but the host app expects files at their original `$HOME/<subpath>` locations.

## Attempted fix: symlinks in isolated virtiofs root

We attempted to replace the `$HOME` source with an isolated `virtiofs-root/` directory containing symlinks to only the explicitly requested paths:

```
virtiofs-root/
├── shared/<subpath> → $HOME/<subpath>   (for guest sdk-daemon prefix)
└── <subpath>        → $HOME/<subpath>   (for direct access)
```

With `--sandbox none` to allow virtiofsd to follow symlinks outside the root.

**This does not work.** virtiofsd exposes symlinks as-is to the guest via the FUSE protocol. The guest kernel sees the symlink target (e.g., `/home/user/.config/Claude/...`) and tries to resolve it in the **guest's** filesystem namespace, where that path does not exist. The `--sandbox none` flag only controls whether the virtiofsd *process* is sandboxed, not how symlinks are presented to the guest.

## Proposed solution: bind mounts in a user namespace

Instead of symlinks, use `mount --bind` inside a user+mount namespace:

```bash
unshare --user --mount --map-root-user -- sh -c '
  mount --bind /home/user/project virtiofs-root/shared/project
  mount --bind /home/user/project virtiofs-root/project
  exec virtiofsd --socket-path=... -o source=virtiofs-root -o cache=auto
'
```

- **No root required**: `unshare --user --mount --map-root-user` creates a user namespace where the current user is mapped to root, enabling `mount --bind`.
[...]

PR #309, opened 2026-03-20
fix: bwrap backend mounts at guest paths and uses minimal sandbox root

ecrevisseMiroir:
## Summary

The BwrapBackend had two compounding issues:

### 1. Security: workDir fell back to $HOME, making entire home writable

The spawn `cwd` arrives as `/sessions/<name>` (a session root without the `/mnt/<mount>` suffix). `resolveWorkDir` could not translate this pattern via `translateGuestPath` and fell back to `os.homedir()`. This caused `--bind $HOME $HOME` to override the preceding `--ro-bind $HOME $HOME`, making the **entire home directory writable** inside the sandbox — defeating the read-only home protection for `~/.ssh`, `~/.gnupg`, etc.

### 2. Broken paths: /sessions/ guest paths don't exist inside bwrap

There are two layers of path translation:

- **cowork-vm-service.js** translates guest paths at spawn time — `cleanSpawnArgs` converts `--add-dir /sessions/.../mnt/project` to `--add-dir /home/user/project`, and `buildSpawnEnv` does the same for `CLAUDE_CONFIG_DIR`.
- **claude-code-vm** (the binary running inside the sandbox) appears to perform its own runtime path translation between host paths and `/sessions/` guest paths.

Since `/sessions/` did not exist inside the bwrap namespace, **all file operations on the working directory failed** regardless of which path form was used.

#### Example: reproducing the issue

When telling cowork to "ls the working directory" with a folder mounted, the agent would attempt:

```
bash
ls /sessions/trusting-jolly-ritchie/mnt/my-project
```

And get:

```
Exit code 2
ls: cannot access '/home/user/my-project': No such file or directory
```

The error shows the host path instead of the guest path. `cowork-vm-service.js` only translates at spawn time (startup args/env), not runtime Bash commands — stdout/stderr from the sandboxed process are piped through unmodified. This indicates `claude-code-vm` itself translates paths at runtime, and always resolves to `/sessions/` guest paths internally.

#### read working directory
Screenshot From 2026-03-20 10-12-25 (Edit)

#### Create a file (Cowork decided to create it in `~/mnt` as a fallback after having a lack of permission. This is unsafe and moreover the desktop app does not allow its visualization because it's outside allowed folders).
Screenshot From 2026-03-20 10-15-20 (Edit)

### Fix

Replaces the `--ro-bind / /` full-root approach with a minimal tmpfs-based sandbox:

- `--tmpfs /` as an empty root (nothing exposed by default)
- `/usr`, `/etc`, `/dev`, `/proc` bound read-only for system tools
- System directories (`/bin`, `/lib`, `/lib64`, `/sbin`) are detected at runtime: symlinks on merged-usr distros, separate bind mounts on others — works across Fedora, Debian, Ubuntu, Arch, etc.
- SDK binary directory bound read-only
- `/sessions/<name>/mnt/` created via `--dir`, with host directories bind-mounted at guest paths (matching the KVM backend layout)
- `--chdir` sets the cwd inside the sandbox to the guest working directory
- Guest-to-host path translation in cowork-vm-service is skipped entirely for bwrap (args and env passed through as-is), since the guest paths now exist natively inside the sandbox

##### Implicitly mentioning the working dir
Screenshot From 2026-03-20 10-23-20 (Edit) (Edit)

#### Creating a file in the work dir and _outputting_ it:
Screenshot From 2026-03-20 10-31-40 (Edit)

## Test plan

- [x] Verified `bwrap --tmpfs / ...` approach works with bind mounts at guest paths
- [x] Confirmed `--chdir /sessions/<name>/mnt/<mount>` sets correct cwd
- [x] Tested `ls`, file creation, and `pwd` all work inside the sandbox
- [x] Verified home directory is not exposed (only an empty `--dir`)
- [ ] Test with plugins that use `--plugin-dir`
- [ ] Test on non-merged-usr distro (e.g., older Debian)

---
Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>
90% AI / 10% Human
Claude: Investigated the root cause across path translation, bwrap mount setup, and claude-code-vm bidirectional path mapping; implemented the fix
Human: Identified the security concern, guided debugging, caught hardcoded paths, and validated the fix on a live cowork session

### DIFF
--- a/scripts/cowork-vm-service.js
+++ b/scripts/cowork-vm-service.js
@@ -581,6 +581,7 @@ class LocalBackend extends BackendBase {
 
         return {
             id,
+            name,
             actualCommand: resolved.command,
             cleanArgs: cleanSpawnArgs(args || [], mountMap),
             mergedEnv: buildSpawnEnv(env, mountMap),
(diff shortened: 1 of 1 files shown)


cbonnissent:
Thank you very much - I have tested and that fix the bubbleWrap implementation and my issue with the access of my home directory and the access to the artifact #306 

It's more light than have a VM that boot (but the isolation is not as strong) 
It's definitively useful !

PR #310, opened 2026-03-20
feat: add SHA-256 checksum validation for downloads

sabiut:
## Summary

- Add `verify_sha256()` helper to `build.sh` for reusable SHA-256 integrity checks
- Verify Claude Desktop EXE after download (skipped for local `--exe` builds)
- Verify Node.js tarball against official `SHASUMS256.txt` from nodejs.org
- Update `check-claude-version.yml` to compute and store hex SHA-256 hashes alongside
  existing SRI hashes, so checksums auto-update with version bumps

Hash placeholders are empty until the next CI version update populates them.
The verify function gracefully warns and skips when no hash is available.
appimagetool is excluded (rolling "continuous" release, checksums impractical).

## Test plan

- [x] `verify_sha256()` tested: empty hash skips, correct hash passes, wrong hash fails
- [x] `bash -n build.sh` syntax check passes
- [x] `--test-flags` mode still works
- [x] shellcheck passes clean
- [ ] Full build with download (requires CI run)

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -73,8 +73,8 @@ verify_sha256() {
 	fi
 
 	echo "Verifying SHA-256 checksum for ${label}..."
-	local actual_hash
-	actual_hash=$(sha256sum "$file_path" | awk '{print $1}')
+	local actual_hash _
+	read -r actual_hash _ < <(sha256sum "$file_path")
 
 	if [[ $actual_hash != "$expected_hash" ]]; then
 		echo "SHA-256 mismatch for ${label}!" >&2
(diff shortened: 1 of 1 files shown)


aaddrick (approved):
## Review

Good feature — addresses a real supply-chain integrity gap. Clean implementation, follows project style, and the graceful empty-hash skip means this merges safely before CI populates the values.

### Suggestions for a follow-up

**1. Hardcode the Node.js hash instead of fetching SHASUMS256.txt at runtime**

The current approach fetches `SHASUMS256.txt` from nodejs.org over HTTPS, but an attacker who can MITM the download can also MITM the checksum file — verification becomes circular. Since the Node.js version is pinned at `20.18.1` and rarely changes, hardcoding the hash (like the Claude EXE approach) is simpler and more secure. Two hashes (amd64/arm64) to maintain, updated only when bumping the Node version.

This also eliminates the `|| true` that silently swallows fetch failures, which currently degrades to no verification with only a stderr warning.

**2. Minor: log a note when `--exe` skips verification**

The local exe path bypasses verification silently. A brief `echo "Note: SHA-256 verification skipped for local installer"` would improve auditability.

Neither of these block the merge — the PR is a clear improvement over the status quo (no verification at all). The Node.js hardcoding can be a follow-up.

---

PR #320, opened 2026-03-20
fix: disable GPU compositing for XRDP/VNC sessions + nightly update scripts

### LINKED ISSUE #319: fix: Claude Desktop blank/crashed window when launched via XRDP or VNC
## Problem

When Claude Desktop is launched inside an **XRDP or VNC remote-desktop session**, the app either:
- Renders a completely blank white window, or
- Crashes at startup with a GPU-related error in `~/.config/Claude/logs/`

This happens because Electron's GPU-accelerated compositor path requires real GPU access, which XRDP and most VNC servers do not provide — they expose a software framebuffer only.

## Environment

- Display server: XRDP (reproduces with `xrdp-sesman` running, or `$XRDP_SESSION` set)
- Also affects: VNC sessions, headless X servers without GPU passthrough

## Fix

Detect remote-desktop sessions in `scripts/launcher-common.sh` and pass
`--disable-gpu --disable-software-rasterizer` to Electron, which forces the
CPU compositor path and renders correctly in all software-only framebuffer environments.

Detection logic (two signals, either is sufficient):
1. `$XRDP_SESSION` environment variable (set by the XRDP X server)
2. `xrdp-sesman` process running (catches VNC-via-xrdp and cases where the env var isn't propagated)

A PR with this fix is attached.

---
Written by Claude Sonnet 4.6 via [Claude Code](https://claude.ai/code)

davidamacey:
## Summary

Fixes #319

Two changes in this PR:

### 1. XRDP / VNC GPU fix

When Claude Desktop runs inside an XRDP or VNC session, Electron's GPU compositor fails — the window either renders blank or crashes at startup. This adds detection in `scripts/launcher-common.sh` to disable GPU compositing for those sessions:

```bash
# Detect remote desktop sessions that lack GPU acceleration (XRDP, VNC)
if [[ -n "${XRDP_SESSION:-}" ]] || pgrep -x xrdp-sesman &>/dev/null; then
    electron_args+=('--disable-gpu' '--disable-software-rasterizer')
    log_message 'Remote desktop detected (XRDP) - GPU compositing disabled'
fi
```

Detection uses two signals (either is sufficient):
- `$XRDP_SESSION` — set by the XRDP X server
- `xrdp-sesman` process probe — catches VNC-via-xrdp and sessions where the env var isn't propagated

### 2. Nightly auto-update scripts

Three optional helper scripts for keeping Claude Desktop and Claude Code CLI up to date automatically:

| Script | Purpose |
|---|---|
| `scripts/nightly-update.sh` | Cron orchestrator — runs both updaters, logs results |
| `scripts/auto-update-desktop.sh` | Checks version, builds .deb, installs via apt |
| `scripts/auto-update-claude-code.sh` | Runs `claude update`, skips if sessions active |

**Key design points:**
- All paths resolved from script location — no hardcoded absolute paths
- `--dry-run` flag: show what would happen without changes
- `--force` flag: bypass safety checks (running-process detection)
- Logs to `$HOME/.local/log/claude-updates/`
- NVM sourced automatically for cron's minimal PATH
- Optional [ntfy](https://ntfy.sh/) notifications (silently skipped if no server reachable) — configure via `NTFY_URL` / `NTFY_TOPIC`

Suggested cron:
```
7 3 * * * /path/to/claude-desktop-debian/scripts/nightly-update.sh
```

---
95% AI / 5% Human
Claude: authored both changes, generalized paths, wrote issue + PR
Human: identified the XRDP problem and requested the PR

### DIFF
--- a/scripts/auto-update-claude-code.sh
+++ b/scripts/auto-update-claude-code.sh
@@ -1,0 +1,167 @@
+#!/usr/bin/env bash
+
+# Auto-update Claude Code CLI via `claude update`
+# Safe: skips update if active Claude Code sessions are detected.
+#
+# Usage: auto-update-claude-code.sh [--force] [--dry-run]
+#   --force    Skip running-session safety checks
+#   --dry-run  Show what would happen; make no changes
+#
+# Optional ntfy notifications (silently skipped if server unreachable):
+#   NTFY_URL    ntfy server base URL  (default: http://localhost:2586)
+#   NTFY_TOPIC  ntfy topic to publish (default: claude-updates)
+
+log_dir="${HOME}/.local/log/claude-updates"
+log_file="$log_dir/claude-code.log"
+
+dry_run=false
+force=false
+
+# ------------------------------------------------------------------
+# Helpers
+# ------------------------------------------------------------------
+
+log() {
+	local msg
+	msg="[$(date '+%Y-%m-%d %H:%M:%S')] $*"
+	echo "$msg"
+	echo "$msg" >> "$log_file"
+}
+
+notify() {
+	local title="$1"
+	local body="$2"
+	local priority="${3:-default}"
+	local url="${NTFY_URL:-http://localhost:2586}"
+	local topic="${NTFY_TOPIC:-claude-updates}"
+	curl -s --max-time 5 \
+		-H "Title: $title" \
+		-H "Priority: $priority" \
+		-d "$body" \
+		"$url/$topic" > /dev/null 2>&1 || true
+}
+
+# Return 0 (true) if active Claude Code sessions appear to be running.
+# We look for node processes running the claude CLI, excluding:
+#   - claude-desktop (Electron app)
+#   - auto-update scripts themselves
+#   - this script's own process tree
+is_claude_code_active() {
+	local our_pid=$$
+	local found=false
+
+	while IFS= read -r pid; do
+		[[ -z $pid ]] && continue
+		(( pid == our_pid )) && continue
+
+		local args
+		args=$(ps -p "$pid" -o args= 2>/dev/null) || continue
+
+		# Skip claude-desktop (Electron / .mount_claude*)
+		[[ $args == *claude-desktop* ]] && continue
+		[[ $args == *mount_claude* ]] && continue
+		# Skip auto-update processes
+		[[ $args == *auto-update* ]] && continue
+		# Skip grep/pgrep calls
+		[[ $args == *pgrep* || $args == *grep* ]] && continue
+
+		found=true
+		break
+	done < <(pgrep -f 'claude' 2>/dev/null)
+
+	$found
+}
+
+# Return the current `claude --version` string.
+claude_version() {
+	claude --version 2>/dev/null || true
+}
+
+# ------------------------------------------------------------------
+# Argument parsing
+# ------------------------------------------------------------------
+
+parse_args() {
+	while (( $# > 0 )); do
+		case "$1" in
+			--force)   force=true;   shift ;;
+			--dry-run) dry_run=true; shift ;;
+			-h|--help)
+				echo 'Usage: auto-update-claude-code.sh [--force] [--dry-run]'
+				exit 0
+				;;
+			*)
+				echo "Unknown option: $1" >&2
+				exit 1
+				;;
+		esac
+	done
+}
+
+# ------------------------------------------------------------------
+# Main
+# ------------------------------------------------------------------
+
+main() {
+	parse_args "$@"
+	mkdir -p "$log_dir"
+
+	log '=== Claude Code update check starting ==='
+	$dry_run && log '[DRY RUN mode]'
+
+	# Verify claude is available
+	if ! command -v claude > /dev/null 2>&1; then
+		log 'ERROR: claude command not found in PATH'
+		return 1
+	fi
+
+	local version_before
+	version_before=$(claude_version)
+	log "Current version: $version_before"
+
+	# --- Safety check ---
+	if ! $force; then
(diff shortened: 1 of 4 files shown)


aaddrick (commented):
Hey! Thanks for putting this together, @davidamacey. The XRDP fix addresses a real usability gap for remote desktop users, and the auto-update scripts show solid defensive design with dry-run, force flags, and graceful ntfy handling. Good stuff overall.

Before diving in, I want to flag a scope concern. This PR bundles two unrelated features: a 6-line launcher fix and ~550 lines of new update automation. The commit history already separates them cleanly. I'd recommend splitting this into two PRs — the XRDP fix is small and merge-ready with one tweak, while the update scripts need a larger revision pass.

---

### \`scripts/launcher-common.sh\`

**Issue 1: \`pgrep -x xrdp-sesman\` false-positives on local sessions**

The \`xrdp-sesman\` process is a system service. It runs whenever xrdp is *installed*, even when the current user is sitting at a local desktop. This will incorrectly disable GPU compositing for local sessions on any machine that has xrdp available.

The \`\$XRDP_SESSION\` environment variable is session-scoped and sufficient on its own. Process scanning doesn't add reliable signal here.

Recommendation:

\`\`\`bash
# Detect remote desktop sessions that lack GPU acceleration
if [[ -n "\${XRDP_SESSION:-}" ]]; then
    electron_args+=('--disable-gpu' '--disable-software-rasterizer')
    log_message 'XRDP session detected - GPU compositing disabled'
fi
\`\`\`

If you want broader VNC coverage, check for VNC-specific env vars like \`\$VNCDESKTOP\` rather than process probing.

---

### \`scripts/auto-update-desktop.sh\`

**Issue 2: Sudoers recommendation is overly permissive**

The comment suggests \`NOPASSWD: /usr/bin/apt install *\` which allows passwordless root installation of *any* package, not just claude-desktop. A compromised user account could install arbitrary \`.deb\` files with pre/post-install scripts running as root. Restrict to the specific package path.

**Issue 3: \`is_cowork_active()\` matches all bubblewrap processes**

\`bwrap\` is used by Flatpak, Firejail, and other sandboxing tools. On any system running Flatpak apps, this will almost always return true, preventing updates even when Claude isn't involved. Should match \`cowork-vm-service\\.js\` specifically.

**Issue 4: \`tee\` pipeline masks git exit status**

\`if ! git ... | tee -a "$log_file"\` tests \`tee\`'s exit status, not \`git\`'s. \`tee\` almost never fails, so git failures go undetected.

**Issue 5: \`git checkout main\` in a working repo is surprising for cron**

An unattended cron job silently switching branches could discard context or build from the wrong state. Consider using \`git worktree\` or building directly from the fetched remote ref without modifying the working tree.

**Issue 6: \`repo_dir\` resolution doesn't guard against \`cd\` failure**

---

### \`scripts/auto-update-claude-code.sh\`

**Issue 7: \`is_claude_code_active()\` uses fragile exclusion-based matching**

Matches everything with "claude" in the cmdline, then tries to exclude known false positives. Should match what you *want* (e.g., \`@anthropic-ai/claude-code\`) rather than excluding what you don't.

---

### All three new scripts

**Issue 8: \`log()\`, \`notify()\`, and \`parse_args()\` are copy-pasted three times** — extract shared helpers into \`scripts/update-common.sh\`, following the existing \`launcher-common.sh\` pattern.

**Issue 9:** \`> /dev/null 2>&1\` instead of \`&>/dev/null\` — inconsistent with codebase convention.

**Issue 10:** Double quotes used where single quotes suffice per STYLEGUIDE.md.

---

### Summary

The XRDP detection fix (Issue 1) is close — narrowing to the env var check only would make it solid. I'd recommend splitting that out as its own PR.

The update scripts need more work before they're ready:

1. **Security**: Tighten the sudoers recommendation (Issue 2)
2. **Correctness**: Fix \`bwrap\` false positives (Issue 3), \`tee\` masking (Issue 4), and Claude Code detection (Issue 7)
3. **Safety**: Rethink \`git checkout\` in a working repo (Issue 5)
4. **Maintainability**: Extract duplicated helpers (Issue 8)
5. **Style**: Minor quoting and redirection form fixes (Issues 9-10)

The defensive design choices (dry-run, force, graceful ntfy fallback) are well thought out. These are solid foundations that will be even better after addressing the issues above.

---

aaddrick:
Hey @davidamacey — thanks again for the diagnosis on #319 and the fix here. The XRDP detection in `launcher-common.sh` is the right shape and solves a real bug, so I want to get that part landed.

Rather than keep this PR open while the auto-update side gets untangled from the XRDP fix, I'm going to close it and open a new PR that picks up just the 6-line XRDP/VNC change on top of current main. You'll be credited as `Co-Authored-By` on the commit and in the PR body.

On the auto-update scripts: I've been going back and forth on this, and I've landed on not taking the project in that direction. The main reason is that we already ship APT, DNF, and AUR repositories, so for users on Debian, Ubuntu, Fedora, and Arch, auto-updates are the package manager's job and they already work. AppImage is the one real gap, but the right answer there is `AppImageUpdate` with embedded zsync info, not a cron path that shells out to sudo and rebuilds from a git clone. My [earlier review](https://github.com/aaddrick/claude-desktop-debian/pull/320#pullrequestreview-3989054965) has the specific concerns if you're curious, but I don't want you to spend more time on it expecting it to land.

I know that's real work you put in and I appreciate it — the diagnosis on #319 alone is worth a lot.

---

aaddrick:
Superseded by #475, which lands the XRDP fix on current main with the detection hardened (loginctl session Type instead of pgrep xrdp-sesman). @davidamacey is credited as `Co-Authored-By` on the commit and in the README Acknowledgments.

Issue #322, opened 2026-03-21
KvmBackend: guest readiness timeout — VM boots but sdk-daemon never connects via vsock


hareendranmg:
## Description

After upgrading to `v1.3.19` (which fixed #315), the Cowork workspace still fails to start on my system. The error has shifted from "Request timed out: configure" to **"Request timed out: isGuestConnected"** / **"VM connection timeout after 60 seconds"**.

The KVM backend is auto-selected (all prerequisites met), QEMU launches successfully, and the guest VM boots (overlay disk grows, QEMU uses ~1.5GB RAM), but the guest `sdk-daemon` never establishes the vsock connection back to the host. The socat vsock listener on port 51234 never receives a connection.

## Environment

- **OS:** Ubuntu 22.04.5 LTS (Jammy Jellyfish)
- **Kernel:** 6.8.0-59-generic
- **CPU:** 13th Gen Intel Core i7-13700K (24 cores)
- **RAM:** 31.10 GB
- **Claude Desktop:** 1.1.7714-1.3.19
- **QEMU:** 6.2.0 (Debian 1:6.2+dfsg-2ubuntu6.28)
- **socat:** 1.7.4.1
- **KVM:** `/dev/kvm` present, `kvm-ok` passes
- **vhost-vsock:** `/dev/vhost-vsock` present, `vhost_vsock` module loaded

## Steps to Reproduce

1. Install `claude-desktop` v1.3.19
2. Ensure KVM prerequisites are met (`/dev/kvm`, `qemu-system-x86_64`, `/dev/vhost-vsock`)
3. Launch `claude-desktop`
4. Open the Cowork tab
5. VM starts but workspace fails with "Failed to start Claude's workspace — Request timed out: isGuestConnected"

### `cowork_vm_node.log`
```
[info] [VM:start] Copying smol-bin.x64.vhdx to bundle (Linux)
[info] [VM:start] smol-bin.x64.vhdx copied successfully
[info] [VM:steps] prepare_session completed (0ms)
[info] [VM:steps] start_vm completed (0ms)
[info] [VM:steps] wait_for_guest completed (0ms)
[info] [VM:start] Still waiting for guest connection... 10020ms elapsed, 21 polls
[info] [VM:start] Still waiting for guest connection... 20041ms elapsed, 41 polls
[info] [VM:start] Still waiting for guest connection... 30064ms elapsed, 61 polls
[...]

Issue #311, opened 2026-03-20
Claude Desktop 1.1.7464 hangs on startup after PR #305 (NixOS regression)


arauhala:
### Summary
Claude Desktop stopped starting after the flake update that included PR #305. 
The process spawns but hangs silently with no window and no output.

### Environment
- OS: NixOS 25.05
- Hardware: Dell XPS 15-9520, NVIDIA RTX 4060 Max-Q
- Installation: aaddrick/claude-desktop-debian NixOS flake

### Versions
- **Working:** 1.1.6041 (commit `3cfb0b91`)
- **Broken:** 1.1.7464 (after PR #305)

### Symptoms
- 3 processes spawn (2 electron + 1 claude-desktop)
- No window appears
- No output to stdout/stderr
- No new log entries written to `~/.config/Claude/logs/main.log`
- Hangs indefinitely until killed

### Key observation
Both working and broken are Claude Desktop **1.1.7464**, but differ in packaging:

**Working (commit 3cfb0b91):**
- `isPackaged: false`
- `[Frame Fix]` wrapper present, explicitly corrects `process.resourcesPath`:
  - Was: `.../electron-unwrapped-37.10.2/libexec/electron/resources`
  - Now: `.../claude-desktop-1.1.7464/lib/claude-desktop/resources`

**Broken (after PR #305):**
- `isPackaged: true`
- No `[Frame Fix]` output — wrapper missing or not running
- Electron resolves resources from wrong path → silent hang, no window

### Root cause hypothesis
PR #305 changed packaging so `isPackaged=true`, bypassing or removing the 
`process.resourcesPath` fix that is required for NixOS to work correctly.

### Workaround
Pin flake to last working commit:
```nix
claude-desktop.url = "github:aaddrick/claude-desktop-debian/3cfb0b91cc755a3ddb0413ed5e1bba08a625bf1e";
```

### Related
- PR #305 introduced NixOS-specific changes that appear to have caused this regression

Issue #282, opened 2026-03-04
--doctor flag doesn't work!


roeybenarieh:
I got the error "VM service not running. The service failed to start." and I wanted to diagnose the problem.
When running "claude-desktop --doctor" I don't get any doctor output.

I am running NixOS 26.05.
Tried installing both `pkgs.claude-desktop-fhs` and `pkgs.claude-desktop` and in both the "--doctor" doesn't work.

## Screenshots
running the command:
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/b7ff5994-e2b4-48c6-a528-4d908b49bf1a" />
viewed error:
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/0a1eb0ef-3cdf-49d1-a505-2a3f10e0963d" />

Appreciate your help :-)

asheriif:
`--doctor` flag doesn't work for me either with the RPM version. Same for `CLAUDE_USE_WAYLAND=1` environment variable actually, also doesn't get picked up.

PR #328, opened 2026-03-21
fix: enable cowork mode on NixOS (isPackaged + resourcesPath + service pre-launch)

SolAstrius:
On NixOS, cowork mode fails with "VM service not running" immediately
after launch.  Three independent problems combine to cause this:

1. ELECTRON_FORCE_IS_PACKAGED is skipped for NixOS (c3cc5a6), keeping
   app.isPackaged false.  The upstream app gates cowork's VM service
   fork behind isPackaged === true, so the service is never spawned.

2. Re-enabling ELECTRON_FORCE_IS_PACKAGED causes a different failure:
   Electron resolves process.resourcesPath from its real binary path
   (following symlinks), which on NixOS points to Electron's store
   path (/nix/store/...-electron-38/libexec/electron/resources/),
   not the app's resources.  The app then fails to find en-US.json,
   smol-bin.x64.vhdx, and other resources at that path.

   Electron's C++-level resourcesPath is cached before any JS runs
   and cannot be overridden via NODE_OPTIONS (blocked for packaged
   apps) or --require.  The frame-fix-wrapper.js corrects
   process.resourcesPath at JS level, but Electron's internal
   patched fs (node:electron/js2c/node_init) still uses the
   original cached value.

3. The app's auto-launcher for cowork-vm-service.js only triggers on
   ECONNREFUSED, not ENOENT.  On a fresh install where no socket has
   ever been created, the connect() returns ENOENT and the fork path
   is never reached.

Fix all three:

  launcher-common.sh:
  - Re-enable ELECTRON_FORCE_IS_PACKAGED for all package types.

  nix/claude-desktop.nix (launcher script):
  - Create a merged resources directory at runtime
    ($XDG_CACHE_HOME/claude-desktop-debian/resources/) containing
    symlinks to both Electron's and the app's resources.
  - Export CLAUDE_RESOURCES_PATH pointing to the merged directory.
  - Pre-launch cowork-vm-service.js before exec'ing Electron so the
    Unix socket exists when the app tries to connect.

  frame-fix-wrapper.js:
  - Read CLAUDE_RESOURCES_PATH and use it as the corrected
    process.resourcesPath (falls back to path.dirname(__dirname)
    for non-NixOS builds — no behavior change).
  - Monkey-patch fs.readFileSync, existsSync, statSync, and other
    fs methods to redirect any access from Electron's original
    resourcesPath to the merged directory.  This catches lookups
    that go through Electron's internal patched fs bindings before
    the JS-level process.resourcesPath correction takes effect.

Tested on NixOS 25.11 with Electron 38.8.1.  The KVM backend
boots successfully, the guest VM connects via vsock, and the SDK
installs.  The remaining failure (sandbox-helper: /usr/local/bin/
claude: no such file or directory) is inside Anthropic's guest VM
image and is tracked separately in #322.

Relates: #282, #311
Relates: #322 (guest-side /usr/local/bin/claude)
Relates: #237 (alternative pre-launch approach via systemd)

### DIFF
--- a/nix/claude-desktop.nix
+++ b/nix/claude-desktop.nix
@@ -157,6 +157,8 @@ stdenvNoCC.mkDerivation {
 # Claude Desktop launcher for NixOS
 
 electron_exec="ELECTRON_PLACEHOLDER"
+electron_resources="ELECTRON_RESOURCES_PLACEHOLDER"
+app_resources="RESOURCES_PLACEHOLDER"
 app_path="RESOURCES_PLACEHOLDER/app.asar"
 
 source "LAUNCHER_LIB_PLACEHOLDER"
@@ -172,6 +174,22 @@ setup_logging || exit 1
 setup_electron_env 'nix'
 cleanup_stale_lock
 cleanup_stale_cowork_socket
+
+# On NixOS, Electron and the app live in separate store paths.
+# Electron hardcodes process.resourcesPath from its binary location,
+# which can't be overridden. Create a merged resources directory
+# with symlinks to both so all file lookups succeed.
+merged_resources="$log_dir/resources"
+mkdir -p "$merged_resources"
+# Symlink Electron's resources (default_app.asar, etc.)
+for f in "$electron_resources"/*; do
+	ln -sf "$f" "$merged_resources/" 2>/dev/null
+done
+# Symlink app resources (overrides Electron's on conflict)
+for f in "$app_resources"/*; do
+	ln -sf "$f" "$merged_resources/" 2>/dev/null
+done
+export CLAUDE_RESOURCES_PATH="$merged_resources"
 
 # Log startup info
 log_message '--- Claude Desktop Launcher Start (NixOS) ---'
@@ -195,6 +213,18 @@ build_electron_args 'nix'
 # Add app path
 electron_args+=("$app_path")
 
+# Pre-launch cowork-vm-service daemon if not already running.
+# The app's auto-launcher only triggers on ECONNREFUSED, not
+# ENOENT, so without a pre-existing socket the service never
+# starts on a fresh install.
+cowork_svc="RESOURCES_PLACEHOLDER/app.asar.unpacked/cowork-vm-service.js"
+cowork_sock="''${XDG_RUNTIME_DIR:-/tmp}/cowork-vm-service.sock"
+if [[ -f "$cowork_svc" && ! -S "$cowork_sock" ]]; then
+	ELECTRON_RUN_AS_NODE=1 "$electron_exec" "$cowork_svc" &
+	disown
+	log_message "Pre-launched cowork-vm-service (PID $!)"
+fi
+
 # Execute Electron
 log_message "Executing: $electron_exec ''${electron_args[*]} $*"
 "$electron_exec" "''${electron_args[@]}" "$@" >> "$log_file" 2>&1
@@ -204,6 +234,7 @@ exit $exit_code
 LAUNCHER
     # Substitute placeholders with Nix store paths
     substituteInPlace $out/bin/claude-desktop \
+      --replace-fail "ELECTRON_RESOURCES_PLACEHOLDER" "${electron}/libexec/electron/resources" \
       --replace-fail "ELECTRON_PLACEHOLDER" "${electron}/bin/electron" \
       --replace-fail "RESOURCES_PLACEHOLDER" "$out/lib/claude-desktop/resources" \
       --replace-fail "LAUNCHER_LIB_PLACEHOLDER" "$out/lib/claude-desktop/launcher-common.sh"
--- a/scripts/frame-fix-wrapper.js
+++ b/scripts/frame-fix-wrapper.js
@@ -8,14 +8,46 @@ console.log('[Frame Fix] Wrapper loaded');
 // Fix process.resourcesPath to match the actual location of app.asar.
 // In Nix builds, electron is a separate store path so process.resourcesPath
 // points to the Electron package's resources dir, not where our tray icons
-// and app.asar.unpacked live. Deriving from __dirname (the asar root) gives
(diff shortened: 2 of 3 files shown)


aaddrick (commented):
Hey! This is solid work tackling a real problem. The merged-resources symlink approach is architecturally sound — it solves the fundamental NixOS split-package issue at the filesystem level rather than fighting it from JS alone. The PR description is excellent, clearly documenting all three problems and how each fix addresses them.

I've found several issues that range from "should fix before merge" to "worth considering." Organized by file below. I've also included an alternative approach at the end that might simplify things significantly.

---

### `scripts/frame-fix-wrapper.js`

**Issue 1: `copyFileSync`/`copyFile` destination path not redirected (Medium)**

These fs methods take two path arguments. The wrapper only redirects the first (source). If the app ever copies *to* a path under `originalResourcesPath`, the destination won't be redirected. Resources are typically read-only, so this is unlikely to break things today, but it's a latent bug.

**Issue 2: `fs.promises` not patched (Medium)**

The wrapper patches `fs.readFile` etc. but not `fs.promises.readFile`. If any app code uses the promises API, the redirect is bypassed.

**Issue 3: `realpathSync` and `realpath` missing from method list (Low)**

Given the NixOS launcher creates a symlink tree, `realpath` is particularly relevant. If code resolves a path under `originalResourcesPath` before the redirect fires, the lookup would fail.

**Issue 4: Buffer/URL path arguments silently bypass redirect (Low)**

Node's fs APIs accept `Buffer` and `URL` objects as paths. The `typeof p === 'string'` guard prevents crashes, which is good defensive coding. Worth documenting with a brief comment but not blocking.

---

### `nix/claude-desktop.nix`

**Issue 5: Merged resources directory accumulates stale symlinks (Medium)**

The merged dir at `$log_dir/resources` persists across launches. After a `nixos-rebuild` that changes Electron or app store paths, old symlinks point to garbage-collected paths. Recommendation: `rm -rf "$merged_resources" && mkdir -p "$merged_resources"` each launch. It's just symlinks so this is cheap.

**Issue 6: Empty glob expansion creates broken symlink named `*` (Medium)**

If the glob matches nothing, bash expands it literally. Add a guard inside the loop:
```bash
for f in "$electron_resources"/*; do
	[[ -e "$f" ]] || continue
	ln -sf "$f" "$merged_resources/" 2>/dev/null
done
```

**Issue 7: Cowork daemon stdout/stderr not redirected (Medium)**

The background daemon's output isn't redirected, which means if the terminal closes, the daemon could get SIGPIPE. The daemon uses file-based logging, so redirecting to `/dev/null` should be safe:
```bash
ELECTRON_RUN_AS_NODE=1 "$electron_exec" "$cowork_svc" \
	>/dev/null 2>&1 &
disown
```

**Issue 8: Electron resources path `libexec` may vary (Low)**

The substitution hardcodes `${electron}/libexec/electron/resources`. The Nixpkgs `electron` package structure can vary between versions. Worth confirming which electron versions this path is valid for.

**Issue 9: Socket path must match `cowork-vm-service.js` (Low)**

The socket path needs to match what the service creates. Worth verifying against the service's source to prevent redundant daemon spawns.

---

**Issue 10: Comment slightly misattributes why the change is safe (Low)**

The comment says frame-fix-wrapper.js "corrects process.resourcesPath at JS load time, so this is safe for all package types." The actual safety on NixOS comes from three things working together: the merged resources directory, the fs monkey-patching, *and* the load-time correction. The comment should reference the full picture.

---

### What's done well

- The merged-symlink directory approach is the right architecture
- Override order is correct — app resources linked after Electron's, so `ln -sf` gives app resources priority
- The `-f "$cowork_svc"` guard handles non-cowork builds gracefully
- The `typeof p === 'string'` guard in `_redirect` prevents crashes on fd numbers
- On deb/AppImage builds, the `if (derivedResourcesPath !== originalResourcesPath)` check means zero overhead
- The PR description is thorough

---

### Alternative approach worth considering

Issues 1-6 and 8 all stem from doing the resources merge at runtime and patching fs to compensate. There's a potentially simpler path: **do the merge at build time in the Nix derivation.**

The derivation already controls the final package layout. During `installPhase`, it could create a single `resources/` directory with both Electron's and the app's resources symlinked together. Then `process.resourcesPath` points to one directory that has everything — no runtime symlink merging, no fs monkey-patching.

That would eliminate the entire fs patching addition to `frame-fix-wrapper.js` (the biggest surface area for edge cases) and the runtime symlink logic in the launcher. The `isPackaged` fix and daemon pre-launch (issue 7) are still valid and would stay — it's just the resources path resolution that gets simpler.

This isn't necessarily the right call — you may have tried it and hit a blocker I'm not seeing. But given how many edge cases the runtime fs patching introduces (promises API, realpath, copyFile destinations, stale symlinks, empty globs), it's worth exploring if the build-time approach is feasible.

---

### Summary

If staying with the current approach, key changes needed before merge:

1. **Clear merged resources dir** on each launch to prevent stale symlinks (Issue 5)
2. **Guard empty globs** in the symlink loops (Issue 6)
3. **Redirect daemon stdout/stderr** to `/dev/null` (Issue 7)
4. **Add `fs.promises` patching** for completeness (Issue 2)
5. Handle `copyFile` destination path (Issue 1) and add `realpath` methods (Issue 3)

The overall direction is right. These are mostly edge-case hardening and cleanup items rather than fundamental design problems.

---

aaddrick:
Closing this out. The `isPackaged`/`resourcesPath` fix was addressed by #368 (build-time Electron co-location), and the remaining cowork service pre-launch piece hasn't been extracted into a separate PR.

Thanks again for the thorough investigation and NixOS testing — the breakdown of the three independent problems was genuinely valuable and informed the approach in #368.

If the daemon pre-launch logic is still needed in a future NixOS cowork effort, the analysis here is a great starting point.

---

PR #338, opened 2026-03-23
feat: add integration tests for build artifacts

sabiut:
## Summary
  - Add post-build validation tests for deb, rpm, and appimage packages
  - Tests verify package metadata, file layout, desktop entries, icons, launcher scripts, asar contents (frame-fix, cowork, native stub, tray icons), and --doctor smoke tests
  - Runs as a reusable workflow (test-artifacts.yml) with matrix strategy (one job per format) between build and release jobs, gating releases on passing artifact validation

  ## Test plan
  - [ ] CI triggers and all three matrix jobs pass (deb, rpm, appimage)
  - [ ] Release job dependency chain includes test-artifacts
  - [ ] Existing test-flags and build jobs unaffected

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,11 @@ jobs:
       artifact_suffix: ${{ matrix.artifact_suffix }}
       release_tag: ${{ startsWith(github.ref, 'refs/tags/v') && github.ref_name || '' }}
 
+  test-artifacts:
+    name: Test Build Artifacts
+    needs: [build-amd64]
+    uses: ./.github/workflows/test-artifacts.yml
+
   build-arm64:
     name: Build Packages (arm64 - ${{ matrix.artifact_suffix }})
     needs: test-flags
@@ -76,7 +81,7 @@ jobs:
   release:
     name: Create Release
     if: startsWith(github.ref, 'refs/tags/v')
-    needs: [test-flags, build-amd64, build-arm64]
+    needs: [test-flags, build-amd64, build-arm64, test-artifacts]
     runs-on: ubuntu-latest
     permissions:
       contents: write
--- a/.github/workflows/test-artifacts.yml
+++ b/.github/workflows/test-artifacts.yml
@@ -2,7 +2,9 @@ name: Test Build Artifacts (Reusable)
 
 on:
   workflow_call:
-  workflow_dispatch:
+
+permissions:
+  contents: read
 
 jobs:
   test-artifact:
@@ -42,7 +44,7 @@ jobs:
         if: matrix.format != 'rpm'
         run: |
           sudo apt-get update
-          sudo apt-get install -y file libfuse2
+          sudo apt-get install -y file libfuse2 nodejs npm
 
       - name: Run artifact tests
         run: |
--- a/.github/workflows/test-artifacts.yml
+++ b/.github/workflows/test-artifacts.yml
@@ -1,0 +1,50 @@
+name: Test Build Artifacts (Reusable)
+
+on:
+  workflow_call:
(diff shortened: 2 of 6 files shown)


aaddrick (changes requested):
Hey! This is solid work. Adding post-build validation tests for all three package formats is a great step toward catching regressions before they reach users. The shared helper approach in `test-artifact-common.sh` keeps things DRY, and the matrix strategy with format-specific containers is well thought out.

A few things to address before merging. Mostly straightforward fixes. This is close to ready.

### `.github/workflows/test-artifacts.yml`

**Issue 1: `workflow_dispatch` will always fail (high)**

The workflow declares `workflow_dispatch` as a trigger, but when dispatched manually there are no artifacts from a prior build job. `actions/download-artifact@v4` will fail with "artifact not found" every time.

Either remove `workflow_dispatch` or make it functional by accepting a `run_id` input. The simplest fix is just keeping `workflow_call` only.

**Issue 2: Missing Node.js in Ubuntu test dependencies (medium)**

The Ubuntu step installs `file` and `libfuse2` but not `nodejs`/`npm`. The `validate_app_contents` function uses `npx` as a fallback for asar extraction. Today this works because `ubuntu-latest` runners have Node pre-installed, but that's an implicit dependency. If the runner image changes, asar extraction silently gets skipped via the "Skipping asar extraction" fallback. That means the most important validation (frame-fix, cowork, native stub) just doesn't happen.

```yaml
      - name: Install test dependencies (Ubuntu)
        if: matrix.format != 'rpm'
        run: |
          sudo apt-get update
          sudo apt-get install -y file libfuse2 nodejs npm
```

**Issue 3 (minor): No explicit permissions block**

The reusable workflow inherits caller permissions. Adding explicit `permissions: contents: read` is good security practice for reusable workflows.

### `tests/test-artifact-deb.sh` and `tests/test-artifact-rpm.sh`

**Issue 4: `echo | grep` should use bash pattern matching (medium)**

Several places pipe `echo "$var" | grep -q` for simple substring checks. The project's style guide prefers bash builtins over external commands:

```bash
# Before
if echo "$pkg_info" | grep -q 'Package: claude-desktop'; then

# After
if [[ $pkg_info == *'Package: claude-desktop'* ]]; then
```

For the rpm regex case (`Name.*claude-desktop`), use `[[ =~ ]]`:

```bash
if [[ $rpm_info =~ Name.*claude-desktop ]]; then
```

4 instances across both files.

**Issue 5 (minor): Ambiguous `2>&1` on install commands**

```bash
if sudo dpkg -i --force-depends "$deb_file" 2>&1; then
```

This merges stderr into stdout but doesn't redirect anywhere. If the intent is to show output, just drop the `2>&1`. If it's to suppress, use `>/dev/null 2>&1`. Same in rpm.sh.

### `tests/test-artifact-common.sh`

**Issue 6 (minor): Arithmetic style**

```bash
_pass_count=$((_pass_count + 1))
```

Style guide prefers `((_pass_count++))`. Same for `_fail_count`.

### What's working well

- The shared test helper library is clean and well-structured
- Graceful degradation when asar tools aren't available is good defensive design
- `fail-fast: false` in the matrix means one format's failure won't mask others
- The `--doctor` smoke test with exit code threshold checking is clever
- RPM testing in a Fedora container is the right call
- Pipeline placement (after build, before release) is exactly right

### Summary

Main changes:
1. **Remove `workflow_dispatch`** or make it functional (Issue 1)
2. **Add `nodejs npm`** to Ubuntu test deps (Issue 2)
3. **Replace `echo | grep`** with `[[ ]]` pattern matching (Issue 4, 4 instances)

The rest are minor polish. This is close. Just a few tweaks and it's good to go.

---

PR #340, opened 2026-03-23
feat: configurable bwrap mount points via Linux config file

### LINKED ISSUE #339: Feature: configurable bwrap mount points in claude_desktop_config.json
## Summary

Currently, all BubbleWrap (bwrap) mount points in the Cowork sandbox are hardcoded in `BwrapBackend.spawn()` (`scripts/cowork-vm-service.js`, lines 774–892). Users have no way to add or remove mount points without modifying the source code.

This is a problem for users who need access to tools or data outside the default paths — for example:
- Developer tools installed in `/opt/` or `/snap/`
- Nix store paths (`/nix/store/`)
- Custom SDK directories
- Shared data mounts (`/media/`, `/mnt/`)

## Proposal

Add a `coworkBwrapMounts` key under `preferences` in `~/.config/Claude/claude_desktop_config.json` that allows users to declare additional bind mounts for the bwrap sandbox.

### Example configuration

```json
{
  "preferences": {
    "coworkBwrapMounts": {
      "additionalROBinds": ["/opt/my-tools", "/nix/store"],
      "additionalBinds": ["/media/shared-data"],
      "disabledDefaultBinds": ["/etc"]
    }
  }
}
```

| Key | Type | Description |
|-----|------|-------------|
| `additionalROBinds` | `string[]` | Extra paths mounted read-only inside the sandbox |
| `additionalBinds` | `string[]` | Extra paths mounted read-write inside the sandbox |
| `disabledDefaultBinds` | `string[]` | Default mounts to skip (e.g. `/etc` if the user provides a custom resolv.conf) |

### Behavior

- `cowork-vm-service.js` reads `claude_desktop_config.json` at daemon startup
- Additional mounts are appended to `bwrapArgs` **after** the default mounts
- `disabledDefaultBinds` removes matching entries from the default mount list before assembly
- Path validation: only absolute paths are accepted; paths like `/`, `/proc`, `/dev`, `/sys` are rejected to prevent breaking the sandbox
- **A daemon restart is required** for configuration changes to take effect (document in `--doctor` output and `CONFIGURATION.md`)

## Security considerations

- The core sandbox structure (`--tmpfs /`, `--unshare-pid`, `--die-with-parent`, `--new-session`, `--dev /dev`, `--proc /proc`) must remain non-configurable
- `additionalBinds` (read-write) should be restricted to paths under `$HOME` or an explicit allowlist
- `additionalROBinds` can accept any absolute path (read-only is safe)
- `disabledDefaultBinds` should warn via `--doctor` if critical mounts like `/usr` or `/etc` are disabled
- Mount order matters: user-configured mounts must not silently override security-critical read-only mounts (ref: the bug fixed in PR #309)

## Implementation notes

- **Where to read config:** `BwrapBackend` constructor or a new `loadConfig()` method called at daemon startup
- **Config path:** `process.env.HOME + '/.config/Claude/claude_desktop_config.json'`
- **Restart requirement:** The daemon is forked and detached at first launch and reused across sessions. Config changes require killing the daemon (`pkill -f cowork-vm-service`) so it gets re-launched on next Cowork session
- **Testing:** `BwrapBackend.spawn()` currently has no unit tests — adding coverage for mount arg construction would be valuable
- **Doctor integration:** `--doctor` should report custom mounts and warn about potentially dangerous configurations

## References

- PR #309 — Rewrote bwrap mount strategy from `--ro-bind / /` to minimal `--tmpfs /` approach
- `scripts/cowork-vm-service.js` — `BwrapBackend` class (lines 740–902)
- `docs/CONFIGURATION.md` — Existing configuration documentation

---

cbonnissent:
## Summary

- Allow users to configure BubbleWrap sandbox mount points through `~/.config/Claude/claude_desktop_linux_config.json` (a dedicated config file for the Linux port, separate from the official `claude_desktop_config.json`)
- Add path validation, config loading, and mount merging functions to `cowork-vm-service.js`
- Integrate `--doctor` diagnostics for custom mount configuration
- 33 new tests covering validation, config loading, mount merging, and doctor integration

### Configuration example

```json
{
  "preferences": {
    "coworkBwrapMounts": {
      "additionalROBinds": ["/opt/my-tools", "/nix/store"],
      "additionalBinds": ["/home/user/shared-data"],
      "disabledDefaultBinds": ["/etc"]
    }
  }
}
```

### Security

- Forbidden paths (`/`, `/proc`, `/dev`, `/sys`) always rejected
- Read-write mounts restricted to `$HOME`
- Critical mounts (`/`, `/dev`, `/proc`) cannot be disabled
- Mount order enforced: user mounts cannot override security-critical read-only mounts

### Applying changes

Daemon restart required (`pkill -f cowork-vm-service`) — documented in `CONFIGURATION.md` and `--doctor` output.

## Test plan

- [x] 33 BATS tests pass (validateMountPath, loadBwrapMountsConfig, mergeBwrapArgs, doctor integration)
- [x] 45 existing cowork-path-translation tests pass (no regression)
- [x] Built and installed .deb package locally
- [x] Verified custom mount (`~/.bun`) accessible inside Cowork sandbox
- [x] Verified `claude-desktop --doctor` reports custom mounts

Fixes #339

---
Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>
80% AI / 20% Human
Claude: Deep research, implementation plan, TDD implementation, issue creation, PR
Human: Feature design, config file naming decision, testing on live system, review

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -48,6 +48,57 @@ CLAUDE_MENU_BAR=visible claude-desktop
 export CLAUDE_MENU_BAR=visible
 ```
 
+## Cowork Sandbox Mounts
+
+When using Cowork mode with the BubbleWrap (bwrap) backend, you can customize
+the sandbox mount points via `~/.config/Claude/claude_desktop_linux_config.json`
+(a dedicated config for the Linux port, separate from the official
+`claude_desktop_config.json`):
+
+```json
+{
+  "preferences": {
+    "coworkBwrapMounts": {
+      "additionalROBinds": ["/opt/my-tools", "/nix/store"],
+      "additionalBinds": ["/home/user/shared-data"],
+      "disabledDefaultBinds": ["/etc"]
+    }
+  }
+}
+```
+
+| Key | Type | Description |
+|-----|------|-------------|
+| `additionalROBinds` | `string[]` | Extra paths mounted read-only inside the sandbox. Accepts any absolute path except `/`, `/proc`, `/dev`, `/sys`. |
+| `additionalBinds` | `string[]` | Extra paths mounted read-write inside the sandbox. **Restricted to paths under `$HOME`** for security. |
+| `disabledDefaultBinds` | `string[]` | Default mounts to skip. Cannot disable critical mounts (`/`, `/dev`, `/proc`). Use with caution: disabling `/usr` or `/etc` may break tools inside the sandbox. |
+
+### Security notes
+
+- Paths `/`, `/proc`, `/dev`, `/sys` (and their subpaths) are always rejected
+- Read-write mounts (`additionalBinds`) are restricted to paths under your home
+  directory
+- The core sandbox structure (`--tmpfs /`, `--unshare-pid`, `--die-with-parent`,
+  `--new-session`) cannot be modified
+- Mount order is enforced: user mounts cannot override security-critical
+  read-only mounts
+
+### Applying changes
+
+The daemon reads the configuration at startup. After editing the config file,
+restart the daemon:
+
+```bash
+pkill -f cowork-vm-service
+```
+
+The daemon will be automatically relaunched on the next Cowork session.
+
+### Diagnostics
+
+Run `claude-desktop --doctor` to see your custom mount configuration and any
+warnings about potentially dangerous settings.
+
 ## Application Logs
 
 Runtime logs are available at:
--- a/scripts/cowork-vm-service.js
+++ b/scripts/cowork-vm-service.js
@@ -398,19 +398,44 @@ function validateMountPath(mountPath, opts) {
 
     const normalized = path.resolve(mountPath);
 
-    if (FORBIDDEN_MOUNT_PATHS.has(normalized)) {
-        return { valid: false, reason: `Path is forbidden: ${normalized}` };
+    // Resolve symlinks when the path exists on disk (defense-in-depth).
+    // This is a TOCTOU situation, but bwrap is the real security boundary;
+    // this just catches honest configuration mistakes.
+    let resolved = normalized;
+    try {
+        resolved = fs.realpathSync(normalized);
+    } catch (_) {
+        // Path doesn't exist yet — use the unresolved form
     }
 
-    for (const forbidden of FORBIDDEN_MOUNT_PATHS) {
-        if (forbidden !== '/' && normalized.startsWith(forbidden + '/')) {
-            return { valid: false, reason: `Path is under forbidden path: ${forbidden}` };
+    function checkForbidden(p) {
+        if (FORBIDDEN_MOUNT_PATHS.has(p)) {
+            return `Path is forbidden: ${p}`;
+        }
+        for (const forbidden of FORBIDDEN_MOUNT_PATHS) {
+            if (forbidden !== '/' && p.startsWith(forbidden + '/')) {
+                return `Path is under forbidden path: ${forbidden}`;
+            }
+        }
+        return null;
+    }
+
+    const normalizedErr = checkForbidden(normalized);
+    if (normalizedErr) {
+        return { valid: false, reason: normalizedErr };
+    }
+
+    if (resolved !== normalized) {
+        const resolvedErr = checkForbidden(resolved);
+        if (resolvedErr) {
+            return { valid: false, reason: `Symlink resolves to forbidden path: ${resolved}` };
         }
     }
 
     if (opts.readWrite) {
         const home = os.homedir();
-        if (normalized !== home && !normalized.startsWith(home + '/')) {
+        const check = resolved !== normalized ? resolved : normalized;
+        if (check !== home && !check.startsWith(home + '/')) {
             return { valid: false, reason: 'Read-write mounts must be under $HOME' };
         }
     }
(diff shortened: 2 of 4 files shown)


aaddrick (changes requested):
Hey! This is solid work. Users who need Nix store paths, custom SDKs, or shared data in the Cowork sandbox had no option besides modifying the source. The config schema is clean, the security layers are sensible, and 33 BATS tests is thorough. Nice job testing on a live system too.

I've found a handful of issues. Most are about hardening the validation and cleaning up duplication in the shell code. The architecture itself is sound.

### `scripts/cowork-vm-service.js`

**Issue 1: `disabledDefaultBinds` paths aren't normalized before use**

`additionalROBinds` and `additionalBinds` go through `validateMountPath()` which calls `path.resolve()`, but `disabledDefaultBinds` skips validation entirely. It only checks `typeof p === 'string'`. A path like `/etc/` or `/dev/.` would slip past the `CRITICAL_MOUNTS` check but could still match (or not match) a default mount destination in unexpected ways.

You should run `disabledDefaultBinds` through normalization and critical-mount filtering at load time, inside `loadBwrapMountsConfig()`:

```javascript
disabledDefaultBinds: Array.isArray(mounts.disabledDefaultBinds)
    ? mounts.disabledDefaultBinds
        .filter(p => typeof p === 'string' && path.isAbsolute(p))
        .map(p => path.resolve(p))
        .filter(p => {
            if (CRITICAL_MOUNTS.has(p)) {
                warn(`BwrapConfig: cannot disable critical mount: ${p}`);
                return false;
            }
            return true;
        })
    : [],
```

**Issue 2: Symlinks can bypass forbidden path checks**

`path.resolve()` normalizes `.` and `..` but doesn't follow symlinks. If `/opt/sneaky` is a symlink to `/proc`, `validateMountPath('/opt/sneaky')` would pass all checks. Bwrap itself resolves symlinks when mounting, so the forbidden path would still get mounted.

You can add a `realpathSync` call as defense-in-depth:

```javascript
let normalized = path.resolve(mountPath);

// Resolve symlinks when the path exists on disk
try {
    normalized = fs.realpathSync(normalized);
} catch (_) {
    // Path doesn't exist yet — use the unresolved form
}
```

This is a TOCTOU situation (the symlink could change after validation), but bwrap is the real security boundary. This just catches honest configuration mistakes.

**Issue 3: Incomplete bwrap flag sets in `mergeBwrapArgs()`**

`TWO_ARG_FLAGS` and `THREE_ARG_FLAGS` only cover the flags currently used in the defaults. If a future commit adds `--setenv`, `--chdir`, `--ro-bind-try`, or `--file` to the defaults, the parser would misparse them. I'd expand the sets now to handle bwrap's full flag vocabulary:

```javascript
const THREE_ARG_FLAGS = new Set([
    '--ro-bind', '--ro-bind-try', '--bind', '--bind-try',
    '--dev-bind', '--dev-bind-try', '--symlink', '--chmod',
    '--setenv',
]);
const TWO_ARG_FLAGS = new Set([
    '--tmpfs', '--dev', '--proc', '--dir', '--remount-ro',
    '--file', '--unsetenv', '--chdir', '--size', '--perms',
]);
```

**Issue 4 (minor): Constructor logging**

Consider adding a log line when the config loads. It would help users troubleshoot why their changes aren't taking effect, since a daemon restart is required.

**Issue 5: The python3/node parser branching is duplicated 4 times**

The `if [[ $parser == 'python3' ]]; then ... else ... fi` pattern appears four times, once for each mount list. Each fork spawns a separate process. This is the most significant style issue.

You could extract all three mount lists in a single script invocation using a delimiter. Run one python3 or node script that prints all three lists separated by `---`, then parse the delimited output in bash. That cuts 4 process forks to 1 and eliminates the branch duplication.

**Issue 6 (minor):** Redundant `_doctor_colors` call in `_doctor_check_bwrap_mounts()`. Already called by the parent `run_doctor()`.

**Issue 7 (minor):** Duplicate "restart daemon" message. One inside the `critical_warned` block and one unconditionally at the end. The unconditional one is sufficient.

### `tests/cowork-bwrap-config.bats`

**Issue 8: Full function source is duplicated between test file and production code**

`NODE_PREAMBLE` contains a complete copy of `validateMountPath()`, `loadBwrapMountsConfig()`, and `mergeBwrapArgs()`. When the production code changes, the test preamble must be manually updated. Consider extracting the testable functions into a separate module that both the service and tests can require, or having the test preamble dynamically read from the production file.

### `docs/CONFIGURATION.md`

Looks good. One small observation: the diff shows the entire file rewritten (56 deletions, 107 additions) but the existing content appears unchanged. Looks like a whitespace or line-ending normalization happened alongside the new section. Worth confirming that was intentional.

### What's done well

- Security layering is thoughtful: forbidden paths, RW-home restriction, critical mount protection, clear separation between configurable defaults and per-spawn mandatory args
- The `validateMountPath()` path traversal check is a nice catch
- Clean early-return patterns in both JS and bash
- Test coverage is comprehensive with good edge cases
- Using a separate config file from upstream `claude_desktop_config.json` is the right call

### Key changes needed before merge

1. **Normalize `disabledDefaultBinds` paths** and move critical-mount filtering to load time (Issue 1)
2. **Add symlink resolution** in `validateMountPath()` as defense-in-depth (Issue 2)
3. **Expand bwrap flag sets** for forward compatibility (Issue 3)
4. **Consolidate the 4x parser duplication** to a single invocation (Issue 5)
5. Consider a strategy to avoid **duplicated function source** between tests and production code (Issue 8)

Issues 4, 6, and 7 are minor polish.

---

pmolodo:
This would greatly expand the usability of cowork!  Looking forward to seeing it merged (hopefully)!

pmolodo:
Tested this out, and it worked great! (Just had to remember to merge main first).

One small feature request, though - could you make it support the `~` shortcut, so I don't have to hardcode a username in my config file?  (and then update the example in docs/CONFIGURATION.md to use `~` instead of `/home/user`, so people will know it works...)

cbonnissent:
Hello @aaddrick any news on the PR ?

aaddrick:
Hey @cbonnissent, 

I've got a maintainer-related question I'd rather take off-thread. Could you drop me a line at aaddrick@gmail.com? Nothing urgent.

Thanks!

PR #388, opened 2026-04-10
fix: move --dir $HOME before --ro-bind sdkDir in bwrap sandbox

filiptrplan:
## Summary

**Bug**: When starting a new Cowork conversation, such an error appeared.
```
bwrap: execvp /home/.../.config/Claude/claude-code-vm/.../claude: No such file or directory
```

- bwrap processes mount args in order — when the SDK binary lives under `$HOME` (e.g. `~/.config/Claude/claude-code-vm/`), the `--ro-bind` of its parent directory was being added **before** `--dir $HOME`
- The later `--dir` wiped out the bind mount, causing `bwrap: execvp /home/.../.config/Claude/claude-code-vm/.../claude: No such file or directory`
- Fix: move `--dir $HOME` first so the home skeleton exists before subdirectory bind mounts are layered on top

### DIFF
--- a/scripts/cowork-vm-service.js
+++ b/scripts/cowork-vm-service.js
@@ -857,14 +857,16 @@ class BwrapBackend extends LocalBackend {
             log('BwrapBackend: could not resolve /etc/resolv.conf:', e.message);
         }
 
+        // Create home directory (needed for ~ expansion) but don't
+        // expose real home contents. Must come before any --ro-bind of
+        // subdirectories inside $HOME (e.g. the SDK binary path), otherwise
+        // bwrap processes --dir after the bind and shadows it.
+        const homeDir = os.homedir();
+        bwrapArgs.push('--dir', homeDir);
+
         // Bind the SDK binary read-only
         const sdkDir = path.dirname(actualCommand);
         bwrapArgs.push('--ro-bind', sdkDir, sdkDir);
-
-        // Create home directory (needed for ~ expansion) but don't
-        // expose real home contents.
-        const homeDir = os.homedir();
-        bwrapArgs.push('--dir', homeDir);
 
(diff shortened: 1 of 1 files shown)


aaddrick (changes requested):
Hey! Good catch on the bwrap mount ordering. The diagnosis is spot on. bwrap processes args sequentially, so `--dir $HOME` after `--ro-bind sdkDir` wipes the bind mount when the SDK path lives under `$HOME`. The reorder is the correct fix.

One thing to clean up before merging: the diff includes ~80 lines of whitespace-only changes in `resolveSubpath`, `buildSpawnEnv`, and `buildMountMap`. Those functions are functionally identical before and after. The extra churn makes the diff harder to review and pollutes `git blame` for future contributors trying to trace actual logic changes.

Can you resubmit with only the functional reorder? The easiest way is usually to reset and re-apply just the lines you intended to change. Something like:

1. `git diff HEAD~1 --stat` to confirm which files changed
2. Reset and selectively stage only the reorder hunks
3. Force-push the cleaned branch

The actual fix is ~15 lines and reads cleanly on its own. I'd like to keep the diff tight so it's obvious what changed and why.

Once the whitespace noise is removed, this is good to merge.

---
Written by Claude Opus 4.6 (1M context) via [Claude Code](https://claude.ai/code)

RayCharlizard:
Hey @filiptrplan — thanks for the diagnosis and the fix here, this was the right call. Since the branch hasn't moved in ~8 days and has since gone stale against main (we've had a few cowork-adjacent merges land), I've opened #426 which picks up your exact functional change on top of current main without the whitespace-only churn @aaddrick flagged.

You're credited as `Co-Authored-By` on the commit and in the PR body. Happy to hand it back if you'd prefer to land it yourself — just let me know and I'll close #426.

---

Issue #373, opened 2026-04-03
Bug: Cowork mode generates double-nested home directory paths


RayCharlizard:
## Summary

Cowork mode sessions inject file paths with the user's home directory doubled, e.g. `/home/username/home/username/.config/Claude/...` instead of `/home/username/.config/Claude/...`. When Claude accesses these paths, the OS creates the entire nested directory tree on disk, leaving behind a persistent `/home/username/home/` directory that must be manually cleaned up after every session.

## Environment

- **Package**: claude-desktop-debian
- **Mode**: Cowork
- **OS**: Debian-based Linux
- **Home directory**: `/home/raycharlizard`

## Reproduction

1. Open Claude Desktop and start a Cowork session.
2. Observe the system-configured paths provided to the agent. All workspace paths are double-nested:
   - Outputs: `/home/raycharlizard/home/raycharlizard/.config/Claude/local-agent-mode-sessions/<session-id>/outputs`
   - Uploads: `/home/raycharlizard/home/raycharlizard/.config/Claude/local-agent-mode-sessions/<session-id>/uploads`
   - Skills: `/home/raycharlizard/home/raycharlizard/.config/Claude/local-agent-mode-sessions/<session-id>/.claude/skills/...`
3. Any file operation by the agent creates the directory tree `/home/raycharlizard/home/raycharlizard/...` on the host filesystem.
4. After the session, `/home/raycharlizard/home/` persists and must be deleted manually.

## Expected behavior

All paths should resolve under the user's actual home directory once:

```
/home/raycharlizard/.config/Claude/local-agent-mode-sessions/<session-id>/outputs
```

## Actual behavior

Paths are prefixed with the home directory twice:

```
/home/raycharlizard/home/raycharlizard/.config/Claude/local-agent-mode-sessions/<session-id>/outputs
```

## Likely cause

The path construction logic appears to concatenate an absolute home-directory path (`/home/raycharlizard/...`) onto a base that already includes the home directory, rather than using a relative path or stripping the leading prefix. This is similar in nature to #350 (host path vs. guest path confusion in `mountPath()`).

## Impact

- A spurious `/home/username/home/` directory tree is created on every Cowork session.
- Users must manually delete it after each session.
- File operations that the agent believes are writing to the correct location may silently succeed at the wrong path, potentially causing confusion about where outputs actually land.

PR #389, opened 2026-04-12
fix: translate CLAUDE_COWORK_MEMORY_PATH_OVERRIDE on HostBackend

RayCharlizard:
## Summary

`buildSpawnEnv()` translates `CLAUDE_CONFIG_DIR` from `/sessions/` guest paths to host paths, but `CLAUDE_COWORK_MEMORY_PATH_OVERRIDE` passes through untranslated. On HostBackend, the `/sessions/` directory does not exist, so the auto-memory path points to a non-existent location and memory writes silently fail.

**Observed behavior:** `CLAUDE_COWORK_MEMORY_PATH_OVERRIDE` is set to `/sessions/eager-affectionate-ramanujan/mnt/.auto-memory` on a `COWORK_VM_BACKEND=host` system. The `/sessions/` directory doesn't exist. Auto-memory persistence is broken — writes fail with `EACCES: permission denied, mkdir '/sessions'`.

**Root cause:** Same pattern as #373 — the Electron app constructs guest paths regardless of backend, and the service needs to translate them. `CLAUDE_CONFIG_DIR` was already handled; `CLAUDE_COWORK_MEMORY_PATH_OVERRIDE` was missed.

## Fix

Adds guest-path translation for `CLAUDE_COWORK_MEMORY_PATH_OVERRIDE` in `buildSpawnEnv()`:

1. **Try `translateGuestPath()`** — works if `.auto-memory` is in `mountMap` via `additionalMounts`
2. **Fall back to `resolveSubpath()`** on the mount-name portion, mirroring what `HostBackend.mountPath()` would return (typically `~/.auto-memory`)
3. **Remove the env var** if neither translation succeeds, so Claude Code falls back to default memory behavior

BwrapBackend is unaffected — it overrides `spawn()` with its own env construction and `/sessions/` paths are real inside the sandbox.

## Test plan

- Verify on a `COWORK_VM_BACKEND=host` system that `CLAUDE_COWORK_MEMORY_PATH_OVERRIDE` resolves to `~/.auto-memory` (or the appropriate host path)
- Verify auto-memory writes succeed after the fix
- Verify BwrapBackend behavior is unchanged (no `/sessions/` translation in its spawn path)

## Environment

- CachyOS (Arch-based), native host backend
- Claude Desktop AppImage
- Diagnosed via `env | grep CLAUDE_COWORK_MEMORY_PATH_OVERRIDE` showing the untranslated `/sessions/` path

### DIFF
--- a/scripts/cowork-vm-service.js
+++ b/scripts/cowork-vm-service.js
@@ -228,8 +228,8 @@ function buildMountMap(additionalMounts, mountBinds) {
 
 /**
  * Build a merged environment for a spawned process. Combines filtered
- * daemon env with app-provided env, and translates CLAUDE_CONFIG_DIR
- * guest paths using mountMap.
+ * daemon env with app-provided env, and translates guest paths in
+ * CLAUDE_CONFIG_DIR and CLAUDE_COWORK_MEMORY_PATH_OVERRIDE using mountMap.
  */
 function buildSpawnEnv(appEnv, mountMap) {
     const mergedEnv = {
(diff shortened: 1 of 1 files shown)


aaddrick (approved):
Clean fix, follows the established pattern from #373 exactly. Confirmed the bug firsthand — auto-memory writes fail on HostBackend because the `/sessions/` guest path doesn't exist.

One minor nit: there's a double blank line before the new block (around line 268), but not worth blocking on.

---

Issue #245, opened 2026-02-19
Working Directory Misreported; CLAUDE.md Not Auto-Loaded


yowshi:
## Summary

When switching from Chat mode back to Cowork mode, the previously selected working folder is shown as still active (cached from the prior session). However, the underlying virtualized mount path from that session (e.g., `/sessions/<session-id>/mnt/rcdsktrials/`) no longer exists in the new session. Claude reports the folder as correctly set, but file operations fail because the path is stale. Additionally, because the working directory is not correctly resolved, any `CLAUDE.md` present in the selected folder is never auto-loaded, silently bypassing any user-defined rules it contains.

## Environment

- Application: Cowork (Claude desktop app, Cowork mode)
- Underlying framework: Claude Code / Claude Agent SDK
- OS: Debian GNU/Linux 12 (Bookworm)
- Desktop Environment: KDE Plasma (X11)

## Steps to Reproduce

1. Use Cowork with a selected folder (e.g., `~/Documents/rcdsktrials/`) in one session
2. Switch to Chat mode, then switch back to Cowork mode
3. Observe that the previous folder is shown as still selected — do not manually re-select it
4. Ask Claude what folder it is working in
5. Claude confirms the folder name (e.g., "rcdsktrials") without flagging any issue
6. Ask Claude to list files or read a file from that folder
7. Observe that Claude fails — the virtualized mount path from the previous session no longer exists

## Expected Behavior

- The working directory should resolve correctly to the actual system path of the selected folder
- `CLAUDE.md` present in the selected folder should be automatically read at session start, before any user interaction, consistent with how Claude Code handles memory files
- Claude should not confirm a working directory without verifying the path is actually accessible

## Actual Behavior

- Claude reports a virtualized path (`/sessions/<id>/mnt/<folder>/`) as the working directory
- That path does not exist or is empty, causing all file operations to fail
- `CLAUDE.md` in the selected folder is never loaded, so any user-defined rules or instructions it contains are silently ignored for the entire session
- Claude accepted and confirmed the misreported path without performing any sanity check

## Impact

- User-defined rules in `CLAUDE.md` (e.g., confirmation requirements before running commands, file operation restrictions) are bypassed entirely — not intentionally, but because they are never seen
- This is a significant trust and safety issue: users may rely on `CLAUDE.md` to enforce guardrails, and those guardrails silently fail to load
- File read/write operations fail unexpectedly, causing confusion

## Workaround

[...]

PR #392, opened 2026-04-13
fix: resolve working directory from primary mount on HostBackend

RayCharlizard:
## Summary

- On HostBackend, `resolveWorkDir()` falls back to `os.homedir()` for every Cowork session because the Electron app sends `cwd=/sessions/{name}` (a session-root guest path) and `translateGuestPath()` requires `/sessions/{name}/mnt/{mount}/...` — the session root never matches.
- This means Claude Code always starts in `~` instead of the user's project directory, breaking file access and CLAUDE.md auto-load (issue #245).
- The fix adds primary-mount derivation (mirroring BwrapBackend's existing logic) as a fallback before resorting to homedir.

Related: #245 (umbrella), #389 (sibling fix for `CLAUDE_COWORK_MEMORY_PATH_OVERRIDE`).

## Root Cause

`translateGuestPath()` uses a regex that requires a `/mnt/{mount}` component in the path. Session-root paths like `/sessions/bold-sharp-clarke` don't have one, so translation always fails. BwrapBackend avoids this by overriding `spawn()` with its own CWD derivation from the primary user mount. HostBackend goes through `resolveWorkDir()` which lacked equivalent logic.

## The Fix

When `translateGuestPath()` can't resolve the cwd, derive the working directory from the primary user mount (first non-dotfile, non-uploads key in `mountMap`) — the same heuristic BwrapBackend uses. Falls back to homedir only when no user mount exists.

The logic is now in a shared `findPrimaryMount(mountMap)` helper used by both `resolveWorkDir()` and `BwrapBackend.spawn()`, so the two call sites can't drift.

### Behavior shift beyond the stated case

The new fallback also fires when the cwd takes the form `/sessions/{id}/mnt/unknown` — where the path *is* shaped correctly but the mount name isn't in `mountMap` (translation fails on the mount-name lookup, not the regex). Previously these landed in `~`; they now land in the primary mount. This is intentional and consistent with what BwrapBackend would do.

## Verification

Tested with a Node.js harness simulating exact spawn parameters from live session logs:

| Scenario | Before | After |
|---|---|---|
| Session root + user mount | `~/` (bug) | `/home/user/project/` (correct) |
| Session root + no user mount | `~/` | `~/` (correct fallback) |
| Session root + only dotfile/uploads mounts | `~/` | `~/` (correct fallback) |
| `/mnt/unknown` mount-name miss | `~/` | primary mount (intentional shift) |
| Full guest mount path | Translated correctly | Unchanged |
| Host path passthrough | Passed through | Unchanged |
| `sharedCwdPath` precedence | Takes precedence | Unchanged |

BATS coverage in `tests/cowork-path-translation.bats`:
- 3 new `resolveWorkDir` cases for the primary-mount fallback paths
- 4 new `findPrimaryMount` cases covering null/empty input, mount-class filtering, and ordering

## Test plan

- [ ] Start a Cowork session via Projects sidebar → verify Claude Code CWD is the project directory, not `~/`
- [ ] Start a Cowork session via "New task" → verify same behavior
- [ ] Start a session with no folder selected → verify graceful fallback to homedir
- [ ] Run `bats tests/cowork-path-translation.bats` → all `resolveWorkDir` and `findPrimaryMount` cases pass

---
Co-Authored-By: Claude Opus 4.6, Claude Opus 4 <noreply@anthropic.com>
40% AI / 60% Human
Claude: diagnosed the bug, implemented the fix, refactored the shared helper, wrote BATS tests, drafted PR description
Human: identified the symptom, validated the diagnosis against live daemon logs, designed the contribution sequence

### DIFF
--- a/scripts/cowork-vm-service.js
+++ b/scripts/cowork-vm-service.js
@@ -227,6 +227,23 @@ function buildMountMap(additionalMounts, mountBinds) {
 }
 
 /**
+ * Find the primary user mount name in mountMap — the first key that
+ * is not a dotfile mount (e.g. .claude, .auto-memory) and not the
+ * uploads mount. Used by both resolveWorkDir (HostBackend) and
+ * BwrapBackend.spawn to derive a sensible cwd from the user-selected
+ * project folder when the Electron app sends a session-root path with
+ * no /mnt/{name} component to translate.
+ *
+ * Returns the mount name (string) or null if no user mount exists.
+ */
+function findPrimaryMount(mountMap) {
+    if (!mountMap) return null;
+    return Object.keys(mountMap).find(
+        n => !n.startsWith('.') && n !== 'uploads',
+    ) || null;
+}
+
+/**
  * Build a merged environment for a spawned process. Combines filtered
  * daemon env with app-provided env, and translates guest paths in
  * CLAUDE_CONFIG_DIR and CLAUDE_COWORK_MEMORY_PATH_OVERRIDE using mountMap.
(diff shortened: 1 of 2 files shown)


aaddrick:
Thanks much!

Issue #233, opened 2026-02-16
fixing blurry text on wayland with CLAUDE_USE_WAYLAND=1


benzipperer:
Thank you for the great package! 

Just want to note that using `CLAUDE_USE_WAYLAND=1` [as suggested](https://github.com/aaddrick/claude-desktop-debian?tab=readme-ov-file#environment-variables) also solved some text rendering issues for me on Wayland in Ubuntu 24.04. I don't know if others had or will have this problem, but in case they do, maybe it's worth documenting this benefit of using native Wayland?

Left-side shows Claude via a web browser and right-side shows claude-desktop-debian, without setting the environment variable.

<img width="2880" height="1800" alt="Image" src="https://github.com/user-attachments/assets/1f63296e-dfb1-4904-8ee6-3ca9dae69bbc" />

With the environment variable set, claude-desktop-debian on the right is much clearer.

<img width="2880" height="1800" alt="Image" src="https://github.com/user-attachments/assets/7ef21611-2fad-4a7d-b93c-eb82083039f2" />

aaddrick:
That's sick, appreciate the photos. 

I'm in a weird spot with Wayland. I'd prefer to support it by default, but so many backends have choppy xdg-desktop integrations which break claude-desktop features.

Going down that rabbit hole, the wayland team doesn't want to address the underlying causes which make xdg-desktop a requirement, which is their choice to make. Teams like COSMIC are slowly working towards getting things like GlobalShortcuts wired up, but the quick access menu won't work until it's done. 

All said, I get less tickets by defaulting to X11.  I'm happy to add more instructions for Wayland if someone wants to submit a PR with the revision.  I started down the route of mapping compatibility between backends and features, but abandoned the effort.

Issue #141, opened 2025-11-18
Global hotkey for quick search bar not working on Linux/Wayland


megaray:
The global hotkey setting exists in the Claude Desktop preferences, but it doesn't work on Linux because the `claude-native` module is a stub with no-op functions.

Expected behavior: Pressing Ctrl+Alt+Space should show a floating search bar overlay
Actual behavior: Nothing happens

Environment:
- OS: Wayland
- DE: GNOME
- Version: 0.14.10 

Would it be possible to implement native global hotkey support using:
- GlobalShortcutsPortal for Wayland
- X11 global hotkeys for X11 sessions

PR #97, opened 2025-07-25
Add Wayland compatibility support with graceful error handling

delorenj:
## 🎉 Add Wayland Compatibility Support

Hey @aaddrick!

First off, **huge thanks** for creating this since apparently Anthropic doesn't have the budget apparently to assign this to an intern that could crank out an official version in less than an hour. 😞 

## ✨ Wayland Support
This PR adds **full Wayland compatibility** to Claude Desktop! 🚀

### 🐛 Problem Solved
Users running Wayland sessions (like GNOME on Wayland, KDE Plasma Wayland, etc.) will experience crashes and display issues when trying to run Claude Desktop.

### ✨ Wayland Solution
The fix implements smart session detection and compatibility mode:
- ✅ **Detects Wayland sessions** automatically
- ✅ **Forces X11 backend** to avoid Wayland GPU issues (`GDK_BACKEND=x11`)
- ✅ **Sets Electron compatibility flags** (`ELECTRON_OZONE_PLATFORM_HINT=x11`)
- ✅ **Disables GPU acceleration** to prevent dmabuf errors (`ELECTRON_DISABLE_GPU=1`)
- ✅ **Seamless user experience** - Nothing else to do - just works.

## 🎁 Bonus: Better Error Handling
As a small enhancement, also added graceful error handling for edge cases (like TTY sessions) with user-friendly error messages instead of cryptic crashes. This kept happening to me while iterating on this fix.

## 🧪 Testing
- ✅ Tested on Wayland sessions - works perfectly!
- ✅ Tested on X11 sessions - maintains compatibility - no change
- ✅ Graceful fallback for unsupported environments

Thanks again for making this! 🎉

Issue #93, opened 2025-07-15
Shortcut doesn't work


shahmal1yev:

aaddrick:
are you talking about the launcher shortcut, or the keyboard shortcut for the quick launch?

aaddrick:
We just added some additional Wayland support, and i think there needs to be some additional logic to get the shortcut working in that environment. I tried running it with the --ozone-platform=x11 flag,  but no dice. I'll dig into it some more this afternoon

shahmal1yev:
I'm currently using GNOME with Wayland.

shahmal1yev:
I noticed the new release yesterday and updated to the latest version right away. However, the shortcut still doesn’t seem to work in my environment.

If there’s anything specific I should try or test on my side, I’d be happy to help.

aaddrick:
OK, so I've been looking into it and it comes down to Wayland handling global shortcut registration differently.

There's a fix in the works via xdg-desktop-portal, but it has to come through all the different desktop environment implementations first. GNOME and KDE both have merge requests pending as of January 2025, but until those land and make it into stable releases, we're stuck with workarounds.

The launcher script has been updated with the right flags (--enable-features=GlobalShortcutsPortal etc.) and switched from X11 compatibility mode to native Wayland backend so it should "just work" once the portal backends catch up.

I'm figuring out if/how to setup a global shortcut manually to give the same functionality for Wayland until xdg-desktop-portal and the others implement the GlobalShortcuts portal.

PR #397, opened 2026-04-13
fix: export GDK_BACKEND=wayland in native Wayland mode

aJV99:
## Problem

When the launcher runs in native Wayland mode (`CLAUDE_USE_WAYLAND=1` or forced by compositor), it sets the correct Electron ozone flags (`--ozone-platform=wayland` etc.) but never exports `GDK_BACKEND`. A system-wide or session-level `GDK_BACKEND=x11` then silently overrides the intent, causing GTK to connect via XWayland and producing blurry rendering — notably on HiDPI displays.

## Fix

Export `GDK_BACKEND=wayland` in the native Wayland branch of `build_electron_args()` so the ozone flags and GDK backend stay in sync.

## Scope

Single-line change. No behaviour change for XWayland mode or X11 sessions.

### DIFF
--- a/scripts/launcher-common.sh
+++ b/scripts/launcher-common.sh
@@ -88,6 +88,10 @@ build_electron_args() {
 		electron_args+=('--ozone-platform=wayland')
 		electron_args+=('--enable-wayland-ime')
 		electron_args+=('--wayland-text-input-version=3')
+		# Override any system-wide GDK_BACKEND=x11 that would silently
+		# prevent GTK from connecting to the Wayland compositor, causing
+		# blurry rendering or launch failures on HiDPI displays.
+		export GDK_BACKEND=wayland
 	fi
 }
 


aaddrick:
Thanks for the PR and the detailed write-up. The blurry rendering on HiDPI Wayland is a real problem — we've had it reported before (#233, #17), and your analysis of `GDK_BACKEND=x11` silently overriding the ozone flags is a good catch.

Some context on where things stand with Wayland in this project:

We've been through a few iterations on this (PR #97 → #102 → #153). PR #102 (Aug 2025) switched to native Wayland with `GlobalShortcutsPortal` enabled, anticipating that `xdg-desktop-portal` backends would catch up. In practice the experience was inconsistent — Electron's `globalShortcut.register()` relies on X11's `XGrabKey` and the portal-based replacement (`org.freedesktop.portal.GlobalShortcuts`) didn't work reliably end-to-end, even on DEs that nominally supported it. The quick entry window (Ctrl+Alt+Space) is a key feature, and losing it generated issues we couldn't fix (#141, #93, #11). PR #153 (Jan 2026) defaulted back to XWayland and put native Wayland behind `CLAUDE_USE_WAYLAND=1`.

On paper the `xdg-desktop-portal` picture has improved — GNOME 48+ and KDE both have `GlobalShortcuts` backends, Electron 35+ has client-side support, and Hyprland has it too. But Sway is still blocked on a missing upstream protocol ([xdg-desktop-portal-wlr #240](https://github.com/emersion/xdg-desktop-portal-wlr/issues/240)), Niri and COSMIC don't have it, and the UX model is fundamentally different — the portal has the DE prompt users to assign keybindings rather than the app grabbing them silently. We haven't re-validated the end-to-end experience on GNOME 48 or recent KDE, so I can't say with confidence that flipping back to native Wayland "just works" today even where the portal exists.

That said, I think there are usable pieces in this PR:

1. **The `GDK_BACKEND=wayland` export** — I'd take this as-is. When someone has opted into native Wayland via `CLAUDE_USE_WAYLAND=1`, a system-level `GDK_BACKEND=x11` silently overriding the intent is a bug.

2. **The HiDPI auto-detection** — the `gsettings org.gnome.desktop.interface scaling-factor` query is GNOME-specific. It won't help KDE (uses `kscreen`/`kwriteconfig`), Sway/Hyprland (`output scale` in config), or fractional scaling setups. More importantly, auto-switching users to native Wayland without them understanding why their global hotkeys changed behavior could generate confusion. I'd be more comfortable incorporating this behind the existing flag infrastructure — something like `CLAUDE_USE_WAYLAND=auto` that enables DE-specific detection, so users opt in knowingly and understand the trade-offs.

The broader native-Wayland-by-default path is a backlog item until someone takes it on as a focused effort to validate the end-to-end experience across DEs — or the ecosystem matures to the point where we can confidently flip the default. Improvements that work within the existing flag structure are welcome in the meantime.

Would you be open to splitting the `GDK_BACKEND` fix out? That's straightforward and I'd merge it quickly. For the HiDPI piece, let's discuss whether a `CLAUDE_USE_WAYLAND=auto` approach would work for your use case.

---

---

aJV99:
Thanks for the thorough context — that background on #97 → #102 → #153 is really helpful and makes the caution around auto-switching completely understandable.

Agreed on splitting. I'll update this branch to just the `GDK_BACKEND=wayland` export so it's a clean, reviewable fix on its own.

On the HiDPI piece — `CLAUDE_USE_WAYLAND=auto` makes sense as the right shape. The idea would be:

- `CLAUDE_USE_WAYLAND=1` → native Wayland (current explicit opt-in, unchanged)
- `CLAUDE_USE_WAYLAND=0` → XWayland (explicit opt-out, unchanged)
- `CLAUDE_USE_WAYLAND=auto` → DE-specific detection, native Wayland only if we can confidently determine HiDPI is active

For detection, we could layer it:
- **GNOME**: `gsettings get org.gnome.desktop.interface scaling-factor`
- **KDE**: `kreadconfig6 --file kdeglobals --group KScreen --key ScaleFactor`
- **Wlroots compositors (Sway/Hyprland)**: `wlr-randr --json` or parse `$WAYLAND_DISPLAY` socket output — admittedly harder, might just skip auto for those
- Fractional scaling (>1 but not integer) would also qualify since XWayland still blurs

The user experience story would be: if you set `CLAUDE_USE_WAYLAND=auto`, the launcher picks native Wayland when it detects HiDPI, and logs why. Global hotkey behaviour is unchanged for anyone who hasn't set the flag.

Happy to follow up with a second PR implementing `auto` once this `GDK_BACKEND` fix lands. Does that direction work for you?

aaddrick:
Hey! The diff looks great now, thanks for splitting it out.

One more thing before I merge: the title and body still describe all three original changes. Could you trim them down to just the `GDK_BACKEND=wayland` export so the merge commit matches the actual scope?

Sorry for not flagging this when I asked for the split — should have called it out then. Once the title and body are updated, I'll get this merged.

---

aJV99:
Apologies for the delayed response. I have updated the title and body now. I'll have a PR out to you soon regarding the HiDPI auto-detection piece.

PR #420, opened 2026-04-17
build.sh: improve regexp quick window patch regexp readibility

Andrej730:
@aaddrick

A small change to make regexps more readable.
Wanted to also use `RegExp.escape`, but it's only available in node 24+, so just moved existing regexp to a separate function `escapeRegExp`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,6 +236,7 @@ Special thanks to:
 - **[davidamacey](https://github.com/davidamacey)** for identifying and fixing the XRDP GPU compositing blank-window issue on remote desktop sessions
 - **[pb3ck](https://github.com/pb3ck)** for diagnosing the Cowork `CLAUDE_CODE_OAUTH_TOKEN` env-strip bug with a working reference diff
 - **[aJV99](https://github.com/aJV99)** for exporting `GDK_BACKEND=wayland` in native Wayland mode to fix XWayland fallback blur on HiDPI displays
+- **[Andrej730](https://github.com/Andrej730)** for the quick-window regex readability refactor (`String.raw` + `escapeRegExp` helper)
 
 ## Sponsorship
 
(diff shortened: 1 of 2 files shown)


aaddrick:
Thanks @Andrej730 — the `String.raw` swap is a nice readability win (the `\\(\\)\\|\\|` string was painful to read), and the inline "matches: ..." comment is helpful for future readers.

The PR got tangled up in a refactor I landed in `ff4821e` on 2026-04-20, which split `build.sh` into topical modules under `scripts/patches/`. The anchor array and `showRe` block you're touching now live in [`scripts/patches/quick-window.sh`](https://github.com/aaddrick/claude-desktop-debian/blob/main/scripts/patches/quick-window.sh) (lines ~87-105). The logic is byte-identical, just relocated — so the rebase should be a clean move of your two-hunk diff to the new file.

If you're up for rebasing onto `main` and retargeting, happy to merge from there. Sorry for the churn — the split landed three days after you opened this.

---

Andrej730:
@aaddrick rebased.

aaddrick (approved):
Hey @Andrej730! Verified the equivalence by running both code paths in node. The escape helper output matches byte-for-byte (`pF` stays `pF`), and the final regex source comes out identical: `pF\(\)\|\|(\w+)\.show\(\)`. Pure readability win, no behavior change. Per our simplifier-skip rule for byte-identical refactors, this one's a pass.

`String.raw` is the right call here. The old `'\\(\\)\\|\\|(\\w+)\\.show\\(\\)'` was rough to scan, and the template literal version reads much closer to the actual regex. Good instinct on falling back to a local `escapeRegExp` since we're on Node 20 and `RegExp.escape` needs 24+.

Rebase looks clean against the `build.sh` → `scripts/patches/` split from `ff4821e`. The two hunks landed in `scripts/patches/quick-window.sh` around lines 87-104, and they don't overlap with #496 (line 72) or #515 (different file), so the merge order doesn't matter. I pushed a `docs(readme)` commit (`4c6f200`) adding you to the Acknowledgments. Approving.

---

Issue #418, opened 2026-04-17
Cowork VM startup crashes with TypeError: e.existsSync is not a function (claude-desktop 1.3109.0-1.3.31)


Joost-Maker:
Here's the issue, ready to paste at https://github.com/aaddrick/claude-desktop-debian/issues/new:

---

**Title:** Cowork VM startup crashes with `TypeError: e.existsSync is not a function` (claude-desktop 1.3109.0-1.3.31)

---

### Summary

Switching from Claude Code mode to Claude Cowork mode in the desktop app reliably crashes the workspace VM startup. The UI shows:

> **Failed to start Claude's workspace**
> Cannot read properties of undefined (reading 'existsSync')

Claude Code mode itself works fine — the crash is specific to Cowork's VM startup path.

### Environment

- **Package:** `claude-desktop 1.3109.0-1.3.31` (amd64, installed via .deb)
- **OS:** Pop!_OS 24.04 LTS
- **Kernel:** 6.18.7-76061807-generic
- **Node (system):** v22.22.2
- **Install path:** `/usr/lib/claude-desktop/`

### Steps to reproduce

1. Launch Claude Desktop.
2. App opens normally in Claude Code mode.
3. Switch to Claude Cowork.
4. App briefly restarts, then the red "Failed to start Claude's workspace" banner appears.

### Expected

Cowork workspace boots and shows the "Let's knock something off your list" entry screen without errors.

### Actual

VM startup fails. Clicking the in-app **"reinstall the workspace"** link doesn't help — auto-reinstall is attempted once and then suppressed (`Skipping auto-reinstall (already attempted once)`).

### Log: `~/.config/Claude/logs/cowork_vm_node.log`

```
[VM:start] Beginning startup, bundlePath=/home/<user>/.config/Claude/vm_bundles/claudevm.bundle
[VM:start] Bundle version: 5680b11bcdab215cccf07e0c0bd1bd9213b0c25d
[VM:steps] download_and_sdk_prepare started
[VM:steps] download_and_sdk_prepare completed (2ms)
[VM:steps] load_swift_api started
[VM:steps] load_swift_api completed (0ms)
[Callbacks] Initializing VM event callbacks...
[Callbacks] VM event callbacks initialized
[VM:start] Startup failed: TypeError: e.existsSync is not a function
    at xTr (/usr/lib/claude-desktop/node_modules/electron/dist/resources/app.asar/.vite/build/index.js:1897:4165)
[...]

PR #419, opened 2026-04-17
Fix cowork patch regexes to match dollar-prefixed minified vars

Smilez1985:
## What this fixes

In Claude Desktop 1.3109.0 (and presumably later releases with the same minification pattern), `patch_cowork_linux()` fails to correctly extract the minified `fs` variable name from the upstream Vite bundle, producing an `app.asar` where the injected Linux `smol-bin` copy block calls `e.existsSync(...)` on the enclosing function's options parameter instead of on `fs`. The Cowork tab then fails to start with:

```
TypeError: existsSync is not a function
```

in the UI, and

```
Cannot read properties of undefined (reading 'existsSync')
    at xTr (...app.asar/.vite/build/index.js:1897:4165)
```

in the launcher log.

## Root cause

Six regexes on lines 1437-1453 of `build.sh` use `\w+` to capture minified variable names. `\w+` only matches `[A-Za-z0-9_]`. In 1.3109.0, Vite binds `fs` as `$e`; the `$` is skipped by `\w+` and the capture returns only `e`, which is the options parameter of the enclosing `async function xTr(e, A, t, i)`.

## Fix

Change `\w+` to `\$?\w+` in all six variable-extraction regexes so a leading `$` is captured as part of the identifier. This is backwards-compatible: non-dollar-prefixed variables continue to match unchanged.

## Verification

Built and installed on an arm64 Raspberry Pi 5, Debian Trixie, upstream Claude Desktop 1.3109.0. Build log now reports:

```
vars: path=ae fs=$e log=qe stream=SL arch=Bre bundle=r
Injected Linux smol-bin copy block (skips _.configure)
```

Note `fs=$e` rather than the previous `fs=e`. After install, the Cowork tab loads and functions normally; no further `existsSync` errors appear in the launcher log.

## Scope

This patch only touches the six regexes in `patch_cowork_linux()`. It does not address the secondary `C.getAppInfoForFile is not a function` that appears elsewhere in the UI log — that is a separate unrelated issue.

## Credit

Root cause and the one-character fix were diagnosed by the repo's `github-actions` triage bot in a comment on issue #417. This PR implements that diagnosis.

Fixes #417

Issue #415, opened 2026-04-17
TypeError: e.existsSync is not a function on Fedora 43


iovcho:
Package: claude-desktop-1.3109.0-1.3.31.fc42.x86_64
Fix: Add to ~/.config/Claude/claude_desktop_linux_config.json:

{
  "coworkIsolationBackend": "host",
  "coworkEnabled": false
}

Root cause: load_swift_api step in VM startup calls e.existsSync where e (minified fs module) is not injected correctly in Electron sandbox context on Linux. Stack trace: xTr at index.js:1897:4165.
Doctor output: All deps pass (bubblewrap, KVM, vsock, QEMU, socat, virtiofsd), but VM image: not downloaded yet.

iovcho:
`yyovchev@Vivobook:~$ cat ~/.config/Claude/logs/main.log
2026-04-17 11:05:01 [info] Starting app {
  appVersion: '1.3109.0',
  isPackaged: true,
  platform: 'linux',
  arch: 'x64',
  nodeVersion: '24.14.1'
}
2026-04-17 11:05:03 [info] [CCD] Initialized with version 2.1.111
2026-04-17 11:05:03 [info] [growthbook] next refresh in 60 min
2026-04-17 11:05:03 [info] [IdleManager:session] Initialized with 900s timeout
2026-04-17 11:05:03 [info] [IdleManager:preview] Initialized with 1800s timeout
2026-04-17 11:05:03 [info] Starting app {
  appVersion: '1.3109.0',
  isPackaged: true,
  platform: 'linux',
  arch: 'x64',
  nodeVersion: '24.14.1'
}
2026-04-17 11:05:04 [info] [office-addin] Module loaded
2026-04-17 11:05:04 [info] [office-addin] MCP Server created
2026-04-17 11:05:04 [info] App is installed, enabling initial check and auto-updates
2026-04-17 11:05:04 [info] Update URL: https://api.anthropic.com/api/desktop/linux/x64/squirrel/update
2026-04-17 11:05:04 [info] [CoworkArtifacts] Protocol handler registered
2026-04-17 11:05:04 [info] [CoworkFilePreview] Protocol handler registered
2026-04-17 11:05:04 [info] [account] Account details not yet available, waiting for IPC...
2026-04-17 11:05:04 [info] [Chrome Extension MCP] Skipping native host setup: binary not found at /usr/lib/claude-desktop/node_modules/electron/dist/resources/chrome-native-host
[...]

PR #421, opened 2026-04-17
fix: cowork existsSync crash on 1.3109+ and unblock node-pty terminal

Joost-Maker:
Two independent Linux bugs surfaced while debugging Cowork on Claude Desktop 1.3109.0 (Pop!_OS 24.04, amd64). Both fixed and verified end-to-end on my machine.

## Bug 1: `TypeError: e.existsSync is not a function` on Cowork VM boot

**Issue:** #418

`patch_cowork_linux()` Patch 9 extracts six minified variable names from the win32 block and uses them to template a Linux smol-bin copy block. The extraction regexes used `(\w+)`, which doesn't match `$` — but JS identifiers can start with `$`, `_`, or a letter.

In Claude 1.3109.0, upstream renamed the local fs reference inside `startVM`'s win32 block from `e` to `$e`, presumably to avoid shadowing the function parameter `e` (which is the options object):

```js
// Win32 block in 1.3109.0 (excerpt)
if (process.platform === "win32") {
    const x = Bre(), U = ae.join(process.resourcesPath, `smol-bin.${x}.vhdx`), J = ae.join(r, "smol-bin.vhdx");
    if ($e.existsSync(U)) {                       // <-- fs is $e, not e
        await SL.pipeline($e.createReadStream(U), SX(J))
        ...
    }
}
```

The existing `(\w+)\.existsSync\(` regex scans `$e.existsSync(U)`, skips the `$`, captures just `e`, and Patch 9 then injects:

```js
if (process.platform === "linux") {
    ...
    e.existsSync(_ls) ? ... : ...   // e is the options object → TypeError
}
```

The "patch's injection works in 1.2773 but breaks in 1.3109" pattern is exactly this. There's no upstream Linux block to strip — the buggy block we've all been hitting **is the patcher's own injection**, made buggy by the regex missing `$`.

**Fix:** widen all six extraction patterns from `(\w+)` to `([$\w]+)`. Also widened the adjacent unanchored matchers in `archMatch` for consistency.

**Defensive bonus:** added a brace-counting strip step before injection, so if a future upstream version emits its own `if(process.platform==="linux"){...}` block right after the win32 close brace, we remove it before injecting our correctly-wired version. No-op on current upstream; future-proofing.

**Verified:** clean rebuild now logs `vars: path=ae fs=$e log=qe stream=SL arch=Bre bundle=r` and the asar's injected block contains `$e.existsSync(_ls)`. Cowork starts cleanly: `[VM:start] Startup complete, total time: 1242ms`, the VM agent spawns, and prompts get responses end-to-end.

## Bug 2: `Failed to load node-pty` (terminal in Claude Code mode)

Surfaced after fixing Bug 1, when the app reached Claude Code mode with a working shell session attempt. Error:

```
Failed to load node-pty {
  error: Error: Failed to load native module: pty.node, checked: build/Release, build/Debug, prebuilds/linux-x64:
    Error: Cannot find module './prebuilds/linux-x64//pty.node'
  - /usr/lib/claude-desktop/.../app.asar/node_modules/node-pty/lib/utils.js
}
```

**Root cause:** `install_node_pty()` copied only `lib/` and `package.json` into `app.asar.contents/.../node-pty/`, and `finalize_app_asar()` packed app.asar without `--unpack`. The `.node` binaries were then separately dropped into `app.asar.unpacked/.../node-pty/build/Release/`.

The asar manifest therefore had no entry for `node-pty/build/` at all. When node-pty's loader (inside the asar) does `require('../build/Release/pty.node')`, Electron's asar→.unpacked redirect never fires because the redirect requires a manifest entry annotated as unpacked. Result: MODULE_NOT_FOUND despite the binary existing on disk.

**Fix:**
1. `install_node_pty()`: also stage `$pty_src_dir/build/` into `app.asar.contents/node_modules/node-pty/`.
2. `finalize_app_asar()`: pass `--unpack '**/*.node'` to `asar pack` so the binaries get moved into `app.asar.unpacked/` AND recorded in the manifest as unpacked.

**Verified:** new asar manifest now reports
```
UNPACKED: /node_modules/node-pty/build/Release/pty.node
UNPACKED: /node_modules/node-pty/build/Release/conpty.node
UNPACKED: /node_modules/node-pty/build/Release/conpty_console_list.node
```
and Claude Code's terminal loads successfully. Tested by running interactive shell sessions in Claude Code mode of the rebuilt .deb.

The pre-existing copy-to-.unpacked step at the bottom of `finalize_app_asar()` is now redundant but harmless (writes the same bytes). Left in place for now to minimize diff and preserve the `--node-pty-dir` (Nix) flow, where the explicit copy may still be load-bearing.

## Test plan

- [x] `./build.sh --build deb --clean no` against upstream 1.3109.0 produces a working .deb
- [x] Installed .deb on Pop!_OS 24.04 (amd64)
- [x] Cowork mode: VM boots, agent spawns, prompts answered (verified `[VM:start] Startup complete` in `cowork_vm_node.log` + actual chat in the UI)
- [x] Claude Code mode: terminal sessions start without the node-pty error
- [ ] Smoke test on an arm64 machine (don't have one handy; the fix is arch-agnostic)
- [ ] Smoke test against an older Claude Desktop version where fs is just `e` not `$e` — `[$\w]+` is a strict superset of `\w`, so existing extraction still works

## Notes

- Both fixes are minimal and surgical; they don't change any control flow that was working.
- The two commits are atomic — each can be reverted independently if needed.

---
85% AI / 15% Human
Claude: bug analysis (asar/JS spelunking, regex root cause, node-pty manifest investigation), patch authoring, build/test loop, commit + PR drafting
Human: ran the rebuilds, ran sudo installs, smoke-tested Cowork & terminal in the UI, decided strategy

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1432,25 +1432,29 @@ if (serviceErrorIdx !== -1) {
             const regionStart = Math.max(0, anchorIdx - 1000);
             const region = code.substring(regionStart, anchorIdx);
 
+            // JS identifier may start with $, _, or letter; \w doesn't
+            // match $ so use [$\w]+ to capture vars like `$e` (Claude
+            // >= 1.3109.0 uses $e for the fs module to avoid collision
+            // with the parameter `e`). See issue #418.
             // path var: VAR.join(process.resourcesPath,
             const pathMatch = region.match(
-                /(\w+)\.join\(\s*process\.resourcesPath\s*,/
+                /([$\w]+)\.join\(\s*process\.resourcesPath\s*,/
             );
             // fs var: VAR.existsSync(
-            const fsMatch = region.match(/(\w+)\.existsSync\(/);
+            const fsMatch = region.match(/([$\w]+)\.existsSync\(/);
             // logger var: VAR.info("[VM:start]
             const logMatch = region.match(
-                /(\w+)\.info\(\s*[`"]\[VM:start\]/
+                /([$\w]+)\.info\(\s*[`"]\[VM:start\]/
             );
             // stream/pipeline var: VAR.pipeline(
-            const streamMatch = region.match(/(\w+)\.pipeline\(/);
+            const streamMatch = region.match(/([$\w]+)\.pipeline\(/);
             // arch function: const VAR=FUNC(), used in smol-bin
             const archMatch = region.match(
-                /const\s+(\w+)\s*=\s*(\w+)\(\)\s*,\s*\w+\s*=\s*\w+\.join/
+                /const\s+([$\w]+)\s*=\s*([$\w]+)\(\)\s*,\s*[$\w]+\s*=\s*[$\w]+\.join/
             );
             // bundlePath var: PATH.join(VAR,"smol-bin.vhdx")
             const bundleMatch = region.match(
-                /\.join\(\s*(\w+)\s*,\s*"smol-bin\.vhdx"\s*\)/
+                /\.join\(\s*([$\w]+)\s*,\s*"smol-bin\.vhdx"\s*\)/
             );
 
             if (pathMatch && fsMatch && logMatch &&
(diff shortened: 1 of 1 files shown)


shawnvanderveer:
Thanks, worked like a charm Ubuntu 25.10

aaddrick (approved):
Hey Joost-Maker, this is solid work. Both fixes are clean, the root cause analysis in the PR description is thorough, and the commit messages are some of the best I've seen on this repo. Approving and merging.

The `$e` capture issue was the #1 reported problem this week across six separate issues (#413, #415, #417, #418, #423, #409). Widening to `[$\w]+` is the right call and it's a strict superset, so older versions where the var is just `e` still work fine.

The node-pty asar manifest fix is the kind of thing that's easy to miss. The binaries were on disk but the manifest didn't know about them, so Electron's redirect never fired. Staging `build/` and adding `--unpack '**/*.node'` is the correct approach. 

A couple minor things I'll track as follow-ups (not blocking this merge):

1. The brace-counting strip for future upstream Linux blocks is a nice defensive addition. One edge case: it doesn't track whether braces appear inside string literals or template strings. Probably fine for minified output where that's rare, but worth a comment in the code so future readers know the limitation.

2. The old manual copy-to-unpacked step in `finalize_app_asar()` is now redundant since `--unpack` handles it. Harmless for now as you noted, but we can clean it up in a follow-up once we verify the Nix flow doesn't depend on it.

I also want to give a shout-out to @RayCharlizard (#414) and @Smilez1985 (#419) who both jumped on the regex fix quickly. Having three independent contributors identify and fix the same root cause within hours of each other is great to see. I'll close #414 and #419 referencing this PR since it bundles the additional node-pty fix and has end-to-end user confirmation.

Thanks for the quick turnaround on this. Merging now.

---

PR #434, opened 2026-04-19
fix: diagnose AppArmor userns block on bwrap probe (#351)

### LINKED ISSUE #351: Using Claude Cowork on Ubuntu 24.04
In recent versions, the VM that Claude Cowork used is backed by [bubblewrap](https://github.com/containers/bubblewrap).

However, since Ubuntu 24.04 + AppArmor blocks unprivileged user namespaces (`apparmor_restrict_unprivileged_userns=1`), so the VM cannot start correctly.

The full picture is:
 1. The installed `cowork-vm-service.js` detects bwrap using `bwrap --ro-bind / / true` — which fails because Ubuntu 24.04 + AppArmor blocks unprivileged user namespaces (`apparmor_restrict_unprivileged_userns=1`)
  2. Detection fails → falls back to KVM
  3. KVM starts QEMU but the guest agent never gets its config → times out
  4. Loop

Currently I fix this issue by permitting user namespaces.

```
# Add AppArmor rules
sudo tee /etc/apparmor.d/bwrap << 'EOF'
  abi <abi/4.0>,
  include <tunables/global>

  profile bwrap /usr/bin/bwrap flags=(unconfined) {
    userns,

    include if exists <local/bwrap>
  }
EOF

# Then reload it:
sudo apparmor_parser -r /etc/apparmor.d/bwrap
```

I've proved that this fix the issue. But it's a system-level solution after all. Any better ways to solve this issue?

RayCharlizard:
## Summary

Fixes [#351](https://github.com/aaddrick/claude-desktop-debian/issues/351). On Ubuntu 24.04+, AppArmor ships with `apparmor_restrict_unprivileged_userns=1` by default, which blocks the unprivileged user namespace `bwrap` needs. The daemon's probe (`bwrap --ro-bind / / true`) fails, auto-detect silently falls through to KVM, and KVM then hangs waiting for a rootfs the user hasn't set up — leaving Cowork stuck in a retry loop with no actionable error in the logs.

## Changes

- **Classify the probe failure** — new `classifyBwrapProbeError()` helper in [scripts/cowork-vm-service.js](scripts/cowork-vm-service.js) sniffs stderr/message for userns-denial signals (`Operation not permitted`, AppArmor keywords, `CLONE_NEWUSER`, `CAP_SYS_ADMIN`, uid/gid-map denials). The daemon uses this to log a specific pointer to the TROUBLESHOOTING doc instead of a generic "bwrap not available".
- **Stop falling through to KVM when bwrap is installed but blocked.** When the probe fails on an installed bwrap, `detectBackend()` now drops to `HostBackend` rather than auto-selecting KVM. Users who legitimately want KVM can still set `COWORK_VM_BACKEND=kvm`. Fallthrough to KVM still happens when bwrap is not installed at all (unchanged behavior for that path).
- **`--doctor` check** — [scripts/launcher-common.sh](scripts/launcher-common.sh) now runs the same probe, surfaces the stderr, and emits the Ubuntu-24.04 remediation hint. The backend-summary line also reflects the new "stop at host" behavior instead of misleadingly implying KVM.
- **Documentation** — new `Cowork on Ubuntu 24.04+ (AppArmor Blocks User Namespaces)` section in [docs/TROUBLESHOOTING.md](docs/TROUBLESHOOTING.md) with the full AppArmor profile snippet and a security note.
- **Acknowledgment** — credited [@hfyeh](https://github.com/hfyeh) in [README.md](README.md) for the diagnosis and the working profile.
- **Tests** — new [tests/cowork-backend-detection.bats](tests/cowork-backend-detection.bats) covers nine classifier cases (five userns-style signals, three non-userns failures, null-safety).

## Test plan

- [x] `shellcheck scripts/launcher-common.sh` clean
- [x] `node -c scripts/cowork-vm-service.js` passes
- [x] Direct-run node harness: all 9 classifier cases pass
- [x] `--doctor` smoke test on working bwrap: new `[PASS] bubblewrap: sandbox probe succeeded` line, all checks green
- [x] `--doctor` smoke test with mocked blocked bwrap (auto-detect): `[FAIL] bubblewrap: sandbox probe failed` + userns diagnosis + TROUBLESHOOTING pointer; isolation line reads `host-direct (bwrap probe failed — see above)`
- [x] `--doctor` smoke test with `COWORK_VM_BACKEND=bwrap` override and mocked blocked bwrap: same diagnosis, isolation line still reflects the override (user-visible misconfiguration)
- [ ] CI green

## Scope note

Issue is in my cowork/bwrap maintenance lane per CLAUDE.local.md. Deliberately not shipping the AppArmor profile in the `.deb` `postinst` (the triage bot's suggestion) — that's a system-wide change to `/usr/bin/bwrap` that affects every bwrap user on the system, not just Claude Desktop. Doc-level remediation is the right altitude; the daemon+doctor changes here make sure the user actually gets pointed to it.

---
25% AI / 75% Human
Claude: drafted the classifier patterns, doctor wiring, docs, and bats cases; verified via direct node harness and mock-bwrap `--doctor` runs
Human: scoped the approach, decided against postinst profile shipping, reviewed boundary between diagnosis and silent fall-through

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,6 +231,7 @@ Special thanks to:
   - Reporting the GNOME quit accessibility issue
   - Confirming tray behavior with AppIndicator
 - **[martin152](https://github.com/martin152)** for detailed diagnosis and a complete patch for three launcher cleanup bugs: `cleanup_orphaned_cowork_daemon` self-match, `cleanup_stale_cowork_socket` socat dependency no-op, and the same self-match in `--doctor`
+- **[hfyeh](https://github.com/hfyeh)** for diagnosing the Ubuntu 24.04 AppArmor unprivileged-userns block on Cowork bwrap and contributing the AppArmor profile workaround
 
 ## Sponsorship
 
--- a/docs/TROUBLESHOOTING.md
+++ b/docs/TROUBLESHOOTING.md
@@ -89,6 +89,51 @@ For enhanced security, consider:
 - Running the AppImage within a separate sandbox (e.g., bubblewrap)
 - Using Gear Lever's integrated AppImage management for better isolation
 
+### Cowork on Ubuntu 24.04+ (AppArmor Blocks User Namespaces)
+
+Ubuntu 24.04 ships with `apparmor_restrict_unprivileged_userns=1`
+by default, which blocks the unprivileged user namespaces that
+Cowork's bubblewrap sandbox relies on. Symptoms:
+
+- `claude-desktop --doctor` reports `bubblewrap: sandbox probe failed`
+  with `Operation not permitted` in stderr.
+- `~/.config/Claude/logs/cowork_vm_daemon.log` contains
+  `bwrap is installed but cannot create a user namespace`.
+- Cowork sessions hang at "Starting VM..." or loop on reconnect.
+
+Permit user namespaces for `bwrap` via an AppArmor profile (one-time
+setup, requires sudo):
+
+```bash
+sudo tee /etc/apparmor.d/bwrap <<'EOF'
+abi <abi/4.0>,
+include <tunables/global>
+
+profile bwrap /usr/bin/bwrap flags=(unconfined) {
+    userns,
+
+    include if exists <local/bwrap>
+}
+EOF
+
+sudo apparmor_parser -r /etc/apparmor.d/bwrap
+```
+
+After applying the profile, run `claude-desktop --doctor` — the
+bubblewrap probe should pass, and Cowork should start without
+falling back to host-direct.
+
+**Security note:** this grants `/usr/bin/bwrap` the unconfined
+profile plus the `userns` capability. It matches the behavior
+bwrap had on Ubuntu 22.04 and earlier, and on most other distros,
+but is a system-wide change that affects every program invoking
+`/usr/bin/bwrap` (not just Claude Desktop). Review the profile
+against your threat model before applying.
+
+Credit: this workaround was contributed by
+[@hfyeh](https://github.com/hfyeh) in
+[#351](https://github.com/aaddrick/claude-desktop-debian/issues/351).
+
 ### Authentication Errors (401)
 
 If you encounter recurring "API Error: 401" messages after periods of inactivity, the cached OAuth token may need to be cleared. This is an upstream application issue reported in [#156](https://github.com/aaddrick/claude-desktop-debian/issues/156).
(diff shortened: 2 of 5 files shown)


slovdahl:
Quoting https://github.com/ocaml/opam/issues/5968#issuecomment-2411439351

> Please don't create an "unconfined" profile for bwrap. This completely bypasses the [unprivileged user namespace creation restriction](https://gitlab.com/apparmor/apparmor/-/wikis/unprivileged_userns_restriction) and makes your machine exploitable by attacks that use it.
>
> You should instead create a profile for the application that uses bwrap instead. We have already created one for [opam](https://gitlab.com/apparmor/apparmor/-/blob/master/profiles/apparmor.d/opam?ref_type=heads)
which is already shipped by the apparmor package.
>
> AppArmor upstream has also created a profile for bwrap that does not make your machine vulnerable. You can find the profile [here](https://gitlab.com/apparmor/apparmor/-/blob/master/profiles/apparmor/profiles/extras/bwrap-userns-restrict?ref_type=heads). It is not enabled by default on Ubuntu, but you can enable it by either copying the file from gitlab, installing it on /etc/apparmor.d/, and load it using `sudo apparmor_parser /etc/apparmor.d/bwrap-userns-restrict`.
>
> You can also use the version shipped by the apparmor-profile package, and create a symbolic link from the extra-profiles directory to /etc/apparmor.d/
`sudo ln -s /usr/share/apparmor/extra-profiles/bwrap-userns-restrict /etc/apparmor.d/`
and load it
`sudo apparmor_parser /etc/apparmor.d/bwrap-userns-restrict`.

slovdahl:
Never mind, the above doesn't work out of the box.

RayCharlizard:
Thanks for flagging this. Both comments are useful and we'll be acting on the feedback.

You're right that attaching `flags=(unconfined)` to `/usr/bin/bwrap` widens the userns grant beyond just our usage. The follow-up is a helpful caveat too: the upstream `bwrap-userns-restrict` profile was actually shipped briefly in Ubuntu noble and then reverted because it broke Flatpak file saves ([LP #2072811](https://bugs.launchpad.net/ubuntu/+source/apparmor/+bug/2072811)), so we can't just point users at it as-is.

The pattern that fits us better is per-application, same shape opam and apptainer use, but attached to a Claude-owned binary path rather than to bwrap. I've opened #542 to work through the exact profile shape with @aaddrick before we update the docs or change packaging. Will credit your feedback there.

---

PR #435, opened 2026-04-19
fix: relax installPlugin gate on Linux for remote marketplaces (#396)

### LINKED ISSUE #396: Plugin installation fails for all "Anthropic & Partners" plugins — knowledge-work-plugins marketplace not available
## Description

All plugins shown in the Directory under the "Anthropic & Partners" tab (Productivity, Design, Marketing, Data, Engineering, Finance, etc.) fail to install with the error "Failed to install plugin."

## Root cause (from logs)

The app correctly identifies these plugins belong to the `knowledge-work-plugins` marketplace, but skips the remote API install path because the current environment is detected as `not-cowork`. It then falls back to the CLI installer, which fails because the `knowledge-work-plugins` marketplace doesn't exist locally.

```
[CustomPlugins] installPlugin: skipping remote API path for plugin 
    productivity@knowledge-work-plugins (reason=not-cowork)
[CustomPlugins] installPlugin: attempting CLI install for plugin 
    productivity@knowledge-work-plugins (scope=default)
[CCDMarketplacePluginManagerCLI] Installing plugin: 
    productivity@knowledge-work-plugins
Failed to install plugin "productivity@knowledge-work-plugins": 
    Plugin "productivity" not found in marketplace "knowledge-work-plugins"
```

The three locally configured marketplaces (`claude-plugins-official`, `anthropic-agent-skills`, `impeccable`) do not include these plugins, so the CLI fallback always fails.

## Steps to reproduce

1. Open Claude Desktop
2. Go to Directory → Plugins → "Anthropic & Partners" tab
3. Click the `+` button on any plugin (e.g. Productivity, Design)
4. Observe "Failed to install plugin." toast

## Expected behavior

Plugin installs successfully, or a clear message indicating this feature is not yet supported on the Linux build.

## Environment

- **OS:** Fedora 44 (Workstation Edition), x86_64
- **Claude Desktop:** 1.1.7714-1.3.23.fc42 (RPM)
- **Claude Code CLI:** 2.1.104
- **Log file:** `~/.config/Claude/logs/main.log`

RayCharlizard:
Closes #396.

## Summary

The Directory → "Anthropic & Partners" tab lists plugins from the `knowledge-work-plugins` marketplace, which isn't set up as a local marketplace. Three client-side gates in `installPlugin` combine to block the remote-API path:

```js
if (!a && (c || (s?.mode)==="cowork") && await A0()) {
  // remote API install
}
```

- `!a` — plugin is not local-sourced
- `c` — `pluginSource === "remote"` (not set for this tab)
- `mode === "cowork"` — not on a normal Linux install
- `A0()` — sparkplug feature flag (`isFeatureEnabled("2340532315")`)

None of `c`, `mode==="cowork"`, or `A0()` hold for a normal Linux user clicking one of those plugins, so it falls through to the CLI installer, which fails with `Plugin "..." not found in marketplace "knowledge-work-plugins"` (the marketplace isn't configured locally).

The backend (`/api/organizations/{orgId}/plugins/...`) isn't platform- or cowork-gated — any Pro/Teams account should reach it — so the extra client-side gates only hurt on Linux.

## Patch

Adds **Patch 11** to `patch_cowork_linux()` in `build.sh`. It wraps the original gate so Linux short-circuits to the remote-API path for any non-local plugin:

```js
if (!a && (process.platform === "linux" || (c || mode === "cowork") && await A0())) {
  // remote API install — tried on Linux for all non-local plugins
}
```

- Anchored on the unique log string `[CustomPlugins] installPlugin: attempting remote API install`.
- Variable names are extracted dynamically via regex (minifier drift handled — `[$\w]+` covers `$e`-style idents like in patch 9).
- Idempotent via `process.platform==="linux"||` substring check.
- Prints a warning instead of failing if the anchor/pattern changes upstream (non-fatal — core cowork patches are gated on their own anchors).

The `else` branch's `reason` expression still references `c`; after this patch the else only fires when `a` (local) is true, so the reason is always `local-sourced`. The other branches become dead code but don't error.

Users without an org still fall through to the CLI fallback — that's expected behavior, since `knowledge-work-plugins` requires a Pro/Teams account.

macOS/Windows behavior is unchanged, though since this build only targets Linux that's theoretical.

## Verification

- [x] shellcheck passes on `build.sh`
- [x] Regex verified against current minified `app.asar` (Claude Desktop 1.3109.0) — gate pattern matches exactly once in `index.js`
- [x] Replacement produces syntactically valid JS (dry-run printed correct output)
- [ ] End-to-end install test — needs a Pro/Teams login and a fresh build. Happy to verify on request before merge.

## Review request

@aaddrick — this touches `build.sh` patch logic against minified JS, which is outside my cowork/bwrap/KVM maintainer scope (per CLAUDE.local.md). **Holding merge for your review** even though the issue is `cowork`-labeled. Let me know if you'd rather take the patch yourself or want me to adjust the approach.

---
30% AI / 70% Human
Claude: wrote the node patch, regex, dry-run verification, and PR body. Human: directed scope decision (tag aaddrick, don't self-merge), reviewed patch strategy.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1698,6 +1698,67 @@ if (serviceErrorIdx !== -1) {
     }
 }
 
+// ============================================================
+// Patch 11: Relax installPlugin gate for Linux (#396)
+// Anchor: unique string
+//   "[CustomPlugins] installPlugin: attempting remote API install"
+//
+// The Directory "Anthropic & Partners" tab lists plugins from
+// the knowledge-work-plugins marketplace, which isn't configured
+// as a local marketplace. The app only tries the remote-API
+// install path when (pluginSource==="remote" OR mode==="cowork")
+// AND a sparkplug feature flag is enabled — none of which hold
+// on a normal Linux install, so it falls back to the CLI
+// installer that fails with "marketplace not found".
+//
+// The backend (/api/organizations/{orgId}/plugins/...) is not
+// gated by platform or cowork mode — any Pro/Teams account
+// should reach it. Short-circuit the extra client-side gates on
+// Linux for non-local plugins by wrapping the original gate in
+// (process.platform==="linux" || <original>). Users without an
+// org still fall through to the CLI fallback (expected — the
+// knowledge-work-plugins marketplace requires a Pro/Teams
+// account).
+// ============================================================
+{
+    const installAnchor =
+        '[CustomPlugins] installPlugin: attempting remote API install';
+    const installAnchorIdx = code.indexOf(installAnchor);
+    if (installAnchorIdx !== -1) {
+        const searchStart = Math.max(0, installAnchorIdx - 300);
+        const beforeAnchor = code.substring(searchStart, installAnchorIdx);
+        if (beforeAnchor.includes('process.platform==="linux"||')) {
+            console.log('  installPlugin gate relaxation already applied');
+        } else {
+            // Gate pattern:
+            //   if(!LOCALVAR&&(REMOTEVAR||(...).mode==="cowork")
+            //     &&await FLAGFN())
+            const gateRe = /if\((!\w+)&&(\(\w+\|\|\([$\w]+==null\?void 0:[$\w]+\.mode\)==="cowork"\)&&await [$\w]+\(\))\)/;
+            const gateMatch = beforeAnchor.match(gateRe);
+            if (gateMatch) {
+                const [whole, localCheck, innerGate] = gateMatch;
+                const gateAbsIdx = searchStart + gateMatch.index;
+                const replacement =
+                    'if(' + localCheck +
+                    '&&(process.platform==="linux"||' +
+                    innerGate + '))';
+                code = code.substring(0, gateAbsIdx) +
+                    replacement +
+                    code.substring(gateAbsIdx + whole.length);
+                console.log(
+                    '  Relaxed installPlugin gate for Linux (#396)');
+                patchCount++;
+            } else {
+                console.log(
+                    '  WARNING: Could not find installPlugin gate (#396)');
+            }
+        }
+    } else {
+        console.log(
+            '  WARNING: installPlugin anchor not found (#396)');
+    }
+}
+
 fs.writeFileSync(indexJs, code);
 console.log(`  Applied ${patchCount} cowork patches`);
 if (patchCount < 5) {


build.sh
```diff
@@ -1698,6 +1698,67 @@ if (serviceErrorIdx !== -1) {
...
+    if (installAnchorIdx !== -1) {
+        const searchStart = Math.max(0, installAnchorIdx - 300);
+        const beforeAnchor = code.substring(searchStart, installAnchorIdx);
+        if (beforeAnchor.includes('process.platform==="linux"||')) {
+            console.log('  installPlugin gate relaxation already applied');
+        } else {
+            // Gate pattern:
+            //   if(!LOCALVAR&&(REMOTEVAR||(...).mode==="cowork")
+            //     &&await FLAGFN())
+            const gateRe = /if\((!\w+)&&(\(\w+\|\|\([$\w]+==null\?void 0:[$\w]+\.mode\)==="cowork"\)&&await [$\w]+\(\))\)/;
```
afurm, line 1736: Line 1737: `const gateMatch = beforeAnchor.match(gateRe);` — If the regex pattern `gateRe` fails to match the bundled code (e.g., due to a code change in a newer version of the bundled file), `gateMatch` is null. The else branch (line 1751-1754) only logs a console warning and silently skips the patch. This means the underlying Linux bug is NOT fixed, but users have no indication anything went wrong — the code silently continues with the original broken behavior. Consider failing explicitly: throw an error, set a `patchFailed` flag, or at minimum use `console.error` instead of `console.log` so it surfaces in error logs.

RayCharlizard: Thanks for the review! The log-warning-not-throw pattern is the established convention across all 20+ sibling patches in `patch_cowork_linux()` — only Patch 1 uses `console.error` + `process.exit(1)`, and that's because the app crashes at startup without it (`FATAL: Failed to patch cowork platform gate for Linux. The app will crash at startup without this patch.` at [build.sh:1149-1153](https://github.com/aaddrick/claude-desktop-debian/blob/main/build.sh#L1149-L1153)).

Patch 11 is non-fatal by comparison: if the regex drifts past a future Claude Desktop update, plugin install still works for any locally-configured marketplace (`claude-plugins-official`, `anthropic-agent-skills`, `impeccable`) — only the remote-API shortcut for `knowledge-work-plugins` stays broken, which matches the current (pre-fix) behavior. Throwing here would turn a graceful degradation into a hard build failure, which would block the whole APT/DNF/AppImage release pipeline.

Converting only Patch 11's miss to `console.error`/throw would also make the warning surface inconsistent with the other ~20 `'  WARNING: ...'` logs (e.g. Patches 4b, 6, 6b, 8, 9, 10). If there's appetite for promoting the whole set to `console.error` so build-CI flags them more visibly, that's a reasonable project-wide convention change — I'd defer to @aaddrick on whether to do that here or in a separate PR.

Leaving as-is unless @aaddrick wants otherwise.

aaddrick:
# Closing as obsolete — upstream fix supersedes

After full investigation, the bug reported in #396 was real on Claude Desktop **1.1.7714** but has been fixed upstream by Anthropic between **1.1.7714** and the current **1.3109.0** — through a coordinated client gate relaxation and renderer behavior change. This PR's client-side Linux short-circuit would have been correct for the bug as it existed at filing time, but is a no-op on current upstream. Full methodology and findings below so this doesn't have to be re-derived next time.

Thanks @RayCharlizard for the investigation and patch — the diagnosis was correct for the bug at the time.

---

## 1. Real-world reproduction check on 1.3109.0

Installed the repo's current upstream build via `dnf` on Fedora 44 (Pro account, no patch applied), clicked install on "Productivity" in Directory → **Anthropic & Partners**.

From `~/.config/Claude/logs/main.log`:
```
2026-04-19 08:52:12 [info] [CustomPlugins] installPlugin: attempting remote API install
  for plugin design@knowledge-work-plugins (source=remote, mode=cowork)
2026-04-19 08:52:13 [info] [RemotePluginManager] Installed plugin: design
2026-04-19 08:52:13 [info] [CustomPlugins] installPlugin: remote API install succeeded
```

Works end-to-end on Linux without this patch. `source=remote, mode=cowork` — the renderer now sends both fields unconditionally for Anthropic & Partners installs on desktop.

Pre-install gates also green:
```
gate=claudeai_cowork_backend_marketplaces value=true
gate=cowork_plugin_host_ops value=true
```

## 2. Side-by-side source diff (smoking gun)

Pulled the exact Claude Desktop version Luca was on (`v1.3.23+claude1.1.7714` release, `reference-source.tar.gz`) and compared `installPlugin` and the listing filter against current 1.3109.0.

### `installPlugin` gate

**1.1.7714** (Luca's version, `index.js:230901`):
```js
if (!c && (a?.mode) === "cowork" && (await Tg())) {
  // remote API
}
// reasons: "local-sourced" | "not-cowork" | "sparkplug-disabled"
```

**1.3109.0** (current, `index.js:490853`):
```js
if (!a && (c || s?.mode === "cowork") && (await A0())) {
  // remote API
}
// reasons: "local-sourced" | "not-cowork-not-remote" | "sparkplug-disabled"
```

Old required `mode === "cowork"` specifically. New accepts `pluginSource === "remote"` **OR** `mode === "cowork"`. Luca's log `reason=not-cowork` matches the old-version string exactly — confirming they were on that code path.

### `listAvailablePlugins` filter

**1.1.7714** (`index.js:231032`):
```js
if ((s?.mode) !== "cowork" || !(await Tg())) return o;  // local only
// else merge remote
```

**1.3109.0** (`index.js:491026`):
```js
if (!(await A0())) return o;
// else merge remote
```

Old required **both** cowork mode **and** sparkplug. New only requires sparkplug.

### Root cause of #396

In Luca's 1.1.7714, `listRemotePluginsPage` (`index.js:231066`) gated only on `Tg()` (sparkplug), not on cowork mode. So the Directory **showed** remote plugins whenever the flag was on. But the install gate required `mode === "cowork"` specifically. Users browsing the Directory outside a cowork session got `pluginContext` without `mode: "cowork"` from the renderer → install gate failed → `reason=not-cowork` → CLI fallback → "marketplace not found."

**Structural bug**: plugins were visible but uninstallable unless the user was actively inside a cowork session.

### How it was fixed upstream

Two coordinated Anthropic-side changes between 1.1.7714 and 1.3109.0:

1. **Client** (shipped in the asar): relax the install gate to accept `pluginSource === "remote"` as equivalent to `mode === "cowork"`.
2. **Renderer** (the claude.ai web UI loaded in the BrowserView): start passing `pluginSource: "remote"` for installs from the Anthropic & Partners Directory, regardless of cowork session state.

Both landed by the time 1.3109.0 shipped. No local patch needed for the bug pattern in #396.

## 3. Live backend validation via main-process DevTools

To confirm the backend itself isn't platform-gated (this PR's body makes that claim), attached Chrome DevTools to the main-process Node inspector and ran a header-spoofing harness.

### Enabling DevTools

```bash
echo '{"allowDevTools": true}' > ~/.config/Claude/developer_settings.json
```
Restart the app. Help menu gains **Enable Main Process Debugger**; enabling it starts a Node inspector on `127.0.0.1:9229`. Connect via `chrome://inspect` in any Chromium browser.

Source refs:
- `allowDevTools` config: `index.js:299085`
- `developer_settings.json` path: `index.js:299089`
- Main-process debugger menu: `index.js:494282`

### Header-spoofing harness

Electron only allows **one** `onBeforeSendHeaders` listener at a time. The app's injector at `index.js:504876` adds `anthropic-client-os-platform`, `anthropic-client-version`, etc. Registering a test listener replaces it, so the harness re-implements the app's injection and adds a per-test override layer:

```js
const { app, session, net } = require('electron');

const APP_HEADERS = {
  'anthropic-client-platform': 'desktop_app',
  'anthropic-client-app': 'com.anthropic.claudefordesktop',
  'anthropic-client-version': app.getVersion(),
  'anthropic-client-os-platform': process.platform,
  'anthropic-client-os-version': process.getSystemVersion(),
  'anthropic-desktop-topbar': '1',
};

globalThis.__testOverrides = {};
globalThis.__testRemove = new Set();

session.defaultSession.webRequest.onBeforeSendHeaders(
  { urls: ['https://claude.ai/*', 'https://claude.com/*'] },
  (d, cb) => {
    const h = { ...d.requestHeaders, ...APP_HEADERS, ...globalThis.__testOverrides };
    for (const k of globalThis.__testRemove) delete h[k];
    cb({ requestHeaders: h });
  }
);

async function runTest(label, { set = {}, remove = [] } = {},
                      url = 'https://claude.ai/api/desktop/features') {
  globalThis.__testOverrides = set;
  globalThis.__testRemove = new Set(remove);
  const res = await net.fetch(url);
  const ct = res.headers.get('content-type') || '';
  const body = ct.includes('json') ? await res.json() : await res.text();
  globalThis.__testOverrides = {};
  globalThis.__testRemove = new Set();
  return { label, status: res.status, body };
}
```

### Results

All tests against a Pro account, single org `d2ef6681-…`, Linux + 1.3109.0:

| # | Override | Endpoint | Result |
|---|---|---|---|
| T1 | baseline (honest) | `/api/desktop/features` | `2340532315: true` (source=`force`, rule=`fr_1fs3nhnmm0p2398`); 122 features |
| T2 | `os-platform: darwin`, `os-version: 15.0` | same | identical result — not OS-gated |
| T3 | `os-platform: win32`, `os-version: 10.0.26100` | same | identical result — not OS-gated |
| T4 | `client-version: 1.1.7714` | same | identical result — not version-gated |
| T5 | remove both OS headers | same | identical result — rule ignores OS headers |
| T6 | baseline | `/marketplaces/list-account-marketplaces` | `{marketplaces: []}` — no user-added markets |
| T7 | baseline | `/marketplaces/list-default-marketplaces` | 1 marketplace: `"Anthropic & Partners"`, `name: "knowledge-work-plugins"`, `is_default: true`, backed by GitHub `anthropics/knowledge-work-plugins` |
| T8 | baseline | `/plugins/list-plugins?installation_preference=...` | 200, 21 plugins, all from `knowledge-work-plugins` — this is the Directory feed |
| T9 | baseline | `/plugins/by-name/productivity?marketplace_name=knowledge-work-plugins` | 200, full plugin object, `installation_preference: "available"` |
| T10 | `os-platform: darwin` | same as T9 | identical body — backend indifferent to platform |
| T11 | baseline | `/plugins/{id}/download` | 200, 32 KB zip bundle |
| T12 | `Cookie: ''` | `/plugins/list-plugins` | 403 `not_found_error: "Feature not available"` — auth required, as expected |

### Findings

1. **Sparkplug `2340532315` is NOT platform- or version-gated for this account.** The force rule `fr_1fs3nhnmm0p2398` fires on any combination of spoofed OS/version headers, and even when OS identification is stripped entirely. It's keyed on user identity.
2. **The plugin API endpoints are NOT platform-gated either.** T9 vs T10 are byte-for-byte identical responses. This validates this PR's body claim about backend behavior.
3. **The backend clearly `does` gate responses on client-identity headers overall**, as T12's "Feature not available" without cookies demonstrates. But platform gating specifically — no evidence of it, at least for Pro accounts with this force rule.
4. **The force rule is user-scoped** (specific to the test account, not "all Linux Pro users"). Whether other Linux Pro users hit a similar rule is opaque to the client and depends on Anthropic's server-side targeting config.

## 4. Where does `pluginContext.mode` come from?

Non-obvious and worth writing down: **the renderer that sets `mode` and `pluginSource` is remote**. The main window is `https://claude.ai/task/new` loaded in a BrowserView. The Anthropic & Partners tab's React code is served from claude.ai, not bundled in the asar. Only **288 KB** of JS lives locally under `.vite/renderer/main_window/assets/`, and neither `installPlugin` nor `pluginContext` appears there. The main-process IPC handler at `index.js:69706` validates the shape via `Qg()` (`index.js:61302`) and trusts whatever the renderer sends.

So:
- You **can't** patch `mode`/`pluginSource` in the client — those are decided server-side.
- The only local patch surface for this class of bug is the main-process gate in `installPlugin`.
- Which is exactly what this PR targeted — correct strategy for the 1.1.7714 bug.

## 5. Disposition

**Closing as obsolete.** The bug doesn't reproduce on current upstream regardless of platform. Users still hitting the old `reason=not-cowork` failure need to upgrade Claude Desktop, not patch the cowork module — our repo tracks upstream, so `dnf upgrade` / `apt upgrade` / `.AppImage` re-download gets them the fix.

Also closing #396 with a pointer to this comment and an upgrade action item.

---

PR #436, opened 2026-04-19
fix: forward userSelectedFolders[0] as sharedCwdPath on cowork spawn (#412)

### LINKED ISSUE #412: Upstream Electron app never sends `sharedCwdPath` on Linux backends — daemon falls back to mountMap heuristics
## Summary

The cowork-vm-service daemon's `resolveWorkDir()` ([scripts/cowork-vm-service.js:405-406](https://github.com/aaddrick/claude-desktop-debian/blob/main/scripts/cowork-vm-service.js#L405-L406)) honours a `sharedCwdPath` field in the spawn payload with highest priority over `cwd`. This field already exists in the daemon-side protocol contract — it would be the natural channel for "this is the user's chosen project root, resolved on the app side." But the patched Electron app never populates it on Linux backends. Every spawn arrives with only a guest-style `cwd=/sessions/{name}` instead, forcing the daemon to reverse-engineer the user's intent from `mountMap` heuristics.

## Why it matters

PRs #389, #392, and #411 each add a daemon-side translation/heuristic to compensate for one symptom of this gap:

- **#389** (merged) — translates `CLAUDE_COWORK_MEMORY_PATH_OVERRIDE` from `/sessions/.../mnt/.auto-memory` to the host path.
- **#392** (open) — derives cwd from the primary user mount when `cwd=/sessions/{name}` (session-root) can't be translated.
- **#411** (open) — translates guest paths embedded inside `--allowedTools` / `--disallowedTools` permission patterns.

All three live in the daemon and use `mountMap` as the source of truth. They work, but each new path-bearing field the upstream introduces is a fresh chance for the daemon to fall behind. The proper fix is on the Electron side: when the user picks a folder for a Cowork session, populate `sharedCwdPath` in the spawn payload with that folder's host path. The daemon already prefers it.

## Suggested approach

This is option 1/3 from @aaddrick's [own scoping comment](https://github.com/aaddrick/claude-desktop-debian/issues/245#issuecomment-4231998466) on #245.

A `build.sh` patch (alongside `patch_cowork_linux`) would need to:

1. Find the spawn-args assembly in the minified Electron source (`app.asar.contents/.vite/build/index.js`). The session object tracks `userSelectedFolders` and `vmProcessName` — the spawn payload is built nearby. @aaddrick's comment on #245 points at the `LocalAgentModeSessions.start` interface (~line 11684 in the version they were inspecting) as the likely anchor.
2. Inject code that reads the user's selected folder from the session state and sets it on the spawn payload as `sharedCwdPath` (a host path, not a guest path).
3. Use `grep -oP` to extract minified variable names dynamically rather than hardcoding them — they change between releases (per [CLAUDE.md](https://github.com/aaddrick/claude-desktop-debian/blob/main/CLAUDE.md#working-with-minified-javascript)).

Once that lands, the daemon-side workarounds in #389/#392/#411 become belt-and-braces — still useful as a safety net, but not the primary mechanism.

## Not a regression

The daemon-side workarounds keep the symptom contained for now. This issue is a research-prerequisite for a future PR, not a fresh bug report. Capturing the context here so the next contributor doesn't re-discover it from scratch.

## References

- Umbrella: #245 (Working Directory Misreported; CLAUDE.md Not Auto-Loaded)
- Related daemon-side fixes: #373, #389, #392, #411
- Code anchors: [`resolveWorkDir`](https://github.com/aaddrick/claude-desktop-debian/blob/main/scripts/cowork-vm-service.js#L398-L424), [`patch_cowork_linux`](https://github.com/aaddrick/claude-desktop-debian/blob/main/build.sh) in `build.sh`

---
Written by Claude Opus 4 via [Claude Code](https://claude.ai/code)

RayCharlizard:
## Summary

Fixes #412. The cowork-vm-service daemon already honors a `sharedCwdPath` field on the spawn IPC payload with priority over `cwd` (see [`resolveWorkDir`](https://github.com/aaddrick/claude-desktop-debian/blob/main/scripts/cowork-vm-service.js#L500) and [`_prepareSpawn`](https://github.com/aaddrick/claude-desktop-debian/blob/main/scripts/cowork-vm-service.js#L923)), but the upstream Electron app never populates it on Linux backends. Every spawn arrives with only `cwd=/sessions/{name}`, forcing the daemon to derive the host path from mountMap heuristics (PRs #389 / #392 / #411 patch the symptoms on the daemon side).

This PR adds **Patch 12** to `patch_cowork_linux` in `build.sh`, threading the user-selected folder's host path through three call sites so the daemon receives it explicitly. The daemon-side workarounds remain as safety nets.

## What it patches (v1.3109.0)

Three coordinated edits in `app.asar.contents/.vite/build/index.js`:

- **12a — config assembly.** At `this.getVMSpawnFunction({...})` inside `LocalAgentModeSessionManager`, inject `sharedCwdPath: SESSION.userSelectedFolders?.[0]` alongside the existing `additionalMounts`/`mountSkeletonHome`/etc. The session-var name is extracted dynamically by grepping `VAR.userSelectedFolders` in the enclosing function scope.
- **12b — Kyr → VMClient.spawn().** Append `SESSION.sharedCwdPath` as a new 13th positional argument. Anchored on `,VAR.mountConda)` which is unique to the 12-arg cowork call (shorter 10-arg one-shot paths don't carry `mountConda`).
- **12c — spawn() method body.** Add a new trailing parameter and set it on the IPC payload with `VAR && (I.sharedCwdPath = VAR)` matching the existing setter chain. The parameter letter is picked dynamically from the unused-letter pool so it stays release-agnostic.

All three sub-patches detect prior application and no-op on re-run.

## Verification

Built the node heredoc into a standalone script and ran it against `index.js` extracted from `Claude-Setup-x64.exe` for v1.3109.0:

- All three patches apply; `grep -c sharedCwdPath` goes from 0 to 3.
- `node --check` passes on the patched file.
- Second run logs "already forwards sharedCwdPath" on all three sites and produces a byte-identical file.
- `shellcheck build.sh` clean.

Patch log on a fresh build:
```
  Injected sharedCwdPath into spawn config (var: t)
  Extended spawn() with h -> I.sharedCwdPath setter
  Forwarded sharedCwdPath in Kyr->spawn call (var: s)
```

## Test plan

- [ ] CI build passes on this branch
- [ ] Local AppImage build: check `Applied N cowork patches` increments by 3
- [ ] Launch AppImage, start a Cowork session with a user folder, confirm daemon log shows `sharedCwdPath=<host path>` and `resolveWorkDir` resolves to that path (not a mountMap fallback)
- [ ] Regression check: confirm PRs #389, #392, #411 still function as safety nets by temporarily forcing one of the sub-patches to fail

## Risks & fallbacks

- **Any anchor misses on a future upstream** → cowork still works via daemon's #392 fallback; console warnings flag the failure in the build log.
- **`userSelectedFolders` empty/undefined** → `?.[0]` yields `undefined`, the setter `h && (I.sharedCwdPath = h)` skips; daemon falls back to today's behavior.
- **User picks multiple folders** → only `[0]` is forwarded; the rest stay in `additionalMounts`, matching the daemon's single-path expectation for `resolveWorkDir`.

## Scope note

This touches a minified-JS patch (normally out-of-scope for the cowork lane per my scoping), but issue #412 is labeled `cowork` and is fundamentally about the cowork IPC contract. Flagging for visibility — happy to defer merging until you've had a look, since it lands alongside my other open cowork PRs.

---
40% AI / 60% Human
Claude: analyzed the extracted v1.3109.0 source, traced the three patch sites, drafted the node-based patch block, ran a dry-run verifying idempotency and syntax validity, authored the commit/PR
Human: routed the task, course-corrected the approach (flagged prior -with-fixes memory, held at plan-mode for review of the three-site approach, approved the branch/PR handoff)

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1726,27 +1726,26 @@ if (serviceErrorIdx !== -1) {
             console.log('  #412 sharedCwdPath already in spawn config');
             site1Done = true;
         } else {
-            // Find session-var name from VAR.userSelectedFolders in the
-            // enclosing function scope before the anchor.
-            const before = code.substring(Math.max(0, cfgIdx - 8000), cfgIdx);
-            const usfMatches = [...before.matchAll(
-                /([$\w]+)\.userSelectedFolders/g)];
-            const sessionVar = usfMatches.length
-                ? usfMatches[usfMatches.length - 1][1]
-                : null;
-            if (!sessionVar) {
-                console.log('  WARNING: #412 no VAR.userSelectedFolders' +
-                    ' near config');
+            // The session-id var is the value of the first field
+            // 'sessionId:VAR' in the config itself — cheap, scoped, and
+            // immune to unrelated *.userSelectedFolders references (e.g.
+            // loop variables) that wander into the enclosing scope.
+            const sidMatch = cfgBlock.match(/\{sessionId:(\w+)\b/);
+            if (!sidMatch) {
+                console.log('  WARNING: #412 no sessionId field in config');
             } else {
-                // Insert just before the closing '}' of the object literal.
+                const sidVar = sidMatch[1];
+                // Route through this.sessions.get() — canonical accessor
+                // the same class already uses, so the injection survives
+                // re-orderings of local vars in the enclosing function.
                 const blockStart = code.indexOf(cfgBlock, cfgIdx);
                 const insertAt = blockStart + cfgBlock.length - 1;
-                const insertion = ',sharedCwdPath:' + sessionVar +
-                    '.userSelectedFolders?.[0]';
+                const insertion = ',sharedCwdPath:this.sessions.get(' +
+                    sidVar + ')?.userSelectedFolders?.[0]';
                 code = code.substring(0, insertAt) +
                     insertion + code.substring(insertAt);
                 console.log('  Injected sharedCwdPath into spawn' +
-                    ' config (var: ' + sessionVar + ')');
+                    ' config (sessionId var: ' + sidVar + ')');
                 patchCount++;
                 site1Done = true;
             }
@@ -1754,7 +1753,6 @@ if (serviceErrorIdx !== -1) {
     }
 
     // --- 12c: accept a 13th param in spawn() method body ---
-    // Done before 12b so the forwarded property name is fixed.
     let site3Done = false;
     const spawnIdempotent =
         /async spawn\([^)]+\)\{const \w+=\{id:[^}]+\};[^{}]*\.sharedCwdPath=/;
@@ -1797,19 +1795,25 @@ if (serviceErrorIdx !== -1) {
     }
 
     // --- 12b: forward SESSION.sharedCwdPath in Kyr -> spawn() call ---
-    // Anchor: ',VAR.mountConda)' — unique to the 12-arg caller (the
-    // shorter 10-arg one-shot call sites lack mountConda).
+    // Anchor: ',VAR.mountConda)' — expected unique to the 12-arg caller
+    // (the shorter 10-arg one-shot call sites lack mountConda). Assert
+    // the uniqueness so a second upstream caller wouldn't silently take
+    // only the first hit.
     let site2Done = false;
     if (/,\w+\.mountConda,\w+\.sharedCwdPath\)/.test(code)) {
         console.log('  #412 caller already forwards sharedCwdPath');
         site2Done = true;
     } else {
-        const callMatch = code.match(/,(\w+)\.mountConda\)/);
-        if (!callMatch) {
+        const callMatches = [...code.matchAll(/,(\w+)\.mountConda\)/g)];
+        if (callMatches.length === 0) {
             console.log('  WARNING: #412 no ",VAR.mountConda)" pattern found');
+        } else if (callMatches.length > 1) {
+            console.log('  WARNING: #412 expected 1 ",VAR.mountConda)" match,' +
+                ' found ' + callMatches.length + '; skipping to avoid' +
+                ' wrong-site forwarding');
         } else {
-            const sessionVar = callMatch[1];
-            code = code.replace(callMatch[0], ',' + sessionVar +
+            const [whole, sessionVar] = callMatches[0];
+            code = code.replace(whole, ',' + sessionVar +
                 '.mountConda,' + sessionVar + '.sharedCwdPath)');
             console.log('  Forwarded sharedCwdPath in Kyr->spawn call' +
                 ' (var: ' + sessionVar + ')');
--- a/build.sh
+++ b/build.sh
@@ -1700,26 +1700,15 @@ if (serviceErrorIdx !== -1) {
 
 // ============================================================
 // Patch 12: Forward user-selected folder as sharedCwdPath (#412)
-// The cowork-vm-service daemon already honors a sharedCwdPath
-// field on the spawn IPC payload with priority over cwd (see
-// resolveWorkDir in scripts/cowork-vm-service.js), but the
-// upstream Electron app never populates it on Linux backends.
-// Every spawn arrives with only cwd=/sessions/{name}, forcing
-// the daemon to derive the host path from mountMap heuristics
-// (PRs #389/#392/#411 cover the symptoms).
-//
-// This patch threads the user's chosen folder through three
-// sites so the daemon receives it explicitly:
-//
-//   12a. At this.getVMSpawnFunction({...}) config assembly,
-//        inject sharedCwdPath: SESSION.userSelectedFolders?.[0].
-//   12b. At the Kyr() → VMClient.spawn() call site, forward
-//        SESSION.sharedCwdPath as a new 13th positional arg.
-//   12c. In the spawn() method body, accept a new trailing
-//        parameter and set it on the IPC payload.
-//
-// If any site fails the daemon-side fallback still works
-// (the primary mount heuristic from #392 covers cwd).
+// The cowork-vm-service daemon honors a sharedCwdPath field on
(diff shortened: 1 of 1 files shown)


RayCharlizard:
@aaddrick — heads-up: this touches `build.sh` (shared file) and a minified-JS patch (your lane per the scoping). Leaving it for your review before merge; no rush. The daemon-side workarounds in #389/#392/#411 stay in place as safety nets either way.

---

aaddrick (commented):
Like this direction a lot. The sub-patch split into 12a/12b/12c is the right move, and wiring `sharedCwdPath` through so the daemon gets the real session folder instead of guessing is a clean fix for the cwd bug.

Two things I'd like to see tightened before merge, then a few follow-ups that can wait.

## What needs fixing

**12a should read from the session map, not from a backscan.** The 8000-char backscan for `VAR.userSelectedFolders` is semantically arbitrary. I ran it against the v1.3109.0 minified source and got 10 matches in the window across 4 different variable names (`t`, `se`, `Ke`, `We`). Last-match happens to be `t` (the real session var) right now, but only by coincidence. `We` in particular is a `for (const We of …)` loop variable and is one upstream re-order away from becoming the new "last."

The same upstream function already has the canonical form you want:

```js
getUserSelectedFolders: () => ((We = this.sessions.get(A)) == null ? void 0 : We.userSelectedFolders) ?? []
```

Inject `sharedCwdPath: this.sessions.get(A)?.userSelectedFolders?.[0]` directly. It's self-contained, doesn't depend on which closure variable happens to be in scope, and survives upstream re-orderings of the enclosing function.

**12b's `,VAR.mountConda)` uniqueness should be verified, not assumed.** It's unique in v1.3109.0 (I checked — 1 match), but the patch uses `code.match()` which silently returns the first hit if a second ever shows up. Swap to `matchAll`, assert `count === 1`, and WARN-and-skip on anything else. Three extra lines and the release-to-release fragility becomes detectable instead of silent.

## Nice to have (follow-ups, don't block)

- **Runtime log signal on the daemon side.** Static patching can't distinguish "Patch 12 works" from "Patch 12 is a no-op and mountMap carried us." A structured log line on every spawn — something like `cowork-spawn: workDir source=sharedCwdPath|mountMap-primary|mountMap-home` — turns the first cowork session after an upgrade into a canary. Separate PR against `scripts/cowork-vm-service.js`.
- **Shape-defensive read on the daemon side.** If upstream ever migrates `userSelectedFolders` from `string[]` to `{path, readOnly}[]`, `sharedCwdPath` becomes an object, coerces to `"[object Object]"`, fails `existsSync`, and falls through silently. Cheap to guard in `resolveWorkDir`: `typeof x === 'string' ? x : x?.path`. Whenever.
- **Drop the "12c before 12b so the property name is fixed" comment.** The property name is a hardcoded string literal in both sub-patches, so the ordering is cosmetic. The comment will confuse a future reader.

A couple of things I want to explicitly flag as non-issues so you don't spend cycles on them:

- Scope break into minified-JS is fine here. The field is cowork-semantic, the patching is cowork-adjacent, no need to split the PR.
- The daemon fallbacks from #389/#392/#411 should stay. They carry the path for users who never pick a folder, so they're not dead code post-Patch-12.

Fix these two up and merge when you're ready.

Heads up on what's coming after this lands: I'll be starting on `refactor/build-split-for-codeowners`, which breaks `build.sh` into topical modules under `scripts/` and adds a CODEOWNERS per subsystem. Should make scoping questions like the one you flagged here easier to answer next time — the minified-JS patches will live in their own file with their own owner, so a cowork-lane PR that has to touch them won't read like a scope break anymore.

RayCharlizard:
Thanks — both blockers addressed in d844037.

**12a — backscan replaced.** Confirmed your count on v1.3109.0: 10 `VAR.userSelectedFolders` hits across `t` (7), `se` (1), `Ke` (1), `We` (1). Swapped to `this.sessions.get(VAR)?.userSelectedFolders?.[0]`, extracting `VAR` from the config's own first field `{sessionId:VAR` — scoped to the block, guaranteed present, immune to loop-var leakage. Bonus: the resulting expression matches your quoted `getUserSelectedFolders` accessor exactly, so it reads as a second caller of the same canonical form rather than a parallel mechanism.

**12b — `matchAll` + `length === 1` assertion.** WARNs-and-skips on 0 or ≥2 matches. In v1.3109.0 there's exactly 1 match (`,s.mountConda)`), so behavior unchanged; a second upstream caller will now surface instead of silently taking the first hit.

**Ordering comment dropped.** You're right, the property name is a literal in both sub-patches — the note was cargo-cult. Gone.

On the follow-ups:
- **Daemon log signal** — I like the `source=sharedCwdPath|mountMap-primary|mountMap-home` idea. Will open a separate PR against `scripts/cowork-vm-service.js`. That gives upgrade canaries a clear signal for *every* Patch-12-like future work, not just this one.
- **Shape-defensive `typeof x === 'string' ? x : x?.path`** in `resolveWorkDir` — will fold into the same daemon-side PR. Cheap insurance.
- **CODEOWNERS split** — sounds great. The "am I in scope here?" friction is real; a topical split on `build.sh` + CODEOWNERS-per-subsystem would make cowork-lane minified-JS patches unambiguous.

Dry-run against 1.3109.0 post-fix: all three sub-patches apply, `node --check` passes, the three sharedCwdPath edits are byte-stable across runs (the existing Patch 9 strip-and-reinject non-idempotency is orthogonal and pre-existing).

Leaving it for your merge call — no rush.

---

aaddrick:
Merging, thanks again!

If you have cycles, try this agent out for reviewing plans, PR's, whatever. Code related or otherwise, it's been great at helping clearing out blind spots.

https://github.com/aaddrick/contrarian/blob/main/.claude/agents/contrarian.md

Issue #321, opened 2026-03-20
Orphaned processes after closing the app — no clean shutdown or quit option


gianluca-peri:
## Description

When closing the Claude Desktop window on Fedora, none of the spawned processes are terminated. Additionally, there is no "Quit" option anywhere in the application, so the user has no way to trigger a full shutdown from within the app.

This means every time a user closes the window, they unknowingly leave a full process tree consuming CPU and memory until they either log out, reboot, or manually kill the processes.

## Steps to Reproduce

1. Launch Claude Desktop on Fedora (installed via this project's build process)
2. Close the application window (click the X / close button)
3. Inspect running processes (e.g. with `htop`)
4. Observe that all processes are still running — Electron main/renderer/zygote processes, `cowork-vm-service`, and VM session processes persist as if the app were never closed.

## Expected Behavior

Closing the application should cleanly terminate all child processes, including:

- Electron main and renderer processes
- `cowork-vm-service`
- VM session processes under `claude-desktop/vm/sessions`

Alternatively, if background persistence is intentional (e.g., for keyboard shortcut responsiveness), there should be an explicit **"Quit"** action available that performs a full and clean shutdown of all related processes.

## Actual Behavior

All processes continue running indefinitely after the window is closed. There is no user-facing mechanism to fully stop the application.

## Workaround

The following shell alias can be added to `~/.bashrc` (or equivalent) to manually kill all lingering processes:

```bash
alias claude-quit='pkill -f cowork-vm-service; pkill -f "claude-desktop/vm/sessions"; pkill -9 -f "claude-desktop/node_modules/electron"'
```

This reliably cleans up all orphaned processes.

## Environment

**Distro:** Fedora 43 (Workstation Edition). **Installation method:** Installed with:

```
# Add the repository
sudo curl -fsSL https://aaddrick.github.io/claude-desktop-debian/rpm/claude-desktop.repo -o /etc/yum.repos.d/claude-desktop.repo

# Install
sudo dnf install claude-desktop
```
as per instructions in the `README`.

Issue #424, opened 2026-04-18
CTRL + C doesn't close the app and process


sagisar1:
The issue : hitting CTRL + C doesn't close the app and the process

```
claude-desktop
^C
```

nothing happens, even if I close the window, something is stuck in there.  I have to manually kill the process
this also makes problems when you change MCP server configurations, since if you don't kill the processes completly the new configuration won't be reloaded

OS : kali linux 2026.1

instasmaignan:
The fix does not solve the CTRL+A issue. Maybe I need to open another issue?

sagisar1:
@aaddrick  I updated to the latest release - 1.3561.0-2.0.0, but the issue still happens. Did you check it after commiting? (I don't know if you or claude alone implemented the fix..)

sagisar1:
@aaddrick Hi, I am in the 1.6259.0-2.0.10 release, still hasn't been fixed. Is this going to be fixed soon? It really disturbs that using CTRL+C doesn't close the app, and I have to close the terminal completely to close claude. Do you want me to submit a PR, will that help? Because this issue is already open almost 3 weeks.

aaddrick:
Hey! Apologies about the delay. Can you run `claude-desktop --doctor` and pay the results?

sagisar1:
```
claude-desktop --doctor                     
Claude Desktop Diagnostics
================================

[PASS] Installed version: 1.6259.0-2.0.10
[PASS] Display server: X11 (DISPLAY=:0.0)
       Menu bar mode: auto (default, Alt toggles visibility)
       Titlebar style: hybrid (default, native frame + in-app topbar)
[PASS] Electron: v41.5.0 (/usr/lib/claude-desktop/node_modules/electron/dist/electron)
[PASS] Chrome sandbox: permissions OK (/usr/lib/claude-desktop/node_modules/electron/dist/chrome-sandbox)
[PASS] SingletonLock: no lock file (OK)
[PASS] MCP config: valid JSON (/home/kali/.config/Claude/claude_desktop_config.json)
       MCP servers configured: 1
[PASS] Node.js: v22.22.2
       Path: /usr/bin/node
[PASS] Desktop entry: /usr/share/applications/claude-desktop.desktop
[PASS] Disk space: 39704MB free

Cowork Mode
----------------
[PASS] bubblewrap: found
[PASS] bubblewrap: sandbox probe succeeded
       KVM: not available
[PASS] vsock: module loaded
       QEMU: not found
[...]

PR #451, opened 2026-04-20
fix(lifecycle): hide main window to tray on close, Linux

### LINKED ISSUE #448: Linux: app quits when last window closed; breaks in-app schedulers and 'minimize to tray' expectation

lizthegrey:
## Summary

- On Linux, intercept `BrowserWindow.close` on main windows and `preventDefault + hide` so the app survives a stray click on X (or a sign-out flow that closes `mainWindow`). Popups (Quick Entry, About) already dismiss via `hide()` and never see close events, so they're unaffected.
- Arm `app._quittingIntentionally` in an `app.on('before-quit')` listener so the close handler still lets windows actually close when the user picks a real quit path — Ctrl+Q global shortcut (registered in this same file), tray Quit, cmd+Q, SIGTERM, or any other `app.quit()`.
- Gated by `CLOSE_TO_TRAY` (const at the top of the file, logged on startup). Default is on; `CLAUDE_QUIT_ON_CLOSE=1` restores Electron-default behaviour for users who prefer X to quit.

## Motivation

The existing tray patches keep the icon alive *while the app runs*, but Electron's default `window-all-closed → app.quit()` means the app dies the moment the last window closes. In-app schedulers (notably the `/schedule` skill) are in-process `setTimeout`s, so a missed overnight window = silently missed cron firings the next morning. My own `~/.config/Claude/logs/main.log` caught this: clean `beforeQuit` at 15:21 PDT one afternoon, no restart until I noticed ~21 hours later, and an 8am scheduled task that simply never ran.

Diagnosis and design are in #448.

## Interaction with existing patches

- **Ctrl+Q global shortcut** (already in this file, `Fixes: #321`) — unchanged; `app.quit()` → `before-quit` arms the flag → windows close normally. 
- **Tray icon** (`scripts/patches/tray.sh`) — its Quit menu item calls `app.quit()`, same flow as Ctrl+Q.
- **Popup detection** — `if (!popup)` already gates the main-window-only code path; the close listener lives there, so Quick Entry / About are untouched.
- **#321 / #424** (orphaned processes, Ctrl+C doesn't close) — different problem, different fix; this PR is about keeping the app alive deliberately rather than cleaning it up.

## Fixes

#448

## Test plan

- [x] `node --check scripts/frame-fix-wrapper.js` clean.
- [x] `./build.sh --build deb --clean no`.
- [x] Launch, close main window via X → window disappears, tray icon remains, `pgrep -f 'app\\.asar'` still shows a live non-helper Electron PID.
- [x] Tray → Show App → window returns with preserved state.
- [x] Tray → Quit → `beforeQuit` fires in `main.log`, all Electron / `cowork-vm-service` PIDs exit.
- [x] Ctrl+Q while window is hidden but mini question bar is visible → app fully quits.
- [x] Ctrl+Q while window is visible → app fully quits.
- [ ] Schedule a near-future `/schedule` task, close main window, wait → task fires on time.
- [ ] Launch with `CLAUDE_QUIT_ON_CLOSE=1`, close window → app exits (env escape hatch works).
- [x] Quick Entry (Ctrl+Alt+Space) still dismisses and re-opens correctly; no hidden-window accumulation in `BrowserWindow.getAllWindows()`.
- [ ] `win.webContents.close()` / signout flow (if exercised via the UI) doesn't leave the app in a weird half-state.

## Known trade-offs worth calling out in review

1. **Default on vs off** — I made this opt-out (default on) because the Electron default is actively harmful for this app (in-process schedulers, tray is otherwise useless, missed crons). If you'd rather ship opt-in first and flip the default later, say the word and I'll reverse the gating.
2. **Programmatic `win.close()`** — if the app itself calls `mainWindow.close()` for a non-quit reason (e.g., a hypothetical reset flow), our handler will hide instead of close. I don't *think* the current app does this outside of the real quit path (all the upstream call sites I've seen go through `app.quit()`), but worth watching for during test.
3. **Autostart interaction** — #128 (separate PR #450) lands the `~/.config/autostart/` integration. With both merged, a user can enable autostart *and* have the app survive stray closes, which is the combination needed for reliable overnight `/schedule` runs. Neither PR depends on the other.

---

75% AI / 25% Human
Claude: pulled log evidence from `~/.config/Claude/logs`, worked out the close/quit event ordering, drafted the wrapper patch and commit/PR copy, ran `node --check`.
Human: reported the missed-cron symptom that surfaced the bug, made the opt-out-vs-opt-in default call, and will run the manual test plan / own the merge decision.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,7 +181,9 @@ Special thanks to:
   - Quick window submit fix
 - **[janfrederik](https://github.com/janfrederik)** for the `--exe` flag to use a local installer
 - **[MrEdwards007](https://github.com/MrEdwards007)** for discovering the OAuth token cache fix
-- **[lizthegrey](https://github.com/lizthegrey)** for version update contributions
+- **[lizthegrey](https://github.com/lizthegrey)**
+  - Version update contributions
+  - Close-to-tray on Linux to keep in-app schedulers, MCP servers, and the tray icon alive across window close
 - **[mathys-lopinto](https://github.com/mathys-lopinto)**
   - AUR package
   - Automated deployment
--- a/scripts/frame-fix-wrapper.js
+++ b/scripts/frame-fix-wrapper.js
@@ -38,6 +38,18 @@ if (resolvedMode !== rawMenuBarMode) {
 }
 console.log(`[Frame Fix] Menu bar mode: ${MENU_BAR_MODE}`);
 
+// Keep the app alive when the main window is closed (hide to tray),
+// so in-app schedulers / MCP servers / the tray icon survive a
+// stray click on X. Explicit quit paths (Ctrl+Q global shortcut
+// registered below, tray menu Quit, File > Quit, cmd+Q, SIGTERM)
+// still go through app.quit() → before-quit, which arms the flag
+// so the close handler lets the windows actually close.
+// Set CLAUDE_QUIT_ON_CLOSE=1 to restore the Electron-default
+// "closing the last window quits the app" behaviour.
+const CLOSE_TO_TRAY = process.platform === 'linux'
+  && process.env.CLAUDE_QUIT_ON_CLOSE !== '1';
+console.log(`[Frame Fix] Close-to-tray: ${CLOSE_TO_TRAY ? 'on' : 'off'}`);
+
 // Detect if a window intends to be frameless (popup/Quick Entry/About)
 // Quick Entry: titleBarStyle:"", skipTaskbar:true, transparent:true, resizable:false
 // About:       titleBarStyle:"", skipTaskbar:true, resizable:false
@@ -142,6 +154,22 @@ Module.prototype.require = function(id) {
             }
 
             if (!popup) {
+              // Close-to-tray: intercept close on main windows and hide
+              // instead. app.on('before-quit') below sets the flag when
+              // the user picks an explicit quit path, so real quits still
+              // let the window actually close. Popups (Quick Entry,
+              // About) already dismiss via hide() in the upstream code;
+              // they never see close events, so they're unaffected.
+              // Fixes: #448
+              if (CLOSE_TO_TRAY) {
+                this.on('close', (e) => {
+                  if (!result.app._quittingIntentionally && !this.isDestroyed()) {
+                    e.preventDefault();
+                    this.hide();
+                  }
+                });
+              }
+
               // Directly set child view bounds to match content size.
               // This bypasses Chromium's stale LayoutManagerBase cache
               // (only invalidated via _NET_WM_STATE atom changes, which
@@ -341,6 +369,17 @@ Module.prototype.require = function(id) {
         registerQuitShortcut();
       } else {
         result.app.once('ready', registerQuitShortcut);
+      }
+
+      // Arm the close-to-tray flag on every real quit path
+      // (app.quit() from Ctrl+Q, tray Quit, cmd+Q, SIGTERM). The
+      // BrowserWindow close handler above checks this flag to
+      // decide whether to hide or actually close. Harmless when
+      // CLOSE_TO_TRAY is off (the close handler is never attached).
+      if (CLOSE_TO_TRAY) {
+        result.app.on('before-quit', () => {
+          result.app._quittingIntentionally = true;
+        });
       }
 
       console.log('[Frame Fix] Patches built successfully');
--- a/scripts/frame-fix-wrapper.js
+++ b/scripts/frame-fix-wrapper.js
@@ -40,10 +40,10 @@ console.log(`[Frame Fix] Menu bar mode: ${MENU_BAR_MODE}`);
 
 // Keep the app alive when the main window is closed (hide to tray),
 // so in-app schedulers / MCP servers / the tray icon survive a
-// stray click on X. Explicit quit paths (Ctrl+Q global shortcut
-// registered below, tray menu Quit, File > Quit, cmd+Q, SIGTERM)
-// still go through app.quit() → before-quit, which arms the flag
-// so the close handler lets the windows actually close.
+// stray click on X. Explicit quit paths (Ctrl+Q via the focused
+// webContents listener above, tray menu Quit, File > Quit, cmd+Q,
+// SIGTERM) still go through app.quit() → before-quit, which arms
+// the flag so the close handler lets the windows actually close.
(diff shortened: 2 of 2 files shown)


aaddrick:
Hey Liz — one flag on the test plan after 4e2b9d7 landed.

The "Ctrl+Q while window is hidden → app fully quits" checkbox is probably a false pass now. Ctrl+Q is handled through `webContents.before-input-event`, which only fires when the webContents has keyboard focus. A hidden window can't hold focus, so the keybind should miss entirely from that state.

Could you re-run that one with the main window genuinely hidden and unfocused (not just minimized or behind another window) and update the test plan with what you actually observe? If it doesn't quit from there it's not a blocker — tray → Quit still works — but worth documenting the real behavior rather than carrying an inaccurate pass.

---

lizthegrey:
> Could you re-run that one with the main window genuinely hidden and unfocused (not just minimized or behind another window) and update the test plan with what you actually observe? If it doesn't quit from there it's not a blocker — tray → Quit still works — but worth documenting the real behavior rather than carrying an inaccurate pass.

Ah yep my bad, so I'm indeed unable to close the main window and get Claude application to be focused to then control Q it. HOWEVER, I am able to double click the Claude application icon to get it to show the "How can claude help?" mini pop-up" and control-q works to close from there indeed.

lizthegrey:
Ah I see the source of my and Claude's confusion. on OS X the claude app remains command tabbable even with no windows open, which is why we thought that was something we had to test

lizthegrey:
@aaddrick any remaining blockers to merge?

aaddrick:
Tested locally on Fedora. Built an RPM from your branch with origin/main merged in, installed it, and ran the close-to-tray flow end to end. X-click hides the window and keeps the app alive, tray persists, Show App restores window state, and Quit exits cleanly. `/schedule` survived a closed window — I set up a task, closed via X, came back after the trigger, and it had fired. `CLAUDE_QUIT_ON_CLOSE=1` exits on close as documented.

One snag: your branch was 8 days behind main and missing #484, which replaced the global Ctrl+Q with a focus-scoped `before-input-event` listener. Git's auto-merge accepted both the new listener and your old `globalShortcut.register` block, so the merged tree had two Ctrl+Q handlers. Real artifact you'd have hit on rebase. Since maintainerCanModify was on and I had the tree open, I pushed the cleanup directly rather than ping-ponging.

Three commits on your branch:

1. Merge of origin/main (brings in #484, #515, #496)
2. `fix(frame-fix-wrapper): drop superseded globalShortcut Ctrl+Q` — removes the dead block at 365-386; the focus-scoped listener at line 152 already covers the original #321 case
3. `docs(readme): credit @lizthegrey for close-to-tray` under your existing Acknowledgments entry

---

lizthegrey:
amazing. thank you so much.

aaddrick:
No worries at all! I'm working through 450 right now, so you should see that land shortly

PR #490, opened 2026-04-23
fix: copy ion-dist static assets for app:// protocol handler

### LINKED ISSUE #488: [bug]: Configure 3rd party inference modal can't be opened

```shell
[PASS] Installed version: 1.3883.0-2.0.1
[PASS] Display server: Wayland (WAYLAND_DISPLAY=wayland-1)
       Desktop: sway
            (disables global hotkeys)
[PASS] Electron: v41.2.2 (/usr/lib/claude-desktop/node_modules/electron/dist/electron)
[PASS] Node.js: v22.22.1
```

### What happened

Upon trying to configure third party inference, which is a new feature released recently for Claude Desktop, a blank window appears.

The data which is supposed to be there for further set up cannot be loaded.

### Steps to reproduce

Upon first installation, instead of signing in to Anthr/Claude, press Alt -> Developer -> Configure Third-party Inference

### Expected behavior

Documented here 

[https://support.claude.com/en/articles/14680741-install-and-configure-claude-cowork-with-third-party-platforms](https://support.claude.com/en/articles/14680741-install-and-configure-claude-cowork-with-third-party-platforms)

This is all that they wrote about this setup:
_With Developer mode enabled, go Developer → Configure third-party inference. This activates a setup UI to configure the required fields_

Config reference: https://claude.com/docs/cowork/3p/configuration

### Logs / errors

```shell
Error: net::ERR_FILE_NOT_FOUND
    at SimpleURLLoaderWrapper.<anonymous> (node:electron/js2c/browser_init:2:134327)
    at SimpleURLLoaderWrapper.emit (node:events:508:28)
(node:844544) electron: Failed to load URL: app://localhost/setup-desktop-3p with error: ERR_UNEXPECTED
[844544:0422/124308.747385:ERROR:mojo/public/cpp/bindings/lib/interface_endpoint_client.cc:748] Message 0 rejected by inter>
Error occurred in handler for '$eipc_message$_4ab9ae55-583a-4867-90be-23b2daff8899_$_claude.buddy_$_BuddyBleTransport_$_rep>
    at Session.<anonymous> (node:electron/js2c/browser_init:2:116002)
    at Session.emit (node:events:508:28)
```

### Anything else

Image

RayCharlizard:
## Summary

Fixes #488 — Third-Party Inference setup window opens blank with `ERR_UNEXPECTED` / `ERR_FILE_NOT_FOUND` on `app://localhost/setup-desktop-3p`.

The Claude Desktop app installs a custom `app://` protocol handler rooted at `process.resourcesPath/ion-dist/` (source: `VGr(uA.join(WWt(),"ion-dist"),...)` in the minified main bundle). That directory ships inside the Windows installer under `lib/net45/resources/ion-dist/` and holds the static SPA assets for internal windows — the Third-Party Inference setup is the visible case, but the same handler routes other `app://localhost/*` internal pages. Our Debian/AppImage build was never copying it out of the nupkg, so every `app://localhost/*` request fell through to the handler's `index.html` fallback, which also missed → handler promise rejection → `ERR_UNEXPECTED` at `loadURL` with `ERR_FILE_NOT_FOUND` in the underlying fetch.

## Changes

1. **`scripts/staging/cowork-resources.sh`** — `copy_cowork_resources()` now copies `ion-dist/` from `$claude_extract_dir/lib/net45/resources/ion-dist` into `$electron_resources_dest/ion-dist` via `cp -a` (matching the idiom in `scripts/staging/electron.sh:74`). Warn-and-continue on absence, mirroring the existing `smol-bin` / `cowork-plugin-shim.sh` blocks in the same file.

2. **`nix/claude-desktop.nix`** — matching stanza in `installPhase`. For Nix builds, `$electron_resources_dest` is a sentinel `nix-resources/` directory that the flake cherry-picks from (see lines 156–163 for the existing `smol-bin` / `cowork-plugin-shim.sh` pattern). Without this hunk, change 1 would drop `ion-dist/` into `nix-resources/` and the Nix build would discard it — leaving NixOS users broken. **@typedrat — `nix/claude-desktop.nix` is your lane per CODEOWNERS; happy to drop this hunk and land the bash side alone if you'd rather take the Nix change separately, just let me know.**

## Impact on package size

`ion-dist` is 84 MB uncompressed (React SPA bundle: `assets/`, `audio/`, `i18n/`, `images/`, `index.html`, `robots.txt`, `favicon.ico`). In the AppImage this compresses to roughly 42 MB of delta: `1.3561.0` AppImage was 138 MB, the `1.3883.0` build with this fix is 180 MB. Some of that is upstream growth between the two versions too. Worth noting because `.deb` and `.AppImage` will grow accordingly — but this is the minimum cost to make an advertised menu feature work, so I don't see a leaner path short of pruning `ion-dist` contents by hand (fragile).

## Test plan

- [x] Fresh `./build.sh --build appimage` against upstream `1.3883.0` — build log shows `Copied ion-dist` in the Cowork Resources section
- [x] `--appimage-extract` of the built AppImage — `usr/lib/node_modules/electron/dist/resources/ion-dist/` present with `index.html` + `assets/` + `i18n/` etc.
- [x] Live install on CachyOS daily-driver (`/opt/claude-desktop/claude-desktop.AppImage` swap), fully relaunched
- [x] `Developer → Configure Third-Party Inference…` now renders the 6-tab config UI (Connection / Sandbox & workspace / Connectors & extensions / Telemetry & updates / Usage limits / Plugins & skills / Egress Requirements) per [support.claude.com/en/articles/14680741](https://support.claude.com/en/articles/14680741)
- [x] `grep -E 'ERR_UNEXPECTED|ERR_FILE_NOT_FOUND' ~/.config/Claude/logs/main.log` — no matches on `setup-desktop-3p`
- [x] `shellcheck` on the modified file — no new findings (existing SC2148/SC2154 warnings are pre-existing, inherent to the sourced-module pattern)
- [ ] Nix build (`nix build .#claude-desktop`) — would appreciate @typedrat spot-check before merge

## Log noise worth explaining

The 3P setup window triggers these `Blocked permission check` warnings:
- `media` (twice — once with empty origin, once with `app://localhost/`)
- `web-app-installation`
- `geolocation`
- `background-sync`

These are `setPermissionCheckHandler` deny events (not `Request` events — no user dialogs, no behavior change). The handler's allowlist (`IXe` / `uXe`) accepts only `https://claude.ai` subdomains, `file://`, and the main-window origin — `app://localhost` is intentionally excluded, which is the right security posture for a local protocol handler. The Ion SPA's framework probes these permissions for feature detection, gets `false`, moves on. Same warnings fire on `claude.ai` pages too (pre-existing since well before this change). Not introduced by this PR.

---

30% AI / 70% Human
Claude: diagnosed the missing-resource root cause by grepping the running asar, drafted the fix + Nix hunk, ran contrarian stress-test, wrote the commit / PR body
Human: approved the plan, performed the live install swap (after Claude nearly SIGKILL'd its own hosting session with the documented `pkill mount_claude` recipe — lesson captured), confirmed the UI rendered, and pushed back on dismissed log warnings until they were actually traced

### DIFF
--- a/nix/claude-desktop.nix
+++ b/nix/claude-desktop.nix
@@ -162,6 +162,14 @@ stdenvNoCC.mkDerivation {
       fi
     done
 
+    # Install ion-dist static assets (app:// protocol handler root for
+    # Third-Party Inference setup — see issue #488)
+    if [[ -d build/electron-app/nix-resources/ion-dist ]]; then
+      cp -r build/electron-app/nix-resources/ion-dist \
+        $electron_tree/resources/
+      echo "Installed cowork resource: ion-dist"
+    fi
+
     # Install locale JSON files into resources
     for locale_json in build/claude-extract/lib/net45/resources/*-*.json; do
       [[ -f "$locale_json" ]] \
--- a/scripts/staging/cowork-resources.sh
+++ b/scripts/staging/cowork-resources.sh
@@ -1,6 +1,7 @@
 #===============================================================================
-# Cowork runtime resources: plugin shim script and architecture-specific
-# smol-bin VHDX for KVM guest SDK access.
+# Cowork runtime resources: plugin shim script, architecture-specific smol-bin
+# VHDX for KVM guest SDK access, and ion-dist static assets for the app://
+# protocol handler used by the Third-Party Inference setup window.
 #
 # Sourced by: build.sh
 # Sourced globals:
@@ -39,6 +40,19 @@ copy_cowork_resources() {
 	else
 		echo "Warning: smol-bin VHDX not found at $smol_vhdx"
 		echo "KVM Cowork will rely on virtiofs for SDK access"
+	fi
+
+	# Copy ion-dist static assets. The app registers an app:// protocol
+	# handler rooted at process.resourcesPath/ion-dist; without these
+	# assets, the Third-Party Inference setup window fails to load with
+	# ERR_UNEXPECTED (see issue #488).
+	local ion_dist_src="$resources_src/ion-dist"
+	if [[ -d $ion_dist_src ]]; then
+		cp -a "$ion_dist_src" "$electron_resources_dest/ion-dist"
+		echo 'Copied ion-dist'
+	else
+		echo "Warning: ion-dist not found at $ion_dist_src" \
+			'— Third-Party Inference setup will fail'
 	fi
 
 	section_footer 'Cowork Resources'


typedrat:
Am spot-checking this right now, looks good superficially.

typedrat (approved):
No regressions at least, but I am not getting the option to open the dialog that was causing the issue this is supposed to close so I can't confirm that it works.

aaddrick:
building now, will get back to you in a few minutes

aaddrick:
image

Works for me without issue on Nobara KDE. Merging.

RayCharlizard:
> No regressions at least, but I am not getting the option to open the dialog that was causing the issue this is supposed to close so I can't confirm that it works.

If you hit the Alt key, then go to Help -> Troubleshooting there should be an option to enable developer mode. After restarting you'll have a new Development menu in the toolbar. The menu should be there!

aaddrick:
Hey @RayCharlizard — post-merge update for you, since you called this one in advance.

You flagged in the PR body that the ion-dist delta was the minimum cost to make the menu feature actually work, with no leaner path short of fragile pruning. That call landed exactly: the combined ion-dist + upstream growth in 1.3883.0 pushed the `.deb` to 129.81 MB and tripped GitHub's 100 MB per-file cap on the `gh-pages` push.

```
remote: error: File pool/main/c/claude-desktop/claude-desktop_1.3883.0-2.0.2_amd64.deb is 129.81 MB; this exceeds GitHub's file size limit of 100.00 MB
remote: error: GH001: Large files detected.
```

[Failed run](https://github.com/aaddrick/claude-desktop-debian/actions/runs/24811974733). Builds, AppImage, RPM, AUR, and the GitHub Release all shipped fine — only the APT and DNF repo updates are stuck. Users on direct downloads, AppImage, or AUR are on 1.3883.0; only `apt update` / `dnf update` users are pinned to the previous version until we unstick this.

Currently exploring on `fix/deb-size-shrink`. First iteration (curate Electron locales 55→13, `dpkg-deb -Z xz -9 --extreme`) brought the `.deb` from 130 MB → 114 MB — closer, still 14 MB over. Removing `LICENSES.chromium.html` (19 MB uncompressed text, very compressible) is in flight as the next lever; static math says it should bridge the gap, but I'll post the actual number when the repack finishes. Legal compliance for the LICENSES file can move to either a `claude-desktop-licenses` companion package or just the GitHub Release page — that's a separate decision.

Bigger picture: shrinking is a stopgap, not the architectural fix. The longer-term plan is a custom domain on the Pages site with a transparent 301 to a Cloudflare-Worker-fronted backend that doesn't carry the per-file cap. APT and DNF both follow 301s on package fetches, GPG signature validates against the same key, no user-side `sources.list` change. That's orthogonal to (and compatible with) the gh-pages-split work in `docs/gh-pages-split-plan.md`.

To be clear: this is in no way a PR-revert situation. The feature is correctly fixed and the size cost was understood going in. We just bumped into the wall the upstream + ion-dist combination was always going to find first; it would have happened the next release anyway. Thanks for the heads-up in the original PR body — saved investigation time.

---

aaddrick:
@RayCharlizard — planning convergence update.

The shrink experiment confirmed your call: best case landed at ~113 MB compressed (locale curation 55→13 + `xz -9 --extreme` + removing `LICENSES.chromium.html`), with a hard floor of ~110 MB. The big lesson — deleting 19 MB of plain text saved 0.16 MB compressed because xz had already crushed it inside the archive. Shrinking is off the table as a primary fix.

Going with: a Cloudflare Worker on a custom domain, fronting GitHub Pages metadata and 302-redirecting binary requests to GitHub Release assets (which already publish successfully every tag — `Create Release` was green in the failed run). Existing `sources.list` entries stay untouched because Pages auto-301 from `*.github.io` to the custom domain bridges the URL transparently. Modeled on Cloudflare's own `pkg.cloudflare.com` setup, with the meaningful tweak that our Worker only emits redirect responses — binary bytes never traverse Cloudflare. That sidesteps both the historical Section 2.8 worry (removed Oct 2025 anyway) and the bandwidth angle.

Tracking: #493
Draft PR with the full plan + risk table + rollback story: #494
Plan doc: [`docs/worker-apt-plan.md`](https://github.com/aaddrick/claude-desktop-debian/blob/feat/apt-worker-redirect/docs/worker-apt-plan.md) on `feat/apt-worker-redirect`

Plan went through two contrarian review rounds. Notable catches that landed in rev3:
- Strip step now gates on a positive Worker liveness probe (`curl -fsI` against the production domain) instead of an env var — destructive `find -delete` only fires when the Worker is genuinely live
- Smoke test walks the redirect chain hop-by-hop in expected order with size-match against the Releases asset, instead of substring-grepping collected `Location:` headers
- Heartbeat workflow opens a tracking issue with format-specific labels (`heartbeat-failure-deb` / `heartbeat-failure-rpm`) on failure, auto-closes on recovery — pure cron-failure surfacing isn't enough
- Bus factor is documented as accepted-risk-with-mitigations rather than fictional shared credentials

Next step is registering a domain. Once that's in place and the Worker hits a staging route, the Phase 2 container matrix (`debian:stable`, `ubuntu:lts`, `fedora:latest`, `rockylinux:9`, plus an `apt-cacher-ng`-fronted variant and an IPv6-only one) confirms `apt update && apt install` works end-to-end with the >100 MB `.deb` through the chain. After that, the strip step lands automatically (no merge required — its probe activates as soon as the production Worker is up) and v2.0.2 reaches `apt update` users.

Shrink branch (`fix/deb-size-shrink`) is parked. Buys headroom but doesn't fix the architectural cap, so not worth merging — once the Worker plan ships, there's no per-file ceiling to fight on the next upstream growth.

---

PR #496, opened 2026-04-23
fix: update visibility function regexp

### LINKED ISSUE #495: [bug]: `patch_quick_window` partially failing in the recent build
### What happened

Trying to run `./build.sh` today (resulting in `claude-desktop_1.3883.0_amd64.deb`), I get `WARNING: Could not find visibility function near pF` in logs - so quick window patch is not fully applied. Apparently, something has changed in the underlying js and the pattern no longer works.

Andrej730:
This function is now starts with `var e;` (e.g. `function aZA(){var e;return!Qt||Qt.isDestroyed()?!1:Qt.isVisible()`)

Can confirm that the change resolves https://github.com/aaddrick/claude-desktop-debian/issues/495 for me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,7 +245,9 @@ Special thanks to:
 - **[davidamacey](https://github.com/davidamacey)** for identifying and fixing the XRDP GPU compositing blank-window issue on remote desktop sessions
 - **[pb3ck](https://github.com/pb3ck)** for diagnosing the Cowork `CLAUDE_CODE_OAUTH_TOKEN` env-strip bug with a working reference diff
 - **[aJV99](https://github.com/aJV99)** for exporting `GDK_BACKEND=wayland` in native Wayland mode to fix XWayland fallback blur on HiDPI displays
-- **[Andrej730](https://github.com/Andrej730)** for the quick-window regex readability refactor (`String.raw` + `escapeRegExp` helper)
+- **[Andrej730](https://github.com/Andrej730)**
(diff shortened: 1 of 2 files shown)


aaddrick (approved):
Hey @Andrej730 - nice diagnosis on this one. You correctly traced the regex break to the upstream shape change: the visibility function body now uses optional chaining (`pe?.webContents?.isFocused()`), which the minifier lowers to a temp var assignment, so `var e;` gets hoisted to the function scope and the old `\{return!` anchor stops matching. Tested fix, clean PR.

I pushed a fixup on top (`666a335`) that makes the `var <name>(,<name>)*;` prefix optional via a non-capturing group, so the regex tolerates three shapes: no var (1.3109.0 and earlier), `var e;` (1.3883.0), and `var e,t;` if the minifier adds another temp later. Same patch behavior as yours, just shouldn't break again on the next drift cycle. Verified end-to-end on the live 1.3883.0 asar (symbols `pF` / `aZA` / `Jo`, both Quick Entry anchors patched, JS validates, idempotent), and backward-tested on 1.3109.0 (`L7A`, no var prefix) - generalized regex still finds it.

Repro on real hardware too: built the AppImage on this branch, ran on Nobara 43 / KDE Plasma 6 / Wayland. Quick entry pop-up, submit, main window reliably shows post-patch. No regressions on regular chat or window restore. Issue #495 is resolved by your PR; the fixup is just future-proofing. Approving.

---

PR #515, opened 2026-04-24
fix: update Linux tray icon in place on OS theme change

IliyaBrook:
## What's in this PR

As asked in the #491 review:

- **`patch_tray_inplace_update`** in `scripts/patches/tray.sh` — fast-path that updates the existing `StatusNotifierItem` in place via `setImage` + `setContextMenu` instead of destroy + recreate
- **`scripts/patches/app-asar.sh`** wiring — calls the new patch between `patch_tray_icon_selection` and `patch_menu_bar_default`
- **`docs/learnings/tray-rebuild-race.md`** — scoped-down learnings covering the race, the fix, and the trigger surface. The Appearance → Theme note you asked for is folded in: Colors / Plasma Style / Global Theme dropdowns all fire the same `nativeTheme.on('updated')` and all reproduce the duplicate icon.
- **`CLAUDE.md`** learnings index entry

## Addressing the review blocker from #491

- **Dynamic extraction of `t` and `e`** — both literal minified locals called out as a blocker are now extracted alongside `tray_var` / `menu_func` / `electron_var`: `path_var` from the `new Tray(nativeImage.createFromPath(X))` call, `enabled_var` via the same pattern as `patch_menu_bar_default`.
- **Idempotency guard re-keyed** — now matches the distinctive `${tray_var}.setImage(${electron_var}.nativeImage.createFromPath(${path_var}))` sequence using post-rename names, instead of the off-cursor `(t))` marker that would've drifted silently.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -12,6 +12,7 @@ The [`docs/learnings/`](docs/learnings/) directory contains hard-won technical k
 - [`cowork-vm-daemon.md`](docs/learnings/cowork-vm-daemon.md) — Cowork VM daemon lifecycle, respawn logic, crash diagnosis
 - [`plugin-install.md`](docs/learnings/plugin-install.md) — Anthropic & Partners plugin install flow, gate logic, backend endpoints, and DevTools recipes
 - [`apt-worker-architecture.md`](docs/learnings/apt-worker-architecture.md) — APT/DNF binary distribution via Cloudflare Worker + GitHub Releases, redirect chain, credential ownership, heartbeat runbook
+- [`tray-rebuild-race.md`](docs/learnings/tray-rebuild-race.md) — why destroy + recreate on `nativeTheme` updates briefly duplicates the tray icon on KDE Plasma, and the in-place `setImage` + `setContextMenu` fast-path that avoids the SNI re-registration race
 
 ## Code Style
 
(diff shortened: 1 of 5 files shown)


aaddrick (approved):
Hey @IliyaBrook! Thanks for the follow-up on this. The dynamic extraction across all five identifiers is exactly what #491 needed, and the idempotency guards on the post-rename names rather than the pre-rename ones is the right call. I also like that the slow path stays intact for initial creation and the `menuBarEnabled === false` case, and the fast-path's `e!==false` lines up with what `patch_menu_bar_default` rewrites `!!e` into, so the two patches stay semantically aligned. The learnings doc is solid too — the race description and the Colors / Plasma Style / Global Theme trigger surfaces are going to save someone a lot of time later.

I pushed a small fixup on top (`fef69f0`) via `maintainerCanModify` so you don't see a stranger's commit on your branch and wonder. Three things: dropped a redundant `electron_var_re_local` in favor of `$electron_var_re` from `_common.sh` (same value), swapped the silent `head -1` on `enabled_var` extraction for an explicit count-and-bail so a future second `const X=fn("menuBarEnabled")` declaration trips the build instead of binding silently to whichever the minifier emitted first, and a cosmetic double-blank-line fix between two functions.

Verified end-to-end on the live 1.3883.0 build asar. The current symbols are `Nh` / `wAt` / `t` / `e` for tray_var / menu_func / path_var / enabled_var, while the build-reference (older asar) carried `fh` / `CZe` / `t` / `e`. Symbol drift across two upstream releases makes the dynamic extraction load-bearing, and your patch handles both cleanly.

Built the AppImage on this branch and ran it on Nobara 43 / KDE Plasma 6 / Wayland. Pre-fix repro confirmed on all three Appearance triggers (Colors, Plasma Style, Global Theme). Post-fix: single icon, updates in place under all three. No regressions I could find. Approving.

---

PR #395, opened 2026-04-13
feat: add BATS unit tests for launcher-common.sh

sabiut:
## Summary

`scripts/launcher-common.sh` is 798 lines handling critical startup logic — display detection, 
Electron arg building, stale lock cleanup, cowork daemon cleanup, and the full `--doctor` 
diagnostic system — but had zero test coverage. A regression in any of these functions could 
silently break app launches across display servers and package formats.

This PR adds 48 BATS tests covering:

- **Logging** (4 tests): `setup_logging`, `log_message` — XDG_CACHE_HOME resolution and fallback
- **Display detection** (10 tests): `detect_display_backend`, `check_display` — X11, Wayland, 
  XWayland default, `CLAUDE_USE_WAYLAND=1` override, Niri auto-detection (via `NIRI_SOCKET` 
  and `XDG_CURRENT_DESKTOP`, including colon-separated and case-insensitive matching)
- **Electron args** (7 tests): `build_electron_args` — all display × package-type combinations 
  (deb/appimage/nix × X11/Wayland/XWayland), verifying `--no-sandbox`, `--ozone-platform`, 
  `--enable-wayland-ime`, `--wayland-text-input-version=3`
- **Environment** (2 tests): `setup_electron_env` — `ELECTRON_FORCE_IS_PACKAGED`, 
  `ELECTRON_USE_SYSTEM_TITLE_BAR`
- **Cleanup** (6 tests): `cleanup_stale_lock` (dead PID removal, live PID preservation, 
  non-numeric PID, regular file), `cleanup_stale_cowork_socket` (stale unix socket removal)
- **Doctor helpers** (14 tests): `_doctor_colors`, `_pass/_fail/_warn/_info` output format, 
  `_cowork_distro_id`, `_cowork_pkg_hint` (debian/ubuntu/fedora/arch/unknown + qemu special 
  case), `_electron_version` (present and missing version file)

### Test isolation

All tests run fully sandboxed — `HOME`, `XDG_CACHE_HOME`, `XDG_CONFIG_HOME`, and 
`XDG_RUNTIME_DIR` are redirected to a temp directory, and host display variables 
(`DISPLAY`, `WAYLAND_DISPLAY`, `NIRI_SOCKET`, etc.) are cleared in `setup()` to prevent 
state leakage between tests and from the host environment.

## Test plan

- [x] All 48 tests pass locally via `npx bats tests/launcher-common.bats`
- [ ] CI passes on PR branch
[...]

PR #520, opened 2026-04-25
ci: run BATS test suite on push and PR

sabiut:
## Summary

The `/tests/` directory contains **186 BATS tests** across launcher and cowork suites, but **no CI workflow has ever invoked `bats`** — the entire test suite is currently inert. A regression in `launcher-common.sh` or `cowork-vm-service.js` would not fail any check today. This PR adds a standalone `tests.yml` workflow that actually runs the suite on push and PR.

## The problem

`/tests/` has 6 BATS test files containing 186 tests total:

| File | Owner (CODEOWNERS) |
|------|--------------------|
| `tests/launcher-common.bats` (48 tests) | @sabiut |
| `tests/launcher-xrdp-detection.bats` (8 tests) | @sabiut |
| `tests/cowork-backend-detection.bats` | @RayCharlizard |
| `tests/cowork-bwrap-config.bats` | @RayCharlizard |
| `tests/cowork-path-translation.bats` | @RayCharlizard |
| `tests/cowork-resources-staging.bats` | @RayCharlizard |

Audit on `main` (`6d281c9`):

```
$ grep -rn bats .github/workflows/
(no output)
```

Confirmed: `test-artifacts.yml`, `test-flags.yml`, `ci.yml`, and `shellcheck.yml` all run, but none invokes `bats`. PR #395's recently-merged launcher-common BATS suite (48 tests) landed under that same gap and has never actually executed in CI.

## The fix

Adds **`.github/workflows/tests.yml`** — a standalone workflow that:

- Runs on every PR to `main` (unfiltered, so required-check semantics stay predictable)
- Runs on every push to `main` (path-filtered to `tests/**`, `scripts/**`, and the workflow itself)
- Installs `bats` and `nodejs` via apt — cowork tests shell out to `node` to load `scripts/cowork-vm-service.js`
- Executes `bats --print-output-on-failure tests/*.bats`
- Has a concurrency group to cancel superseded runs
- Declares explicit `permissions: contents: read`

Also updates **`.github/CODEOWNERS`** to claim the new workflow under `@sabiut`, matching the existing pattern for `test-artifacts.yml` and `test-flags.yml`. `/tests/cowork-*.bats` review ownership stays with @RayCharlizard.

## Bug surfaced by wiring CI

The first CI run on this PR caught two pre-existing test failures in `tests/cowork-bwrap-config.bats` (tests 46 and 47 — `doctor: reports custom bwrap mounts` and `doctor: warns about disabled critical mount /usr`).

Root cause: `_doctor_check_bwrap_mounts` in `scripts/doctor.sh` resolves the config dir via `${XDG_CONFIG_HOME:-$HOME/.config}/Claude`. The test `setup()` only sandboxed `HOME` via `TEST_TMP`. GitHub Actions runners export `XDG_CONFIG_HOME` ambient, which overrode the per-test `HOME` and made the function read the runner's real config dir. Assertions silently failed against empty output.

The bug existed on `main` before this PR, but was hidden because the BATS suite never ran in CI — exactly the gap this PR exists to close. Fixed in the second commit (`fix(tests): unset XDG_CONFIG_HOME in cowork-bwrap-config setup`) by adding `unset XDG_CONFIG_HOME` to the test's `setup()`. Verified locally that all 186 tests pass with `XDG_CONFIG_HOME` set ambient (reproduces CI env).

## Why standalone instead of extending `test-artifacts.yml`?

- BATS unit tests should run in seconds on every push, not wait ~10 minutes for a full artifact build.
- Cleaner failure attribution — a red `BATS Tests` check means "your code broke a test", not "the build broke and we never reached tests".
- Smaller blast radius for a first-wiring PR — no `ci.yml` changes to coordinate, easier to roll back.
- Trivial to promote into `ci.yml` as a build gate later once it's proven stable.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -73,6 +73,7 @@
 /scripts/doctor.sh                      @sabiut
 /.github/workflows/test-artifacts.yml   @sabiut
 /.github/workflows/test-flags.yml       @sabiut
+/.github/workflows/tests.yml            @sabiut
 
 # Shared review — either owner can approve.
 # TROUBLESHOOTING is mostly the --doctor user-facing guide; lint
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,45 @@
+name: BATS Tests
+run-name: |
+  BATS: ${{
+    github.event_name == 'pull_request' && format('PR #{0} by @{1} - {2}', github.event.pull_request.number, github.actor, github.event.pull_request.title) ||
+    github.event_name == 'push' && github.event.head_commit && format('Push by @{0} - {1}', github.actor, github.event.head_commit.message) ||
+    format('{0} triggered by @{1}', github.event_name, github.actor)
+  }}
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "tests/**"
+      - "scripts/**"
+      - ".github/workflows/tests.yml"
+  pull_request:
+    branches: [main]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: bats-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  bats:
+    name: BATS unit tests
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Install BATS and Node.js
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y bats nodejs
+
+      - name: Run BATS test suite
+        # Cowork tests load scripts/cowork-vm-service.js via `node` —
+        # the `nodejs` install above is what they need.
+        run: bats --print-output-on-failure tests/*.bats
--- a/tests/cowork-bwrap-config.bats
+++ b/tests/cowork-bwrap-config.bats
@@ -46,6 +46,12 @@ function assertDeepEqual(actual, expected, msg) {
 setup() {
 	TEST_TMP=$(mktemp -d)
 	export TEST_TMP
+
+	# The doctor checks resolve config via ${XDG_CONFIG_HOME:-$HOME/.config}.
+	# Sandboxing HOME alone is insufficient because GitHub Actions runners
+	# (and many user environments) export XDG_CONFIG_HOME ambient, which
+	# overrides the per-test HOME and makes the function read the wrong dir.
+	unset XDG_CONFIG_HOME
 }
 
 teardown() {


aaddrick (commented):
@sabiut — 186 inert tests only show up when someone actually runs them. Nice catch wiring this in. `_doctor_check_bwrap_mounts` (`scripts/doctor.sh:140`) reads `${XDG_CONFIG_HOME:-$HOME/.config}/Claude`, and GH runners export `XDG_CONFIG_HOME` ambient. That wins the `:-` expansion before the inline `HOME=` at the call site gets a chance. Your `unset` in setup() forces the fallback path so the HOME override takes effect.

I checked out the branch and ran the suite with deliberately polluted env: empty `XDG_CONFIG_HOME`, populated, and pointed at a config crafted to match the test's assertion patterns (the false-positive scenario). 186/186 in every case. Reverting the `unset` reproduces the failure deterministically. Fix is correct.

Workflow shape is what I'd have asked for. No reason to gate seconds-fast unit tests behind a 10-minute artifact build. Push path-filtered with PR unfiltered lines up with required-check semantics under GH branch protection, and the concurrency group plus `permissions: contents: read` are both good to see.

@RayCharlizard — `tests/cowork-bwrap-config.bats` is on your surface per CODEOWNERS. Soft tag for awareness, not blocking. Fix is verified, so ship-it whenever you've had a look. One optional observation if consistency matters: `launcher-common.bats:21-30` does the full XDG sandbox in setup() itself rather than per-call inline. Converging cowork-bwrap-config on that pattern would be cleaner, but not for this PR.

---

RayCharlizard (approved):
Hey @sabiut — confirming the fix from the cowork side. Good catch.

Walking through the path-resolution claim: `_doctor_check_bwrap_mounts` at `scripts/doctor.sh:140` resolves the config dir as `${XDG_CONFIG_HOME:-$HOME/.config}/Claude`. The `:-` only falls back when XDG_CONFIG_HOME is unset or empty. The two failing tests use `HOME="${TEST_TMP}" run _doctor_check_bwrap_mounts` — that scoped HOME override is meaningless if XDG_CONFIG_HOME is already exported in the runner env, which it is on `ubuntu-latest`. Your `unset XDG_CONFIG_HOME` in setup() forces the fallback path so the inline HOME actually wins. Targeted and correct.

One thing worth noting that strengthens the case: the third doctor test in this file (`doctor: no output when no custom mounts configured` at line 657) was probably also reading from the runner's real config dir before this fix. It passes either way because it asserts empty output and the runner has no `claude_desktop_linux_config.json` to find — green for the wrong reason. Your unset in shared setup() fixes that too.

I checked the other cowork suites for the same ambient-leak pattern. `cowork-backend-detection.bats` and `cowork-path-translation.bats` only invoke `node` against pure JS functions — no `XDG_*` resolution, no shell. `cowork-resources-staging.bats` works on a `TEST_TMP` staging layout via `claude_extract_dir`/`electron_resources_dest`, no XDG paths. The fix is complete, not partial.

Workflow shape is fine. `bats --print-output-on-failure tests/*.bats` picks up all four cowork-*.bats and both launcher-*.bats files. Concurrency group, `permissions: contents: read`, path filters scoped to `tests/**` and `scripts/**`, PR trigger unfiltered for predictable required-check semantics — all the pieces I'd expect. CODEOWNERS update is clean: only adds the new `tests.yml` line under @sabiut, leaves cowork-*.bats ownership where it sits.

Echoing aaddrick's optional observation — `launcher-common.bats:21-30` does the full XDG sandbox in setup() (exports HOME and all XDG_* into TEST_TMP, then `mkdir -p`). Converging cowork-bwrap-config on that pattern would be more defensive than a single targeted `unset`, since any future test in the file that touches a different XDG var inherits the same trap. Not a blocker, and expanding the sandbox here would belong in its own change.

Confirmed CI is green at HEAD (`BATS unit tests` passes in 35s alongside everything else). Approving from the cowork side — ship it whenever.

---

Issue #526, opened 2026-04-25
[bug]: Claude Desktop spawns every stdio MCP server twice (chat + Code/Agent panel)


communitytranslations:
### Version (`claude-desktop --doctor` output)

```shell
$ claude-desktop --doctor  
Claude Desktop Diagnostics
================================

[WARN] claude-desktop not found via dpkg (AppImage?)
[PASS] Display server: X11 (DISPLAY=:0)
       Menu bar mode: auto (default, Alt toggles visibility)
[PASS] Electron: v41.3.0 (/tmp/.mount_claudeSdUUDR/usr/lib/node_modules/electron/dist/electron)
[FAIL] Chrome sandbox: perms=755, owner=root
       Fix: sudo chown root:root /tmp/.mount_claudeSdUUDR/usr/lib/node_modules/electron/dist/chrome-sandbox
            sudo chmod 4755 /tmp/.mount_claudeSdUUDR/usr/lib/node_modules/electron/dist/chrome-sandbox
[PASS] SingletonLock: no lock file (OK)
[PASS] MCP config: valid JSON (/home/USER/.config/Claude/claude_desktop_config.json)
       MCP servers configured: 5
[PASS] Node.js: v22.20.0
       Path: /home/USER/.nvm/versions/node/v22.20.0/bin/node
[WARN] Desktop entry not found (expected for AppImage installs)
[PASS] Disk space: 57325MB free

Cowork Mode
----------------
[PASS] bubblewrap: found
[PASS] bubblewrap: sandbox probe succeeded
[PASS] KVM: accessible
[PASS] vsock: module loaded
[PASS] QEMU: found
[PASS] socat: found
       virtiofsd: not found
       VM image: not downloaded yet
       Cowork isolation: bubblewrap (namespace sandbox)
[WARN] Cowork daemon: orphaned (PIDs: 451401)
       Fix: Restart Claude Desktop (daemon will be cleaned up automatically)
[WARN] Log file: 12169KB (consider clearing: rm '/home/USER/.cache/claude-desktop-debian/launcher.log')

1 check(s) failed.
See above for fixes.
```

### What happened

When a session has both the classic chat panel and the Code/Agent (Cowork) panel active, **every stdio MCP server declared in `claude_desktop_config.json` is spawned twice** by the Electron main process. The two processes run in parallel, each connected to a different panel, and corrupt any state the MCP keeps outside its own process (shared sockets, files on disk, external services).

[...]

PR #527, opened 2026-04-25
docs(learnings): document MCP double-spawn upstream bug (#526)

RayCharlizard:
## Summary

Adds `docs/learnings/mcp-double-spawn.md` capturing [#526](https://github.com/aaddrick/claude-desktop-debian/issues/526)'s root-cause analysis so future contributors hitting the symptom find a quick pointer instead of re-investigating from zero.

The bug: stdio MCP servers in `claude_desktop_config.json` get spawned twice when both the chat panel and the Code/Agent (Cowork) panel are active. Reporter (`@communitytranslations`) traced it to two parallel session managers in upstream Electron main (`LocalSessions` and `LocalAgentModeSessions`) each holding an independent Claude Agent SDK `query` whose stdio transport (`Du.spawn`) bypasses the global MCP registry (`hZ` map / `oUt` accessor) that *would* dedupe them.

I independently verified our packaging is not implicated:
- All 7 patches in `scripts/patches/*.sh` — zero references to MCP, mcpServer, LocalSessions, LocalAgentModeSessions, transportToClient, MessageChannelMain, n2t, hZ, oUt
- `scripts/launcher-common.sh` and `scripts/packaging/*.sh` — no MCP or config-load logic
- `scripts/doctor.sh:420` only reads the config to JSON-lint it; not in the runtime spawn path

The entry documents:
- Symptoms with example `ps -ef` output
- Upstream root cause with the relevant minified symbols (flagged as drift-prone)
- Status: **upstream Anthropic bug, not patchable here** — explicitly waves off future contributors from regex-patching minified symbols
- Workaround for affected MCP authors (lockfile + idempotent writes, per reporter's `cb7bfbb`)
- Routing guidance: primary venue is in-app feedback or `support@anthropic.com`; `anthropics/claude-code` defensible as secondary since the SDK transport is shared surface area; the embedded CLI is **not** the cause

Also adds the entry to the `docs/learnings/` index in `CLAUDE.md`.

## Test plan

- [x] `docs/learnings/mcp-double-spawn.md` renders cleanly
- [x] `CLAUDE.md` learnings index includes the new entry in the right position
- [x] No code paths touched; no patch surface modified

---
70% AI / 30% Human
Claude: surveyed patches to verify packaging exoneration, drafted the learnings entry, wrote the comment plan
Human: routed the issue, caught a missed contrarian pass and triggered the policy update, approved scope and tone

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -13,7 +13,7 @@ The [`docs/learnings/`](docs/learnings/) directory contains hard-won technical k
 - [`plugin-install.md`](docs/learnings/plugin-install.md) — Anthropic & Partners plugin install flow, gate logic, backend endpoints, and DevTools recipes
 - [`apt-worker-architecture.md`](docs/learnings/apt-worker-architecture.md) — APT/DNF binary distribution via Cloudflare Worker + GitHub Releases, redirect chain, credential ownership, heartbeat runbook
 - [`tray-rebuild-race.md`](docs/learnings/tray-rebuild-race.md) — why destroy + recreate on `nativeTheme` updates briefly duplicates the tray icon on KDE Plasma, and the in-place `setImage` + `setContextMenu` fast-path that avoids the SNI re-registration race
-- [`mcp-double-spawn.md`](docs/learnings/mcp-double-spawn.md) — Stdio MCPs spawn 2× when both chat and Code/Agent panels are active; upstream Anthropic bug, MCP-author workaround documented
+- [`mcp-double-spawn.md`](docs/learnings/mcp-double-spawn.md) — Stdio MCPs spawn 2× when chat and Code/Agent panels are both active, root cause in upstream session managers, MCP-author workaround
 
 ## Code Style
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -13,6 +13,7 @@ The [`docs/learnings/`](docs/learnings/) directory contains hard-won technical k
 - [`plugin-install.md`](docs/learnings/plugin-install.md) — Anthropic & Partners plugin install flow, gate logic, backend endpoints, and DevTools recipes
 - [`apt-worker-architecture.md`](docs/learnings/apt-worker-architecture.md) — APT/DNF binary distribution via Cloudflare Worker + GitHub Releases, redirect chain, credential ownership, heartbeat runbook
 - [`tray-rebuild-race.md`](docs/learnings/tray-rebuild-race.md) — why destroy + recreate on `nativeTheme` updates briefly duplicates the tray icon on KDE Plasma, and the in-place `setImage` + `setContextMenu` fast-path that avoids the SNI re-registration race
+- [`mcp-double-spawn.md`](docs/learnings/mcp-double-spawn.md) — Stdio MCPs spawn 2× when both chat and Code/Agent panels are active; upstream Anthropic bug, MCP-author workaround documented
 
 ## Code Style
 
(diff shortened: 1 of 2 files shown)


aaddrick (changes requested):
Verified the exoneration claims hold. Re-grepped case-sensitive across `scripts/patches/*.sh`, `scripts/launcher-common.sh`, `scripts/packaging/*.sh`, and extended to `nix/`, `worker/`, `scripts/setup/`, `build.sh`, `frame-fix-wrapper.js`, `cowork-vm-service.js`, and `staging/`. Zero relevant hits. `scripts/doctor.sh:420` is JSON-lint only, and CI is green. Analysis is solid.

One real bug on line 46: the IPC namespace table has `claude.web_$_LocalAgentModeSessions_*` but should be `claude.web_$_LocalAgentModeSessions_$_*`. The `_$_` separator is structural — `build-reference/app-extracted/.vite/build/index.js:64565` has the actual channel names (`claude.web_$_LocalAgentModeSessions_$_start`, `_$_sendMessage`). The `LocalSessions` row above it is correct, so the separator probably got dropped during column alignment. This matters because it's a literal grep target for the next debugger.

The Routing Upstream Reports section lists `anthropics/claude-code` as defensible secondary because the SDK transport is shared surface area, but the SDK lives in `anthropics/claude-agent-sdk-typescript` (separate public repo, issues enabled). Pointing SDK-transport reports at `claude-code` while the doc itself says the CLI isn't on the spawn path is the exact contradiction your "don't frame as CLI double-spawning" warning was trying to prevent — CLI maintainers can close as off-topic and bounce the reporter. Either swap in `claude-agent-sdk-typescript` or drop the secondary venue.

Nit while you're on line 46: the original issue body had a 4-column table with a `Logs prefix` column (`[CCD]` / `[LAM]`). Those prefixes are what users grep `~/.config/Claude/logs/` for to confirm they're hitting this bug specifically — worth adding back. Optional and probably skip on its own, but if you re-roll: the workaround section punts to `cb7bfbb` for the lockfile reference, and a one-liner about atomic reclaim (rename-over, not unlink-then-open) would head off the naive port. @RayCharlizard once the separator and routing get fixed I'm good to merge.

---

RayCharlizard:
Thanks for the catch on the separator — that one would have wasted real time for the next person grepping. Pushed [`5fe1d69`](https://github.com/aaddrick/claude-desktop-debian/pull/527/commits/5fe1d69) with all four:

- **Separator fix on line 46:** table now reads `claude.web_\$_LocalAgentModeSessions_\$_*`. Verified my edit lines up with the channel names you cited at `index.js:64565` (`_\$_start`, `_\$_sendMessage`).
- **Logs prefix column restored:** `[CCD]` / `[LAM]` back in a 4-column table, plus a one-liner pointing readers at `~/.config/Claude/logs/` so they know what to grep when confirming the bug.
- **Routing swapped:** secondary venue is now [`anthropics/claude-agent-sdk-typescript`](https://github.com/anthropics/claude-agent-sdk-typescript) (verified public + issues enabled). You're right that pointing SDK-transport reports at `claude-code` while the doc itself says the CLI isn't on the spawn path was the exact rake the warning paragraph was trying to head off.
- **Atomic reclaim added:** workaround section now calls out `rename()` over `unlink()`-then-re-open near the `cb7bfbb` reference. Cheap to add, heads off the naive port.

Ready when you are.

---

aaddrick (approved):
Re-reviewed `5fe1d69` against the previous round. All four items landed: the separator lines up with the IPC channel names at `index.js:64565+`, the routing target swap holds, the logs-prefix column came back with a paragraph of context explaining what to grep for, and the atomic-reclaim note is correctly placed alongside the lockfile workaround.

While answering "could we patch it?" I went back into `index.js` and ended up surfacing findings that sharpen the doc's framing — specifically around the actual root cause (two independent registries, not a race) and the routing destination for the upstream report. None of it blocks this PR; it's a follow-up amendment, not another round of changes here. Filing it as a separate issue so we can fold it into a small follow-up commit.

Approving and merging. Thanks @RayCharlizard.

---

PR #531, opened 2026-04-27
feat(bwrap): support {src, dst} mount form in coworkBwrapMounts

### LINKED ISSUE #530: Feature: support {src, dst} mount form in coworkBwrapMounts to allow distinct host/sandbox paths (e.g. persistent /tmp)
## Summary

`coworkBwrapMounts` (added in #339) lets users add bind mounts to the bwrap
sandbox but only as a *string* — which `mergeBwrapArgs` translates to
`--bind <p> <p>` (host path mounted at the same path inside the sandbox).

This means the host source and the sandbox destination are always identical,
which prevents a legitimate use case: replacing a default sandbox mount with a
persistent host directory mounted at the same in-sandbox path.

## Motivating use case: persistent `/tmp` across Bash tool calls

Each Bash tool invocation in local-agent / Cowork mode spawns a fresh `bwrap`
process with `--tmpfs /tmp` (`cowork-vm-service.js:1277`) and
`--die-with-parent` (`cowork-vm-service.js:1365`). The tmpfs is destroyed when
the command exits, so anything written to `/tmp` is lost before the next call.

This breaks workflows that legitimately stage state in `/tmp` between commands:
- multi-step downloads (`curl -o /tmp/...` then process in a later call)
- `pip install --user` writing wheels to `/tmp/...` and reusing them
- tools that hardcode `/tmp` paths (`get-pip.py`, some installers)

Today there is no workaround:
- `additionalBinds: [\"/tmp\"]` would expose the *host's* `/tmp` to the sandbox
  (regression of the isolation the bwrap rewrite in #309 set up), and even then
  it requires `disabledDefaultBinds: [\"/tmp\"]` to remove the default tmpfs.
- There is no way to say \"mount `~/.cache/claude-tmp` from the host *as*
  `/tmp` inside the sandbox\" — which is exactly what the user wants.

## Proposal

Extend `additionalROBinds` and `additionalBinds` to also accept entries of the
form `{ \"src\": \"...\", \"dst\": \"...\" }` in addition to the existing string form.

### Example: persistent `/tmp` mapped to a host cache dir

```json
{
  \"preferences\": {
    \"coworkBwrapMounts\": {
      \"additionalBinds\": [
        { \"src\": \"/home/user/.cache/claude-tmp\", \"dst\": \"/tmp\" }
      ],
      \"disabledDefaultBinds\": [\"/tmp\"]
    }
  }
}
```

Resulting bwrap args (excerpt):
```
... --proc /proc --tmpfs /run --bind /home/user/.cache/claude-tmp /tmp ...
```

Files written to `/tmp` inside the sandbox now survive across Bash calls,
without exposing the host `/tmp`.

### Validation rules

- `src`: same checks as the current string form
  - absolute, not in `FORBIDDEN_MOUNT_PATHS` (`/`, `/proc`, `/dev`, `/sys`)
  - if RW (`additionalBinds`): must be under `\$HOME`
- `dst`: only the absolute + non-forbidden checks
  - the `\$HOME` constraint is intentionally skipped because the *whole point*
    of `{src, dst}` is to map onto sandbox paths outside `\$HOME` (`/tmp`,
    `/sandbox/...`, etc.)
- Malformed objects (missing `src`/`dst`, non-string values) are silently
  filtered out with a `BwrapConfig: rejected ...` warning, matching the
  existing string-validation behavior

### Backwards compatibility

100%. The string form (`\"/path\"` → `--bind /path /path`) is preserved exactly
as-is. All 36 existing tests in `tests/cowork-bwrap-config.bats` pass without
modification.

## Implementation status

I have a local patch that:
- Adds object-form support in `loadBwrapMountsConfig.filterMounts`
  (`scripts/cowork-vm-service.js`)
- Updates `mergeBwrapArgs` to emit `--bind src dst` per type
  (`scripts/cowork-vm-service.js`)
- Updates `_doctor_check_bwrap_mounts` to render the object form as
  `src -> dst` in `--doctor` output (`scripts/doctor.sh`, both Python and
  Node parser branches)
- Adds 13 new BATS tests covering accept/reject paths, mixed string+object
  configs, the persistent-`/tmp` recipe end-to-end, and the doctor
  rendering — all passing (58/58 total)

Tested end-to-end on a locally rebuilt `.deb`: the resulting bwrap command
line contains the expected `--bind ~/.cache/claude-tmp /tmp` and the
default `--tmpfs /tmp` is correctly stripped via `disabledDefaultBinds`.

Happy to open a PR if the design is acceptable.

## References

- #339 — original `coworkBwrapMounts` feature
- #309 — bwrap mount strategy rewrite (where `--tmpfs /tmp` was introduced)
- `scripts/cowork-vm-service.js` — `loadBwrapMountsConfig` / `mergeBwrapArgs`
  (lines 739–855)

---

cbonnissent:
Closes #530

## Summary

- Extends `coworkBwrapMounts` (introduced in #339) so `additionalROBinds`
  and `additionalBinds` accept entries of the form `{ src, dst }` in
  addition to the existing string form
- Unlocks the persistent-`/tmp` use case: each Bash tool call spawns a
  fresh `bwrap` with `--tmpfs /tmp` + `--die-with-parent`, so anything
  in `/tmp` is wiped between calls; the old string-only API
  (`--bind p p`) had no way to map a host dir under `$HOME` onto `/tmp`
  inside the sandbox without exposing the host `/tmp` itself

## Configuration example

```json
{
  "preferences": {
    "coworkBwrapMounts": {
      "additionalBinds": [
        { "src": "/home/user/.cache/claude-tmp", "dst": "/tmp" }
      ],
      "disabledDefaultBinds": ["/tmp"]
    }
  }
}
```

Resulting bwrap args (excerpt):
```
... --proc /proc --tmpfs /run --bind /home/user/.cache/claude-tmp /tmp ...
```

## Validation rules

- `src`: absolute, not in `FORBIDDEN_MOUNT_PATHS`, must be under `$HOME`
  when RW (same checks as the string form)
- `dst`: absolute and non-forbidden only — the `$HOME` constraint is
  intentionally skipped because the whole point of the form is to map
  outside `$HOME` (e.g. `/tmp`)
- Malformed objects (missing `src`/`dst`, non-string values) are filtered
  out with a `BwrapConfig: rejected ...` warning, matching the existing
  string-validation behavior

## Backwards compatibility

100%. The string form (`"/path"` → `--bind /path /path`) is preserved
exactly as-is. All 36 pre-existing tests in
`tests/cowork-bwrap-config.bats` pass without modification.

## Doctor (`--doctor`)

Renders the object form as `src -> dst` in both the Python and Node
parser branches of `_doctor_check_bwrap_mounts`.

## Files

- `scripts/cowork-vm-service.js` — `loadBwrapMountsConfig.filterMounts`
  accepts both forms; `mergeBwrapArgs` emits `--bind src dst` per type
- `scripts/doctor.sh` — updated `_doctor_check_bwrap_mounts` rendering
- `tests/cowork-bwrap-config.bats` — 13 new BATS tests covering
  accept/reject paths, mixed string+object configs, the persistent-`/tmp`
  recipe end-to-end, and the doctor rendering (58/58 passing)

## Test plan

- [x] `bats tests/cowork-bwrap-config.bats` passes (58/58)
- [x] `./build.sh --build deb --clean no` succeeds with all 12 cowork
      patches applied
- [x] Manual: extracted .deb shows the patched `filterMounts` function in
      `app.asar.unpacked/cowork-vm-service.js`
- [x] Manual: feeding the example config through
      `loadBwrapMountsConfig` + `mergeBwrapArgs` produces
      `--bind ~/.cache/claude-tmp /tmp` with the default `--tmpfs /tmp`
      correctly stripped
- [ ] End-to-end runtime validation (requires Claude Desktop restart, not
      yet performed by the author since the PR session itself runs in the
      sandbox being modified)

---
30% AI / 70% Human
Claude: drafted the patch (filterMounts polymorphism, doctor rendering, BATS tests), wrote the commit and PR descriptions
Human: identified the root cause, chose the polymorphic-format design over alternatives, validated the approach, drove the iteration loop and end-to-end test plan

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -69,15 +69,47 @@ the sandbox mount points via `~/.config/Claude/claude_desktop_linux_config.json`
 
 | Key | Type | Description |
 |-----|------|-------------|
-| `additionalROBinds` | `string[]` | Extra paths mounted read-only inside the sandbox. Accepts any absolute path except `/`, `/proc`, `/dev`, `/sys`. |
-| `additionalBinds` | `string[]` | Extra paths mounted read-write inside the sandbox. **Restricted to paths under `$HOME`** for security. |
+| `additionalROBinds` | `(string \| {src, dst})[]` | Extra paths mounted read-only inside the sandbox. Accepts any absolute path except `/`, `/proc`, `/dev`, `/sys`. |
+| `additionalBinds` | `(string \| {src, dst})[]` | Extra paths mounted read-write inside the sandbox. **`src` is restricted to paths under `$HOME`** for security; `dst` is unconstrained. |
 | `disabledDefaultBinds` | `string[]` | Default mounts to skip. Cannot disable critical mounts (`/`, `/dev`, `/proc`). Use with caution: disabling `/usr` or `/etc` may break tools inside the sandbox. |
+
+### Distinct host/sandbox paths (`{src, dst}` form)
+
+By default a string entry like `"/opt/tools"` mounts the host path at the
+*same* path inside the sandbox. To map a host directory to a different path
+inside the sandbox, use the object form `{ "src": "...", "dst": "..." }`.
+
+The most common use case is making `/tmp` persistent across Bash tool calls.
+Each Bash invocation spawns a fresh `bwrap` with `--tmpfs /tmp` and
+`--die-with-parent`, so the default `/tmp` is wiped between calls. Mapping a
+host cache directory onto `/tmp` keeps state across calls without exposing the
+host's real `/tmp`:
+
+```json
+{
+  "preferences": {
+    "coworkBwrapMounts": {
+      "additionalBinds": [
+        { "src": "/home/user/.cache/claude-tmp", "dst": "/tmp" }
+      ],
+      "disabledDefaultBinds": ["/tmp"]
+    }
+  }
+}
+```
+
+`disabledDefaultBinds: ["/tmp"]` is required to remove the default
+`--tmpfs /tmp` so the bind takes effect.
+
+The string and object forms can be mixed freely in the same array.
 
 ### Security notes
 
 - Paths `/`, `/proc`, `/dev`, `/sys` (and their subpaths) are always rejected
-- Read-write mounts (`additionalBinds`) are restricted to paths under your home
-  directory
+  for both `src` and `dst`
+- For read-write mounts (`additionalBinds`), `src` must be under your home
+  directory. `dst` has no `$HOME` constraint — that is the entire purpose of
+  the object form (e.g. mapping onto `/tmp`)
 - The core sandbox structure (`--tmpfs /`, `--unshare-pid`, `--die-with-parent`,
   `--new-session`) cannot be modified
 - Mount order is enforced: user mounts cannot override security-critical
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -103,6 +103,10 @@ host's real `/tmp`:
 
 The string and object forms can be mixed freely in the same array.
 
+> **Caution:** Mapping `dst` onto a default RO mount (`/usr`, `/etc`, `/bin`,
+> `/sbin`, `/lib`, `/lib64`) silently replaces it inside the sandbox; you
+> almost never want this, and `--doctor` will warn if you do.
+
 ### Security notes
 
 - Paths `/`, `/proc`, `/dev`, `/sys` (and their subpaths) are always rejected
--- a/scripts/cowork-vm-service.js
+++ b/scripts/cowork-vm-service.js
@@ -763,21 +763,41 @@ function loadBwrapMountsConfig(configPath, logFn) {
         return empty;
     }
 
-    function filterPaths(arr, readWrite) {
+    function filterMounts(arr, readWrite) {
         if (!Array.isArray(arr)) return [];
         return arr.filter(p => {
-            if (typeof p !== 'string') return false;
-            const result = validateMountPath(p, { readWrite });
-            if (!result.valid) {
-                warn(`BwrapConfig: rejected path "${p}": ${result.reason}`);
+            // String form: "/path" → bind /path /path
+            if (typeof p === 'string') {
+                const r = validateMountPath(p, { readWrite });
+                if (!r.valid) {
+                    warn(`BwrapConfig: rejected path "${p}": ${r.reason}`);
+                }
+                return r.valid;
             }
-            return result.valid;
+            // Object form: { src, dst } → bind src dst (different paths)
+            if (p && typeof p === 'object'
+                && typeof p.src === 'string' && typeof p.dst === 'string') {
+                const srcRes = validateMountPath(p.src, { readWrite });
+                if (!srcRes.valid) {
+                    warn(`BwrapConfig: rejected mount src "${p.src}": ${srcRes.reason}`);
+                    return false;
+                }
+                // dst is the in-sandbox path — skip the $HOME constraint
+                // (the whole point of {src,dst} is to map outside it).
+                const dstRes = validateMountPath(p.dst, { readWrite: false });
+                if (!dstRes.valid) {
+                    warn(`BwrapConfig: rejected mount dst "${p.dst}": ${dstRes.reason}`);
+                    return false;
+                }
+                return true;
+            }
+            return false;
         });
     }
 
     return {
-        additionalROBinds: filterPaths(mounts.additionalROBinds, false),
-        additionalBinds: filterPaths(mounts.additionalBinds, true),
+        additionalROBinds: filterMounts(mounts.additionalROBinds, false),
+        additionalBinds: filterMounts(mounts.additionalBinds, true),
         disabledDefaultBinds: Array.isArray(mounts.disabledDefaultBinds)
             ? mounts.disabledDefaultBinds
                 .filter(p => {
@@ -844,11 +864,19 @@ function mergeBwrapArgs(defaultArgs, config) {
         }
     }
 
-    for (const p of config.additionalROBinds) {
-        result.push('--ro-bind', p, p);
+    for (const m of config.additionalROBinds) {
+        if (typeof m === 'string') {
+            result.push('--ro-bind', m, m);
+        } else {
+            result.push('--ro-bind', m.src, m.dst);
+        }
     }
-    for (const p of config.additionalBinds) {
-        result.push('--bind', p, p);
+    for (const m of config.additionalBinds) {
+        if (typeof m === 'string') {
+            result.push('--bind', m, m);
+        } else {
+            result.push('--bind', m.src, m.dst);
+        }
     }
 
     return result;
--- a/scripts/cowork-vm-service.js
+++ b/scripts/cowork-vm-service.js
@@ -770,7 +770,7 @@ function loadBwrapMountsConfig(configPath, logFn) {
             if (typeof p === 'string') {
                 const r = validateMountPath(p, { readWrite });
                 if (!r.valid) {
-                    warn(`BwrapConfig: rejected path "${p}": ${r.reason}`);
+                    warn(`BwrapConfig: rejected mount "${p}": ${r.reason}`);
                 }
                 return r.valid;
             }
--- a/scripts/doctor.sh
+++ b/scripts/doctor.sh
@@ -189,25 +189,53 @@ JSEOF
 	if [[ $parser == 'python3' ]]; then
 		parsed_output=$(python3 - "$mounts_json" 2>/dev/null <<'PYEOF'
 import json, sys
+def fmt(p):
+    if isinstance(p, str):
+        return p
+    if isinstance(p, dict) and isinstance(p.get('src'), str) \
+            and isinstance(p.get('dst'), str):
+        return p['src'] + ' -> ' + p['dst']
+    return None
 m = json.loads(sys.argv[1])
 for p in m.get('additionalROBinds', []):
-    print(p)
+    s = fmt(p)
+    if s is not None:
+        print(s)
 print('---')
 for p in m.get('additionalBinds', []):
-    print(p)
+    s = fmt(p)
+    if s is not None:
+        print(s)
 print('---')
 for p in m.get('disabledDefaultBinds', []):
-    print(p)
+    if isinstance(p, str):
+        print(p)
 PYEOF
 )
 	else
 		parsed_output=$(node - "$mounts_json" 2>/dev/null <<'JSEOF'
+function fmt(p) {
+    if (typeof p === 'string') return p;
+    if (p && typeof p === 'object'
+        && typeof p.src === 'string' && typeof p.dst === 'string') {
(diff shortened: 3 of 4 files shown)


RayCharlizard (commented):
Hey, this is solid work. The polymorphic shape, the test coverage, and the honest unchecked e2e box all show care. I pulled the branch, re-ran `bats tests/cowork-bwrap-config.bats` (58/58 pass), walked the validation logic, and traced the resulting bwrap arg order. A few notes below — none of them block the PR direction; one is a real footgun worth addressing.

**Issue 1: `dst` can silently shadow default RO mounts (`/usr`, `/etc`, `/bin`, `/lib*`, `/sbin`).**

This is the only finding I'd really push on. The forbidden set is `{/, /proc, /dev, /sys}`, so a config like:

```json
{ "additionalROBinds": [{ "src": "/home/u/fake-etc", "dst": "/etc" }] }
```

passes validation. In `mergeBwrapArgs` the user's `--ro-bind /home/u/fake-etc /etc` is appended *after* the default `--ro-bind /etc /etc`, and bwrap honors the later mount, so the system view of `/etc` is silently replaced. Same goes for `/usr`, `/bin`, `/lib*`, `/sbin`.

This isn't a sandbox escape — the user is already inside the sandbox and `additionalBinds` requires `src` under `$HOME` anyway. But it's a config footgun that breaks the sandbox in a non-obvious way. The existing `disabledDefaultBinds` flow already surfaces a doctor `WARN` when someone disables `/usr` or `/etc`; the object form deserves the same surface.

Recommendation: add a doctor-side warning when an additional mount's `dst` lands on (or under) one of the default RO mount points. Something like:

```sh
# in scripts/doctor.sh, after the rw_binds rendering loop
local dst_warned=false
for line in $ro_binds $rw_binds; do
    [[ $line == *' -> '* ]] || continue
    local dst=${line##* -> }
    case $dst in
        /usr|/usr/*|/etc|/etc/*|/bin|/bin/*|/sbin|/sbin/*|/lib|/lib/*|/lib64|/lib64/*)
            _warn "Mount dst '${dst}' shadows a default sandbox mount" \
                  '(may break system tools inside the sandbox)'
            dst_warned=true
            ;;
    esac
done
```

That keeps the flexibility (`/tmp`, `/sandbox/...` still pass clean) while flagging the genuinely surprising case.

**Issue 2: docs gap on the shadowing implication.**

`docs/CONFIGURATION.md` says `dst` is unconstrained because the point is `/tmp`, but doesn't mention that `dst` can also overlap other defaults. One line under "Distinct host/sandbox paths" — something like "Mapping `dst` onto a default RO mount (`/usr`, `/etc`, etc.) replaces it inside the sandbox; you almost never want this, and `--doctor` will warn." — would close the loop.

**Issue 3: missing test for `dst` *under* a forbidden path.**

The existing object-form reject tests cover `dst: '/proc'` (exact) and forbidden src, but not `dst: '/proc/self'` style subpaths. `validateMountPath` handles it via the same code path as exact matches, so it's covered transitively, but adding one explicit case keeps the contract pinned:

```bash
@test "loadBwrapMountsConfig: rejects {src,dst} with dst under forbidden path" {
    run node -e "${NODE_PREAMBLE}
const configPath = '${TEST_TMP}/config.json';
fs.writeFileSync(configPath, JSON.stringify({
    preferences: {
        coworkBwrapMounts: {
            additionalROBinds: [{ src: '/opt/tools', dst: '/proc/self' }]
        }
    }
}));
const result = loadBwrapMountsConfig(configPath);
assertDeepEqual(result.additionalROBinds, [], 'dst under /proc rejected');
"
    [[ "$status" -eq 0 ]]
}
```

**Issue 4 (nit): warning text consistency.**

The string form warns `BwrapConfig: rejected path "/x"`, the object form warns `BwrapConfig: rejected mount src "/x"` / `... mount dst "/x"`. Mildly inconsistent — `path` vs `mount src`/`mount dst`. Not worth churning the diff for, but if you're already touching the file, harmonizing on `BwrapConfig: rejected mount ... "/x"` for everything would read a bit cleaner.

**Verified locally:**

- `bats tests/cowork-bwrap-config.bats` — 58/58 pass, matches your claim
- Reject-path coverage is genuinely thorough: malformed objects (missing src, missing dst, non-string `src: 42`, empty `{}`, `null`), forbidden src, forbidden dst, non-absolute src/dst, RW src outside `$HOME`. Not just happy paths.
- The `realpathSync` defense-in-depth on `dst` does run on the host (since dst is also passed through `validateMountPath`), so a host-side `/tmp -> /proc/foo` symlink would be caught.
- Doctor `src -> dst` rendering is clean in both Python and Node branches; the malformed-entry filtering via `fmt() returning null` correctly mirrors `filterMounts`.

**Backwards compat is real:** the string form goes through the exact same `--bind p p` path as before, and the existing 36 tests all still pass without modification. No migration risk.

The persistent-`/tmp` use case is a legitimate gap — I had not considered that the per-tool-call `bwrap` + `--die-with-parent` wipes `/tmp` between calls. The motivation in #530 reads cleanly. If you address Issue 1 (or push back with a reason it's fine), I think this is good to merge.

---

sabiut (approved):
Approving from a CODEOWNER perspective on `scripts/doctor.sh` and `/tests/`:

- `_doctor_check_bwrap_mounts` rendering of `src -> dst` is correct in both Python and Node parser branches; malformed-entry filtering via `fmt()` returning `null` mirrors `filterMounts` in `cowork-vm-service.js`.
- Verified `bats tests/cowork-bwrap-config.bats` locally: **58/58 pass**.
- Test 51 (`doctor: renders {src,dst} mounts as 'src -> dst'`) pins the rendering contract.

@RayCharlizard's Issue 1 (doctor warning when `dst` shadows a default RO mount like `/usr`, `/etc`, `/bin`, `/sbin`, `/lib*`) is a real footgun and worth addressing — but it applies to the existing string form too, so I'd rather track it as a follow-up than expand this PR's scope. I'll open an issue and pick it up myself.

No objection to merging once @RayCharlizard's other points (docs note, optional extra reject test) are resolved or punted.

cbonnissent:
Thanks for the careful review! Pushed `7cb1357` addressing all four points.

**Issue 1 — `dst` shadowing default RO mounts:** `_doctor_check_bwrap_mounts` now scans every `src -> dst` line from `additionalROBinds`/`additionalBinds` and emits `_warn` when `dst` is `/usr`, `/etc`, `/bin`, `/sbin`, `/lib`, `/lib64`, or a subpath. Pattern lifted from your snippet, with a short comment explaining why only the object form can trigger this (string form mounts `src=dst`, and `additionalBinds` requires `src` under `$HOME`, so it never overlaps the default RO set). I considered punting per @sabiut's suggestion, but the user preferred to land it here since the diff is tiny and the warning is genuinely useful.

**Issue 2 — docs gap:** added a `> **Caution:**` callout under "Distinct host/sandbox paths" that names the shadowed default mounts and points at the new `--doctor` warning.

**Issue 3 — reject test for `dst` under a forbidden path:** added `loadBwrapMountsConfig: rejects {src,dst} with dst under forbidden path` testing `dst: '/proc/self'`. As you noted, `validateMountPath` already covers it transitively, but the test now pins the contract.

**Issue 4 — warning text consistency:** harmonized on `BwrapConfig: rejected mount …` for all three variants. The pre-existing `loadBwrapMountsConfig: logs rejected paths` test only asserts on path-substring presence so it still passes — verified.

Test count: 58 → 61 passing (1 reject-subpath + 2 new doctor shadow tests, positive + negative).

Re-ran `bats tests/cowork-bwrap-config.bats` locally: 61/61. `shellcheck scripts/doctor.sh` clean (only the pre-existing SC2148 about the missing shebang on this sourced file).

---

sabiut (approved):
All four review points addressed:

- **Issue 1 (shadow warning):** ✅ Correct. The "only `{src, dst}` form can trigger this" reasoning checks out — string-form binds always mount `src=src`, so they produce a no-op redundancy rather than a content-replacing shadow. Narrowing the warning to `*' -> '*` lines is the right call.
- **Issue 3 (reject `dst` under forbidden path):** ✅ Pins the contract cleanly.
- **Issue 4 (warning text harmonization):** ✅ `rejected path` → `rejected mount` reads consistently across the three variants.

Verified locally: `bats tests/cowork-bwrap-config.bats` 61/61 pass, `shellcheck scripts/doctor.sh` clean.

**Approving.** Two non-blocking nits I'll fold into a follow-up rather than churn this PR:

1. **`docs/CONFIGURATION.md` contradiction** — Issue 2 added the `> **Caution:**` callout but left the existing bullet *"Mount order is enforced: user mounts cannot override security-critical read-only mounts"* unchanged. That bullet is only true for `{/, /proc, /dev, /sys}`; the new callout correctly says the opposite for `/usr`, `/etc`, `/bin`, `/sbin`, `/lib`, `/lib64`. Doc now disagrees with itself — worth a one-line correction.

2. **Restart-hint UX (minor)** — When the new shadow warning fires *without* a critical-disable, the user sees only the soft `Note: Restart daemon for config changes:` line. The shadow case is actionable (fix config + restart), so promoting it to the imperative `Restart the daemon after config changes:` arm — same one critical-disable already takes — would read better.

Both fit naturally in a slim follow-up alongside a few extra BATS cases (subpath shadow, `/usrlocal` boundary, RW object form) — I'll open it as #532.

---

RayCharlizard (approved):
Re-reviewed `7cb1357` against my four prior points. All landed cleanly:

- **Issue 1 (shadow warning):** added in [scripts/doctor.sh:274-296](https://github.com/aaddrick/claude-desktop-debian/blob/7cb1357/scripts/doctor.sh#L274-L296) with the case statement covering `/usr`, `/etc`, `/bin`, `/sbin`, `/lib`, `/lib64` and subpaths. Narrowed to ` -> ` lines per cbonnissent's reasoning. The string-form-RO gap (`additionalROBinds: ["/etc"]` has no `$HOME` constraint) is genuinely uncovered, but #532 has it tracked, so not blocking.
- **Issue 2 (docs note):** Caution callout added under "Distinct host/sandbox paths" in `docs/CONFIGURATION.md`. (sabiut's contradicting-bullet nit folds into #532.)
- **Issue 3 (reject test for `dst` under forbidden path):** new test pins `dst: '/proc/self'` rejection.
- **Issue 4 (warning text harmonization):** string form now reads `BwrapConfig: rejected mount "/x"` matching the object-form variants.

**Verified locally:**
- `bats tests/cowork-bwrap-config.bats` — 61/61 pass
- 3 new tests confirmed: 1 reject-subpath + 2 doctor shadow positive/negative
- CI green at HEAD (BATS, shellcheck, codespell, all 6 build matrix combos, all 3 artifact validations)

**Merging.** The unchecked e2e box in the test plan is acknowledged — author honestly punted because their PR session itself ran in the sandbox being modified, and the substitute validation (feeding the example config through `loadBwrapMountsConfig` + `mergeBwrapArgs` and confirming the bwrap arg output) covers most of the e2e signal. Unit tests + CI + sabiut's CODEOWNER approval + #532's residual-scope tracking is enough to land this. If anything surfaces in real use, we re-address.

Thanks for the careful iteration cbonnissent — and for the honest `[ ]` rather than a fake `[x]`.

---

PR #534, opened 2026-04-28
test: isolate cleanup_stale_cowork_socket BATS from host pgrep state

### LINKED ISSUE #533: BATS: cleanup_stale_cowork_socket test fails when a cowork-vm-service daemon is running on the host
## Symptom

`tests/launcher-common.bats:340` (`cleanup_stale_cowork_socket: removes stale socket file`) fails on any host with a live `cowork-vm-service.js` daemon (i.e., Claude Desktop running). Reproduces on `main`; not introduced by any specific PR.

```
not ok 34 cleanup_stale_cowork_socket: removes stale socket file
# (in test file tests/launcher-common.bats, line 355)
#   `[[ ! -S "$sock" ]]' failed
```

## Root cause

`cleanup_stale_cowork_socket` (`scripts/launcher-common.sh:224`) intentionally bails out early if a daemon is alive:

```sh
if pgrep -f 'cowork-vm-service\.js' &>/dev/null; then
    return 0
fi
```

The test creates a real unix socket but does not stub `pgrep`. On a developer machine running Claude Desktop, `pgrep` finds the live daemon, the function correctly returns without `rm`, and the assertion fails. CI passes only because no daemon runs there.

This is a **test isolation bug**, not a bug in `cleanup_stale_cowork_socket` — the function is doing the right thing.

## Proposed fix

Stub `pgrep` in the test's bash environment so it returns nonzero, isolating the test from host process state. Pattern:

```bash
@test "cleanup_stale_cowork_socket: removes stale socket file" {
    # ... existing setup ...
    pgrep() { return 1; }
    export -f pgrep
    cleanup_stale_cowork_socket
    [[ ! -S "$sock" ]]
}
```

---

sabiut:
## Summary

Fixes #533. The BATS test `cleanup_stale_cowork_socket: removes stale socket file` (`tests/launcher-common.bats:341`) fails on any developer machine running Claude Desktop because it never stubs `pgrep`.

`cleanup_stale_cowork_socket` (`scripts/launcher-common.sh:224`) intentionally bails out early if a `cowork-vm-service.js` daemon is alive — that's correct behavior. The test, however, was written to exercise the "no daemon → remove stale socket" branch but used real `pgrep`, so the developer's own running daemon would trip the early-return path and the assertion `[[ ! -S "$sock" ]]` would fail.

This PR stubs `pgrep` inside the test to always return nonzero, isolating the test from host process state.

## Test plan

- [x] `bats tests/launcher-common.bats --filter cleanup_stale_cowork_socket` — 2/2 pass with Claude Desktop running on the host (PID confirmed via `pgrep -af 'cowork-vm-service\.js'`)
- [x] `bats tests/*.bats` — 186/186 pass
- [x] `shellcheck -x tests/launcher-common.bats` — no new findings (preexisting SC2053 on line 16 unrelated)

### DIFF
--- a/tests/launcher-common.bats
+++ b/tests/launcher-common.bats
@@ -349,6 +349,12 @@ s.bind(sys.argv[1])
 s.close()
 " "$sock" 2>/dev/null || skip "Cannot create test unix socket"
 
+	# Stub pgrep so the test is isolated from host process state:
+	# a real cowork-vm-service daemon on the developer machine would
+	# trip the function's "daemon alive, leave socket alone" branch.
+	pgrep() { return 1; }
+	export -f pgrep
+
 	setup_logging
 	# socat connection should fail since nothing is listening
 	cleanup_stale_cowork_socket
--- a/tests/launcher-common.bats
+++ b/tests/launcher-common.bats
@@ -353,10 +353,8 @@ s.close()
 	# a real cowork-vm-service daemon on the developer machine would
 	# trip the function's "daemon alive, leave socket alone" branch.
 	pgrep() { return 1; }
-	export -f pgrep
 
 	setup_logging
-	# socat connection should fail since nothing is listening
 	cleanup_stale_cowork_socket
 	[[ ! -S "$sock" ]]
 }


aaddrick:
Hey @sabiut — thanks for catching this and turning it around the same day you filed #533. The fix is right: `cleanup_stale_cowork_socket` checks `pgrep -f 'cowork-vm-service\.js'` at `scripts/launcher-common.sh:233`, so on any developer box with Claude Desktop running the test sees a live daemon and skips the `rm -f`. Stubbing `pgrep` to return 1 isolates the test from host state.

Two small revisions before I merge:

1. Drop `export -f pgrep`. `cleanup_stale_cowork_socket` runs in the same shell as the test body, and bash function lookup beats PATH, so the export to child processes isn't doing anything here.
2. Drop the `# socat connection should fail since nothing is listening` comment on the line above the call. The implementation moved to `pgrep` a while back, so the socat reference is misleading now.

Heads up — @RayCharlizard filed #544 two days later with the same diagnosis and exactly those two cleanups already applied. You beat him to it by two days, so I'm taking yours and folding the cleanups in via this review. I'll close his as a duplicate with credit for the independent catch.

---

sabiut:
Pushed both cleanups in 34cd228:

- Dropped `export -f pgrep` — confirmed no-op as you described (function lookup beats PATH in the same shell)
- Dropped the misleading `# socat connection should fail since nothing is listening` comment

Verified locally with a live `cowork-vm-service.js` daemon on the host (PID 6222), so the stub is genuinely doing the isolation work the test claims. `bats tests/launcher-common.bats` → 48/48, shellcheck warning count unchanged.

Ready for merge whenever.

aaddrick:
Confirmed end-to-end. I reproduced #533 by prepending a fake `pgrep` to `PATH` that always exits 0 (stand-in for a live daemon) and ran the test against `d4db728`. It fails with the exact `[[ ! -S "$sock" ]]` assertion from the issue. Same fake `pgrep` against `34cd228` passes, because the bash function lookup at `tests/launcher-common.bats:344` shadows the PATH shim before `cleanup_stale_cowork_socket` ever resolves it.

The function stub is the right pattern here. `cleanup_stale_cowork_socket` calls `pgrep` directly in the same shell (`scripts/launcher-common.sh:233`), so function-name resolution wins. Dropping `export -f pgrep` was safe for the same reason: the export only matters when the consumer is in a child process, and there's no child here. The XRDP tests in `tests/launcher-xrdp-detection.bats` use a PATH shim because `loginctl` runs via `$(loginctl ...)` command substitution, which is a different beast.

I checked the rest of `launcher-common.bats` for the same class of leak. None of the other tests hit unstubbed `pgrep` / `pidof` / `loginctl` paths. `cleanup_stale_lock` uses `kill -0` against PIDs the test controls, and `cleanup_orphaned_cowork_daemon` isn't covered at all (separate gap, out of scope).

`bats tests/*.bats` 186/186 on the PR branch. Merging.

---

PR #536, opened 2026-04-28
docs: add multi-instance profiles guide

CyPack:
## Summary

Adds documentation and helper scripts for running **multiple isolated Claude Desktop instances in parallel** on the same Linux machine — useful for multi-monitor workflows and keeping separate accounts in their own windows.

The trick is straightforward: give each instance its own `--user-data-dir` so the Electron `SingletonLock` doesn't merge them, then **selectively symlink** the session-bearing subdirectories (`local-agent-mode-sessions/`, `claude-code/`, `claude-code-sessions/`, etc.) so all instances see the same chat history and Claude Code sessions, while keeping `Cookies`, `IndexedDB`, `Cache/`, `config.json` (incl. `oauth:tokenCache`), and `Singleton*` per-profile.

This is a **docs-only contribution** — no install/build flow changes, no patches to the main app.

## What's added

- `docs/multi-instance/README.md` — full walkthrough:
  - architecture (which dirs to share, which to keep isolated)
  - quick-start (3 commands)
  - per-profile color-coded icons (ImageMagick recipe)
  - GNOME dock favorites integration
  - the `claude://` link-handler quirk and the `xdg-mime` swap trick used during login
  - optional Firefox `handlers.json` `action: 1` tweak to skip the dialog
  - WM_CLASS post-launch fix (`xdotool set_window`) for desktops that won't pick up `--class` on the main BrowserWindow
  - anti-patterns and rollback steps
- `docs/multi-instance/claude-launch` — launcher script following the repo style guide (tabs, `[[ ]]`, lowercase vars, no `set -e`/`eval`). Subcommands: `1|2|3|N`, `all`, `status`, `kill <N>`, `kill-extras`. Includes a first-run-only `oauth:tokenCache` sync that does **not** overwrite an existing token (so users can log into different accounts per profile).
- `docs/multi-instance/claude-desktop-N.desktop.template` — sample `.desktop` entry with placeholders for `N` and the user's home path
- `docs/multi-instance/recolor-icon.sh` — ImageMagick recipe that produces the 6 standard hicolor sizes for a tinted profile icon
- One-line link from the main `README.md` Configuration section

## Test plan

- [x] All scripts pass `bash -n` and `shellcheck` cleanly
- [x] Tested on Fedora 43 + GNOME on Wayland with claude-desktop wrapper v2.0.5+ (Electron 41) running 4 parallel profiles
- [x] Each profile launches with its own `WM_CLASS` and gets its own taskbar icon (after applying the `xdotool` post-launch tweak documented in the guide for the main BrowserWindow)
- [x] `claude://` callbacks reach the intended profile via the temporary `xdg-mime default claude-desktop-N.desktop x-scheme-handler/claude` swap during login
- [x] `claude-launch kill N` and `kill-extras` correctly target only side profiles, leave the main untouched
- [x] No hardcoded usernames, emails, UUIDs, tokens, hostnames, or PIDs in any of the new files (paths use `${HOME}` / `$USER`)

Happy to iterate on tone, structure, or scope — for example, splitting the script into `scripts/` if that's preferred over keeping everything under `docs/multi-instance/`.

---
85% AI / 15% Human
Claude: drafted the documentation, the `claude-launch` script, and the recolor recipe based on the user's session-by-session debugging of multi-profile parallelism. Ran shellcheck, syntax checks, and a credential-leak audit before commit.
Human: discovered the requirement, validated the symlink boundary and the `xdg-mime` swap pattern in practice on their own machine across all four profiles, and reviewed every step before approving the push.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,8 @@ Model Context Protocol settings are stored in:
 
 For additional configuration options including environment variables and Wayland support, see [docs/CONFIGURATION.md](docs/CONFIGURATION.md).
 
+For running multiple isolated profiles in parallel (one per monitor, separate accounts, color-coded taskbar icons), see [docs/multi-instance/README.md](docs/multi-instance/README.md). Includes a launcher script, sample `.desktop` template, icon recolor recipe, and notes on the `claude://` link-handler quirk.
+
 ## Troubleshooting
 
 Run `claude-desktop --doctor` for built-in diagnostics that check common issues (display server, sandbox permissions, MCP config, stale locks, and more). It also reports cowork mode readiness — which isolation backend will be used, and which dependencies (KVM, QEMU, vsock, socat, virtiofsd, bubblewrap) are installed or missing.
--- a/docs/multi-instance/README.md
+++ b/docs/multi-instance/README.md
@@ -1,0 +1,246 @@
+# Multi-Instance Claude Desktop Profiles
+
+Run **multiple isolated Claude Desktop instances in parallel** on the same Linux machine — each with its own login, its own window, its own taskbar icon — while sharing local Claude Code session history, agent-mode sessions, and other read-mostly state across all profiles.
+
+> Tested on Fedora 43 with GNOME on Wayland, claude-desktop wrapper v2.0.5+ (Electron 41). Should work on any Linux desktop where this RPM/DEB build runs.
+
+---
+
+## Why?
+
+The Claude Desktop Electron app uses a `SingletonLock` in its `userData` directory. Launching it twice with the same `userData` just focuses the existing window instead of opening a second one. This guide shows how to run **N parallel instances** by giving each its own `userData` directory, while symlinking the *session-bearing* subdirectories so all instances see the same chat history and Claude Code sessions.
+
+Useful for:
+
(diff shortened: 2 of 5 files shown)


aaddrick:
Hey @CyPack! 

Appreciate the work here, and the offer to move the script under `scripts/`.

The core reason I'm going to pass is project policy. I don't want to add features that aren't in upstream Claude Desktop, unless they're compensating for something missing across Linux environments. Multi-instance isn't a Linux gap. Claude Desktop doesn't support multiple instances on any platform, so this would put us ahead of upstream in a way I try to avoid.

On the technical side, the symlink scheme has a real concurrency risk worth flagging. `local-agent-mode-sessions/` holds agent-mode session state plus a skills-plugin sync layer that runs on a 600s interval, guarded by an in-process Promise mutex. That mutex doesn't cross processes, so two instances pointed at the same directory each run their own sync independently. Sessions are keyed by `sessionId`, so distinct sessions are fine on disk, but the whole point of sharing the dir is shared session history. Two instances opening the same session at once is exactly where you'd see corruption. The `oauth:tokenCache` hand-sync into `config.json` has the same shape of issue.

The right home for this is the Show & Tell discussion category. Feel free to repost the writeup and the script there.

Issue #554, opened 2026-05-01
Security review: CI/CD shell injection + path traversal findings


HumboldtJoker:
## Context

We've been working with you on #545 and #553 and figured a security review of the codebase would be a useful contribution while we're here. Ran Semgrep with `--config=auto` — 33 findings, 7 ERROR severity. Most are actionable.

This is a neighborly heads-up, not a bounty report. We use this project and want it to be solid.

## HIGH: GitHub Actions Shell Injection (6 findings)

Direct `${{ }}` interpolation in `run:` blocks allows shell injection via workflow inputs.

**build-amd64.yml:45 and build-arm64.yml:45:**
```yaml
run: |
  ./build.sh ${{ inputs.build_flags }} $TAG_FLAG
```

**ci.yml:488:**
```yaml
git commit -m "Update APT repository for ${{ github.ref_name }}"
```

**Risk:** A collaborator (or anyone with `workflow_dispatch` access) can inject arbitrary shell commands via `build_flags` or `release_tag` inputs. `ref_name` is lower risk but still injectable via crafted branch names.

**Fix:** Use environment variables instead of direct interpolation:
```yaml
env:
  BUILD_FLAGS: ${{ inputs.build_flags }}
  RELEASE_TAG: ${{ inputs.release_tag }}
run: |
  TAG_FLAG=""
  if [[ -n "$RELEASE_TAG" ]]; then
    TAG_FLAG="--release-tag $RELEASE_TAG"
  fi
  ./build.sh $BUILD_FLAGS $TAG_FLAG
```

Same pattern for `ref_name`:
```yaml
env:
  REF_NAME: ${{ github.ref_name }}
run: |
  git commit -m "Update APT repository for ${REF_NAME}"
```

## MEDIUM: Path Traversal in cowork-vm-service.js (14 findings)

User-controlled input flows into `path.join()` / `path.resolve()` without traversal checks at multiple points (lines 223, 224, 248, 252, 365, 366, 506, 509, 575, 692, 809, 816).

**Risk:** Lower severity since the service runs locally and the input comes from the Electron app's IPC, not from network. However, if the IPC protocol is ever exposed or a malicious extension sends crafted paths, it could escape the intended sandbox directories.

**Fix:** Validate that resolved paths remain within the expected root:
```javascript
function safePath(root, userPath) {
  const resolved = path.resolve(root, userPath);
  if (!resolved.startsWith(root + path.sep) && resolved !== root) {
[...]

Issue #553, opened 2026-05-01
Cowork VM service fails to initialize on Linux — Swift VM addon not available


HumboldtJoker:
## Bug Description

The Cowork workspace feature fails to start on Linux with: `Swift VM addon not available`. The app correctly identifies the platform and attempts to load the TypeScript VM client, but session initialization still fails through the Swift path.

Additionally, the app is unstable on systems with older NVIDIA GPUs — the GPU process crashes cause blank windows that require cache clearing to recover from.

## Related Issue

This is a follow-up to #545 (keyboard input fix). That issue's workaround (`GTK_IM_MODULE=xim --no-sandbox --disable-gpu`) resolves keyboard input and GPU crashes, but Cowork remains broken.

## Environment

- **OS:** Ubuntu 24.04 (x86_64, X11, GNOME)
- **Package version:** claude-desktop 1.5354.0-2.0.5 (amd64)
- **GPU:** NVIDIA Quadro K2200 (4GB, driver 535.288.01)
- **Launch flags:** `--no-sandbox --disable-gpu`

## Error Logs

From `~/.config/Claude/logs/main.log`:
```
[ScheduledTasks] VM not ready (tick 15), requesting startVM
[ScheduledTasks] startVM failed: Error: Swift VM addon not available
```

Repeats every 5 minutes indefinitely.

```
Session initialization failed for local_ditto_...: Swift VM addon not available
```

## Analysis

The app code at the VM loader function has:
```javascript
const e = (ys || process.platform === "linux") ? "vmClient (TypeScript)" : "@ant/claude-swift";
```

Where `ys = process.platform === "win32"`. So Linux correctly selects the TypeScript VM client path. However, session initialization at a different code path still calls through to the Swift addon and fails.

The native binding at `app.asar.unpacked/node_modules/@ant/claude-native/claude-native-binding.node` is a **Windows PE32+ DLL** — it cannot load on Linux regardless.

## What We Tried

1. The `cowork-vm-service.js` daemon runs fine when started manually — creates socket at `$XDG_RUNTIME_DIR/cowork-vm-service.sock`, bwrap backend available
2. The app doesn't connect to the manually started service
3. Various `--no-sandbox` configurations don't help
4. `COWORK_VM_BACKEND=host` and other env vars don't bypass the Swift check

## Additional Instability

The `--disable-gpu` flag is required for the app to run without blank-window crashes on this GPU (Quadro K2200). Without it:
```
[...]

Issue #542, opened 2026-04-30
bwrap fallback: Claude-owned wrapper binary for a scoped AppArmor profile


RayCharlizard:
## Context

[@slovdahl](https://github.com/slovdahl) [pointed out on PR #434](https://github.com/aaddrick/claude-desktop-debian/pull/434#issuecomment-4352273336) that the AppArmor profile we recommend in `docs/TROUBLESHOOTING.md` (added in #434 to fix Ubuntu 24.04+ bwrap blockage from #351) attaches `flags=(unconfined)` to `/usr/bin/bwrap`, which grants user-namespace creation to **every program on the system that invokes bwrap**, not just Claude Desktop. Their quoted source is the AppArmor maintainer asking opam to do the same thing differently.

## Current state

[`docs/TROUBLESHOOTING.md:107-120`](https://github.com/aaddrick/claude-desktop-debian/blob/main/docs/TROUBLESHOOTING.md#L107-L120) tells users to install:

```
profile bwrap /usr/bin/bwrap flags=(unconfined) {
    userns,
    include if exists <local/bwrap>
}
```

Effect: any process on the box that execs `/usr/bin/bwrap` (Flatpak runtimes, Steam, the user's own scripts) runs unconfined and can create user namespaces.

## Why we can't just point users at the upstream `bwrap-userns-restrict` profile

@slovdahl's [follow-up comment](https://github.com/aaddrick/claude-desktop-debian/pull/434#issuecomment-4352458235) noted it "doesn't work out of the box," and the history explains why: Ubuntu briefly shipped `bwrap-userns-restrict` as a default profile in `apparmor 4.0.1-0ubuntu0.24.04.2`, and reverted it because it broke Flatpak file saves ([LP #2072811](https://bugs.launchpad.net/ubuntu/+source/apparmor/+bug/2072811) — affected KeePassXC, Steam, Telegram, Ksnip via broken `O_TMPFILE`+linkat atomic writes). The same broken profile sits unloaded in `/usr/share/apparmor/extra-profiles/` today. Recommending users symlink and load it would re-introduce the breakage.

## The pattern that fits — per-application profile

Both [opam](https://gitlab.com/apparmor/apparmor/-/blob/master/profiles/apparmor.d/opam) (upstream apparmor reference) and [Apptainer](https://github.com/apptainer/apptainer/pull/2262) ship the same shape, but attached to **their own application binary** rather than to `bwrap`:

```
[...]

Issue #537, opened 2026-04-29
[bug]: Cowork project creation fails silently — folder created but process never completes


danisalgueiro:
### Version (`claude-desktop --doctor` output)

```shell
Claude Desktop Diagnostics
================================

[PASS] Installed version: 1.5220.0-2.0.5
[PASS] Display server: Wayland (WAYLAND_DISPLAY=wayland-0)
       Desktop: Unity
       Mode: X11 via XWayland (default, for global hotkey support)
       Tip: Set CLAUDE_USE_WAYLAND=1 for native Wayland
            (disables global hotkeys)
       Menu bar mode: auto (default, Alt toggles visibility)
[PASS] Electron: v41.3.0 (/usr/lib/claude-desktop/node_modules/electron/dist/electron)
[PASS] Chrome sandbox: permissions OK (/usr/lib/claude-desktop/node_modules/electron/dist/chrome-sandbox)
[PASS] SingletonLock: held by running process (PID 437639)
[PASS] MCP config: valid JSON (/home/USER/.config/Claude/claude_desktop_config.json)
       MCP servers configured: 0
[PASS] Node.js: v22.17.0
       Path: /home/USER/.nvm/versions/node/v22.17.0/bin/node
[PASS] Desktop entry: /usr/share/applications/claude-desktop.desktop
[PASS] Disk space: 284230MB free

Cowork Mode
----------------
[PASS] bubblewrap: found
[PASS] bubblewrap: sandbox probe succeeded
[PASS] KVM: accessible
[PASS] vsock: module loaded
[PASS] QEMU: found
[PASS] socat: found
[PASS] virtiofsd: found
       VM image: not downloaded yet
       Cowork isolation: bubblewrap (namespace sandbox)
[PASS] Cowork daemon: running (parent alive)
[PASS] Log file: 270KB (/home/USER/.cache/claude-desktop-debian/launcher.log)

All checks passed.
```

## Environment
- OS: Ubuntu 24.04
- Package: claude-desktop (aaddrick/claude-desktop-debian, .deb)
- Claude Desktop version: [run `claude-desktop --version` and fill in]

## Description
When trying to create a new project in Cowork, the folder is created on disk
but the process never completes. The UI shows an error and asks to retry,
but retrying always produces the same result.

## What works
- Normal Cowork sessions (Tasks, Code) work correctly
- Chat projects work fine
- `claude-desktop --doctor` passes all checks

## What doesn't work
[...]

PR #555, opened 2026-05-01
fix(cowork.sh): allow $ in minified identifier anchors; defensive lastIndexOf

sirfaber:
On Claude Desktop 1.5354.0, three regex anchors in patch_cowork_linux()
no longer match the upstream minified output, shipping a half-patched
app.asar that fails Cowork startup with "Swift VM addon not available".

Patch 2b (vm: module assignment) and Patch 6 step 2 (retry-delay
auto-launch) silently no-op because the minifier emits identifiers
containing '$' (e.g. C$i, Z$x). JS \w excludes '$', so the inner
captures never match. Widening both inner captures to [\w$]+ fixes
both.

Patch 6 also switched from indexOf to lastIndexOf for serviceErrorStr.
In 1.5354.0 the string occurs once, so the change is a no-op. It is
defensive against future upstream versions that reintroduce the
string in onboarding/sample text far from the real retry-loop site.

Symptoms before fix on 1.5354.0:
- ~/.config/Claude/logs/cowork_vm_node.log: load_swift_api failed
- ~/.config/Claude/logs/cowork_vm_daemon.log: never created
- pgrep -af cowork-vm-service: nothing
- ${XDG_RUNTIME_DIR}/cowork-vm-service.sock: missing

Build log before:  Applied 10 cowork patches (no 2b success line,
"WARNING: Could not find retry delay for auto-launch patch")
Build log after:   Applied 12 cowork patches ("Patched VM module
assignment for Linux", "Added service daemon auto-launch on Linux")

Tested on Fedora 43 Workstation (Wayland/GNOME), x86_64, Cowork
backend bwrap. After rebuild + reinstall + relaunch, the in-app
workspace shell starts and the daemon log is populated as expected.

Two pre-existing warnings remain on this build but are not bugs:
"Could not find reinstall file list array" and "Could not find
mkdtemp("wvm-") for tmpdir patch". Both anchors target upstream
code structures that were refactored in 1.5354.0 (rootfs.img is now
inside Eo.files manifest; "wvm-" was renamed to a constant VHe and
the call already uses the bundle dir). Out of scope for this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,10 +250,14 @@ Special thanks to:
 - **[hfyeh](https://github.com/hfyeh)** for diagnosing the Ubuntu 24.04 AppArmor unprivileged-userns block on Cowork bwrap and contributing the AppArmor profile workaround
 - **[davidamacey](https://github.com/davidamacey)** for identifying and fixing the XRDP GPU compositing blank-window issue on remote desktop sessions
 - **[pb3ck](https://github.com/pb3ck)** for diagnosing the Cowork `CLAUDE_CODE_OAUTH_TOKEN` env-strip bug with a working reference diff
+- **[Joost-Maker](https://github.com/Joost-Maker)** for fixing the `$e` fs reference crash in cowork Patch 9 on Claude Desktop 1.3109.0, introducing the `[$\w]+` identifier-capture pattern at `cowork.sh:482-501` (#421)
 - **[aJV99](https://github.com/aJV99)** for exporting `GDK_BACKEND=wayland` in native Wayland mode to fix XWayland fallback blur on HiDPI displays
 - **[Andrej730](https://github.com/Andrej730)**
   - Quick-window regex readability refactor (`String.raw` + `escapeRegExp` helper)
-  - Fixing the visibility-function regex break on Claude Desktop 1.3883.0 (#495)
+  - Fixing the visibility-function regex break on Claude Desktop 1.3883.0 (#496)
+- **[HumboldtJoker](https://github.com/HumboldtJoker)** for diagnosing the cowork Patch 2b silent failure on Claude Desktop 1.5354.0 — identifying that the log line was patched but session init still routed through the Swift addon (#553)
+- **[zabka](https://github.com/zabka)** for identifying that `cowork-vm-service.js` was never auto-spawned on Linux and contributing a systemd-unit workaround that scoped the daemon auto-launch fix (#445)
+- **[sirfaber](https://github.com/sirfaber)** for fixing the `$`-in-minified-identifier breakage of cowork Patch 2b (vm module assignment) and Patch 6 step 2 (retry-delay auto-launch) on Claude Desktop 1.5354.0 (#555)
 
 ## Sponsorship
 
--- a/scripts/patches/cowork.sh
+++ b/scripts/patches/cowork.sh
@@ -118,7 +118,7 @@ if (vmClientLogMatch) {
     const assignRe = new RegExp(
         '(?<!\\|\\|process\\.platform==="linux"\\)?)' +
         win32Var.replace(/[.*+?^${}()|[\]\\]/g, '\\$&') +
-        '(\\s*\\?\\s*\\(?\\s*\\w+\\s*=\\s*\\{\\s*vm\\s*:\\s*\\w+\\s*\\}\\s*\\)?)'
+        '(\\s*\\?\\s*\\(?\\s*[\\w$]+\\s*=\\s*\\{\\s*vm\\s*:\\s*[\\w$]+\\s*\\}\\s*\\)?)'
     );
     if (assignRe.test(code)) {
         code = code.replace(assignRe,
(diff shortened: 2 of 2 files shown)


aaddrick:
Hey @sirfaber! 

Thanks for this. Initial checks confirms your findings. I'm working through a review now.

aaddrick (commented):
Pulled 1.5354.0 from the URL pinned in `scripts/setup/detect-host.sh` and confirmed the installer SHA-256 against `claude_exe_sha256` (`e619c7bd...`) before extracting. Worked off the real minified `index.js` (sha256 `6c4a8197426e16dd46c81148c42cb3f21222d5d0cce9d84cf29f48ad189882d7`), not the beautified `build-reference/` copy — early on I got false negatives there because of whitespace, which is worth flagging for anyone else trying to repro.

Ran both regex anchors against the actual bytes. Patch 2b matches the assignment `ys?JN={vm:C$i}` (where `ys` is the extracted win32 platform var) and Patch 6 step 2 matches `await new Promise(n=>setTimeout(n,g$i))`. Old `\w+` fails on both, new `[\w$]+` succeeds on both. Without this fix Patch 2b no-ops without even a warning and Patch 6 step 2 emits one — either way the resulting app.asar ships half-patched.

The `lastIndexOf` switch is a no-op on 1.5354.0 — `serviceErrorStr` appears exactly once — so I read it as defensive in the direction @sirfaber describes. Pattern matches the existing precedent at lines 482-501, where we already use `[$\w]+` for the same `$`-in-identifier reason (PR #421 fixing issue #418, and PR #253 before that).

Since drafting, I ran the full pipeline. Build log shows `Applied 12 cowork patches` (was `Applied 10` in the broken state per @sirfaber's PR description), and the two previously-failing patches now report `Patched VM module assignment for Linux` and `Added service daemon auto-launch on Linux`. The two remaining warnings (`Could not find reinstall file list array`, `Could not find mkdtemp("wvm-") for tmpdir patch`) line up with the PR description's out-of-scope notes — both are upstream refactors in 1.5354.0 (rootfs.img is now inside the `Eo.files` manifest, `wvm-` was renamed to `VHe`).

Extracted the built AppImage's `app.asar` and confirmed all 9 cowork patch markers are present in the shipped `index.js`: vm-assignment linux gate, ECONNREFUSED-on-linux, `__coworkDaemonPid` global, `cowork-linux-daemon-shutdown` handler, vmClient log gate, Unix socket path, empty linux bundle manifest, getDownloadStatus suppression, sharedCwdPath thread-through.

Then launched the AppImage on Fedora 43 / KDE Plasma 6 / Wayland (different DE+compositor than @sirfaber's Fedora 43 / GNOME / Wayland, worth noting). Cowork daemon auto-launched at app startup. Daemon log:

```
[cowork-vm-service] startup pid=65453 sock=/run/user/1000/cowork-vm-service.sock
[cowork-vm-service] Service started on /run/user/1000/cowork-vm-service.sock
[cowork-vm-service] listening /run/user/1000/cowork-vm-service.sock
```

Socket was bound and owned by the daemon PID (verified via `ss -lpx`). On app SIGTERM, Patch 10's quit handler fired and the daemon exited with `exit code=0`. Full lifecycle (startup → listen → graceful shutdown) is the path that was broken before this PR. I didn't click into a Cowork workspace and run an actual VM session, so I can't speak to the deeper IPC. But the daemon path PR #555 unblocks is verifiably alive end-to-end.

Filed https://github.com/aaddrick/claude-desktop-debian/issues/559 as follow-up — three recurrences of this `$`-in-minified-identifier bug class is enough signal to document. The issue covers the docs + CLAUDE.md example fix and explicitly default-declines a preemptive sweep of remaining `\w+` sites.

Still want @RayCharlizard's read before landing — cowork is his area.

---

aaddrick:
## Affected issues

Cross-referencing the open `cowork`-labeled issues against this PR's scope (`\w+` → `[\w$]+` in two anchors + defensive `lastIndexOf`):

**Directly fixed:**

- **#558** — Cowork sandbox fails on 1.5354.0-2.0.7, `Swift VM addon not available`, no bubblewrap fallback. The named target of this branch.
- **#553** — Same symptom on 1.5354.0-2.0.5. The reporter independently diagnosed it: *"the log/description line was patched, but session initialization still calls through to the Swift addon"* — i.e. Patch 2a ran, Patch 2b silently no-op'd. That's exactly the `[\w$]+` gap this PR closes.
  - Note: #553 also bundles an unrelated NVIDIA Quadro K2200 GPU-crash complaint (already worked around with `--disable-gpu`). Out of scope here.

**Possibly fixed (pending confirmation):**

- **#445** — "Failed to start workspace", workaround is manually spawning `cowork-vm-service.js` via systemd. Reporter's diagnosis: *"cowork-vm-service.js is never auto-spawned on Linux"*. The Patch 6 step 2 fix in this PR restores the daemon auto-launch injection that was silently failing — the build log gains the `Added service daemon auto-launch on Linux` line only after the fix. If #445's installed version had `$` in retry-delay identifiers, this PR likely fixes it too. The issue doesn't pin a version, so worth a re-test post-merge before closing.

**Not addressed (different scope, listed for completeness):** #554, #542, #537, #533, #532, #526, #442, #383, #378, #369, #297.

---

---

aaddrick:
Hey @RayCharlizard,

I'm going to merge this one without your review as it's addressing an active break.

PR #562, opened 2026-05-02
Set desktopName for Wayland grouping

### LINKED ISSUE #561: KDE Wayland: pinned Claude launcher opens a separate generic Electron taskbar entry
## Summary

On KDE Plasma running a native Wayland session, launching the pinned `claude-desktop` taskbar icon opens a separate generic Electron taskbar entry instead of grouping under the pinned Claude launcher.

The installed desktop entry and icon are already correct, but the runtime app identity exposed by Electron does not fully line up with the desktop file metadata that Plasma uses for grouping.

## Environment

- Desktop environment: KDE Plasma
- Session type: Wayland (`XDG_SESSION_TYPE=wayland`)
- `XDG_CURRENT_DESKTOP=KDE`
- Claude Desktop installed from this repository

## Reproduction

1. Pin `Claude` to the KDE taskbar.
2. Launch it from the pinned icon.
3. Observe that Plasma opens a second taskbar entry with the generic Electron icon instead of reusing the pinned Claude launcher.

## What I found

The desktop file itself looks correct at a glance:

```ini
[Desktop Entry]
Name=Claude
Exec=/usr/bin/claude-desktop %u
Icon=claude-desktop
StartupWMClass=Claude
```

The icon assets are also present in the expected icon theme paths.

However, the running app is using native Wayland, and on Wayland Plasma groups windows using the app identity / desktop file name exposed by the client, not just the icon field in the launcher.

The current launch path starts Electron like this:

```text
/usr/lib/claude-desktop/node_modules/electron/dist/electron \
  --disable-features=CustomTitlebar \
  --no-sandbox \
  --enable-features=UseOzonePlatform,WaylandWindowDecorations \
  --ozone-platform=wayland \
  --enable-wayland-ime \
  --wayland-text-input-version=3 \
  /usr/lib/claude-desktop/node_modules/electron/dist/resources/app.asar
```

Inside the packaged app, `resources/app.asar` contains a `package.json` with:

```json
{
  "name": "@ant/desktop",
  "productName": "Claude"
}
```

and no `desktopName` field.

That appears to be the root problem for native Wayland. Electron derives the desktop-file identity from packaged app metadata, and the current metadata does not match `claude-desktop.desktop`.

## Fix needed

The Linux packaging/repackaging step should set the packaged app's desktop file name explicitly so native Wayland compositors can match the running window to the installed desktop entry.

In practice, the package should ensure that the app metadata inside `app.asar` includes:

```json
{
  "desktopName": "claude-desktop.desktop"
}
```

## Additional cleanup

For consistency across both native Wayland and X11/XWayland, it would also be good to make the window class / launcher metadata agree everywhere:

- `claude-desktop.desktop` basename
- Wayland app id / Electron desktop name
- `StartupWMClass`
- any explicit Chromium/Electron `--class=...` flag if one is used

Right now a launcher-only tweak is not sufficient on Wayland; the packaged app metadata needs to be fixed.

## Suggested validation

After patching the package metadata, verify on KDE Plasma Wayland that:

1. Pinning `Claude` to the taskbar
2. Launching it from the pinned launcher
3. Reuses the pinned Claude taskbar entry
4. Shows the Claude icon rather than a separate generic Electron icon

## Relevant references

- Electron docs note that Linux desktop integration can depend on the `desktopName` field in `package.json`.
- Wayland `xdg_toplevel.set_app_id` recommends matching the basename of the `.desktop` file so compositors can group windows correctly.

I confirmed locally that adding `desktopName: "claude-desktop.desktop"` to the packaged app metadata is the missing Claude-specific fix, whereas changing only the desktop file/icon was not enough.

jslatten:
## Summary

This sets `desktopName` in the packaged `app.asar` `package.json` during the Linux repack step so native Wayland compositors can match the running Electron app to `claude-desktop.desktop`.

## Root cause

The repack step already rewrites `app.asar.contents/package.json` for Linux-specific behavior, but it did not set Electron's Linux desktop file metadata. On KDE Plasma Wayland, that leaves the running app without a matching desktop-file identity and Plasma can group it as a separate generic Electron app instead of reusing the pinned Claude launcher.

## Changes

- set `pkg.desktopName = 'claude-desktop.desktop'` in `scripts/patches/app-asar.sh`
- add an artifact-level regression assertion in `tests/test-artifact-common.sh`

## Validation

- `bash -n scripts/patches/app-asar.sh`
- `git diff --check`

Full package build/runtime validation was not run in this checkout.

Closes #561.

### DIFF
--- a/scripts/patches/app-asar.sh
+++ b/scripts/patches/app-asar.sh
@@ -37,16 +37,21 @@ EOFENTRY
 
 	# Update package.json
 	echo 'Modifying package.json to load frame fix and add node-pty...'
+	local desktop_name='claude-desktop.desktop'
+	if [[ ${build_format:-} == 'appimage' ]]; then
+		desktop_name='io.github.aaddrick.claude-desktop-debian.desktop'
+	fi
 	node -e "
 const fs = require('fs');
 const pkg = require('./app.asar.contents/package.json');
 pkg.originalMain = pkg.main;
 pkg.main = 'frame-fix-entry.js';
+pkg.desktopName = process.argv[1];
 pkg.optionalDependencies = pkg.optionalDependencies || {};
 pkg.optionalDependencies['node-pty'] = '^1.0.0';
 fs.writeFileSync('./app.asar.contents/package.json', JSON.stringify(pkg, null, 2));
-console.log('Updated package.json: main entry and node-pty dependency');
-"
+console.log('Updated package.json: main entry, desktopName, and node-pty dependency');
+" "$desktop_name"
 
 	# Create stub native module
 	echo 'Creating stub native module...'
--- a/tests/test-artifact-appimage.sh
+++ b/tests/test-artifact-appimage.sh
@@ -94,7 +94,7 @@ assert_contains "$appdir/AppRun" 'build_electron_args' \
 
 # --- App contents (asar) ---
 resources_dir="$appdir/usr/lib/node_modules/electron/dist/resources"
-validate_app_contents "$resources_dir"
+validate_app_contents "$resources_dir" "${component_id}.desktop"
 
 # --- Cleanup ---
 rm -rf "$extract_dir"
--- a/tests/test-artifact-common.sh
+++ b/tests/test-artifact-common.sh
@@ -59,8 +59,10 @@ assert_command_succeeds() {
 
 # Validate app contents inside an Electron resources directory.
 # $1 = path to the resources/ dir containing app.asar
+# $2 = expected desktopName in app/package.json
 validate_app_contents() {
 	local resources_dir="$1"
+	local expected_desktop_name="${2:-claude-desktop.desktop}"
 
 	assert_file_exists "$resources_dir/app.asar"
 	assert_dir_exists "$resources_dir/app.asar.unpacked"
@@ -94,6 +96,11 @@ validate_app_contents() {
 		assert_contains "$extract_dir/app/package.json" \
 			'frame-fix-entry.js' \
 			"package.json main field references frame-fix-entry.js"
+
+		# package.json desktopName matches the installed desktop file
+		assert_contains "$extract_dir/app/package.json" \
+			"\"desktopName\": \"$expected_desktop_name\"" \
+			"package.json desktopName matches $expected_desktop_name"
 
 		# .vite/build/index.js exists (main process code)
 		assert_file_exists "$extract_dir/app/.vite/build/index.js"


sabiut (approved):
Test assertion LGTM — verified locally that it correctly fails against an unpatched package.json and passes against the PR #562-patched output. The test is well-placed inside `validate_app_contents` next to the other package.json field assertions.

**Non-blocking heads-up on the AppImage side** (outside my CODEOWNERS scope, flagging for @aaddrick):

The patch hardcodes `pkg.desktopName = 'claude-desktop.desktop'`, but the actual installed `.desktop` filename varies by packaging format:

| Format | Installed `.desktop` filename | Source | Matches `desktopName`? |
|--------|-------------------------------|--------|------------------------|
| deb | `claude-desktop.desktop` | `scripts/packaging/deb.sh:78` | ✓ |
| rpm | `claude-desktop.desktop` | `scripts/packaging/rpm.sh:62` | ✓ |
| AppImage | `io.github.aaddrick.claude-desktop-debian.desktop` | `scripts/packaging/appimage.sh:19,125` | ✗ |

Electron's `desktopName` tells the compositor "this running app corresponds to *that* `.desktop` file" — Wayland uses the basename (without `.desktop`) as the `app_id` for window grouping, taskbar icon lookup, and pinning. When the asar's `desktopName` doesn't match the installed file's basename, the compositor can't link the window to the entry, so:

- **deb / rpm users:** Wayland grouping works as intended ✓
- **AppImage users:** the patch is a no-op — Wayland still falls back to a generic `app_id` (typically the executable name), so windows don't group with a pinned launcher and the desktop-entry icon isn't picked up

Two options if you want full coverage:
1. Make the patch format-conditional — write `claude-desktop.desktop` for deb/rpm builds and `io.github.aaddrick.claude-desktop-debian.desktop` for AppImage builds. Requires the patch to know which format is being built (currently it doesn't).
2. Accept AppImage Wayland grouping as best-effort and document the deb/rpm-first behavior. AppImage users on Wayland are a smaller slice and AppImage `.desktop` integration is already opt-in.

The test as written still passes for AppImage builds because it only checks the asar field, not cross-format consistency — so the test isn't masking anything; this is purely a patch-side decision. Happy to extend the test if you go with option 1 (e.g. derive the expected value from the format being validated).

jslatten:
@sabiut Thank you for the feedback. I pushed an additional change that hopefully should work for both. I ran an appimage build and the test seems to be passing successfully.. Not my normal packaging method so hopefully it works as expected for everyone. On my local system it appears to be grouping the icon correctly when I run the appimage.

aaddrick (approved):
@jslatten the format-conditional rev is the right outcome of @sabiut's review. The current revision branches on `${build_format:-}` and writes the basename that each packaging format actually installs, so the Wayland `xdg_toplevel.app_id` lines up with `claude-desktop.desktop` for deb/rpm and with `io.github.aaddrick.claude-desktop-debian.desktop` for AppImage.

I built all three formats (deb/rpm/appimage) on a Fedora 43 host, extracted the resulting app.asar, and inspected `package.json` directly. deb wrote `claude-desktop.desktop`; rpm wrote `claude-desktop.desktop`; appimage wrote `io.github.aaddrick.claude-desktop-debian.desktop`. Each value matches the basename its packaging script installs (`scripts/packaging/deb.sh:78`, `rpm.sh:62`, `appimage.sh:125`). The new assertion in `validate_app_contents` fired correctly on each artifact (`[PASS] package.json desktopName matches <expected>`). One unrelated rpm test failure on `rpm -ivh` (sandbox lock path) is pre-existing environment noise.

I didn't have nix on this host, so I didn't build that path. But `patch_app_asar` runs for nix builds too (`build.sh:284`, before the `if [[ $build_format != 'nix' ]]` branches kick in), and the else-branch writes `claude-desktop.desktop` — which matches `nix/claude-desktop.nix:51`'s `makeDesktopItem name = "claude-desktop"`, so nix Wayland users should pick up the same fix transparently.

End-to-end Wayland-grouping confirmation depends on the compositor and integration tooling (Plasma vs GNOME, Gear Lever vs appimaged for AppImage), and @jslatten confirmed grouping on their KDE Plasma Wayland setup. Merging.

aaddrick:
Hey @typedrat — heads up that #562 just landed and `patch_app_asar` runs ahead of the format split in build.sh, so the nix path picks this up too. The else-branch writes `claude-desktop.desktop`, which lines up with `makeDesktopItem name = "claude-desktop"` in `nix/claude-desktop.nix`, so KDE Plasma Wayland users on the nix build should get the same window-to-launcher matching fix transparently.

I only verified this by static reading — I didn't actually build the nix output on my end (no NixOS host handy). Let me know if anything weird shows up on that side and I'll take another look.

Issue #481, opened 2026-04-21
[BUG] "About This App" dialog shows minified JavaScript error instead of version info


Hayao0819:
## Description

Clicking the "About This App" menu item opens a dialog that displays minified JavaScript source code (OpenTelemetry instrumentation code) instead of the expected version/about information.

The displayed code contains constants from `@opentelemetry/instrumentation` packages (METRIC_SIGNALR_SERVER, METRIC_KESTREL, METRIC_JVM, etc.) and instrumentation code for Redis, MySQL2, and ioredis.

## Steps to Reproduce

1. Launch Claude Desktop
2. Click the menu / tray icon
3. Select "About This App"
4. Observe: dialog shows minified JS code instead of version info

## Expected Behavior

The About dialog should display application name, version, and other metadata.

## Actual Behavior

A dialog/window opens showing thousands of characters of minified JavaScript source code from the bundled `index.js`, specifically OpenTelemetry instrumentation modules.

<details>
<summary>Error content (truncated)</summary>

de=1,Object.defineProperty(ii,"__esModule",{value:!0}),ii.METRIC_SIGNALR_SERVER_ACTIVE_CONNECTIONS=ii.METRIC_KESTREL_UPGRADED_CONNECTIONS=ii.METRIC_KESTREL_TLS_HANDSHAKE_DURATION=...
(continues with Redis/MySQL2/ioredis instrumentation code)

</details>

## Environment

- **OS:** Debian GNU/Linux 13 (trixie)
- **Kernel:** 6.12.73+deb13-amd64
- **Architecture:** amd64
- **Package:** claude-desktop 1.3561.0-2.0.0
- **Desktop:** GNOME (X11)

## Related Issues

- anthropics/claude-code#15779 — Minified JS source code leaking into user-facing error messages

Drizzt321:
This is basically **https://github.com/aaddrick/claude-desktop-debian/issues/399** as well, not just CTL-A, but CTL-Q, and other keys, as people are finding.

aaddrick:
I'm digging into it, and I'm having trouble reproducing on Nobara (Fedora) KDE.

Tested both claude-desktop.x86_64 1.3109.0-1.3.32.fc42 and claude-desktop.x86_64 1.3561.0-2.0.0.fc42 

Still poking around

aaddrick:
Hey @Hayao0819, thanks for filing this.

[...]

PR #564, opened 2026-05-03
feat(lifecycle): notify and offer restart on in-place package upgrade

lizthegrey:
## Summary

Watch `app.asar` for in-place dpkg/rpm replacement; on detection, show a click-to-restart notification. Fixes the Quick Entry "raw JS" / About "minified source" / Ctrl+Q "intermittent" symptom cluster that traces to the running v(N) main process loading v(N+1) renderer assets after an upgrade. Linux only; AppImage and Nix immutable-store paths are inert as expected.

## The bug

dpkg/rpm replace `app.asar` via `rename()` while the main process keeps its in-memory JS. Any `BrowserWindow` opened after the swap reads HTML/asset files fresh from disk, where the hashed asset filenames now point at v(N+1) bundles that the in-memory v(N) IPC and preload no longer match. Symptoms observed across recent triage:

- Quick Entry rendering as raw JS text (renderer's bundled JS shown in `<body>` because the v(N)-shaped HTML→asset path landed in a v(N+1) renderer)
- About dialog showing minified source (#481)
- Ctrl+Q intermittently failing (any path crossing the in-memory↔on-disk boundary)

Before this PR users had no signal that the running process was stale. The macOS / Windows clients get this from Squirrel; Linux deb/RPM has no equivalent.

## The fix

Inserted in `scripts/frame-fix-wrapper.js` as a sibling to the autostart shim — same "main-process boot" surface, runs once on first `require('electron')`.

- **Watch the parent dir, not the file.** File-level `fs.watch` loses the inode across rename-replace; dir-level inotify reports the new entry via `IN_MOVED_TO` / `IN_CREATE`. Filename-filtered to `app.asar` so unrelated activity in the resources dir is ignored.
- **5s debounce past the last event** to clear dpkg's `.dpkg-new` → rename dance and similar multi-stage swaps in rpm / Nix.
- **Compare both `ino` and `mtimeMs`** so the rename-replace path *and* the (rare) in-place-rewrite path both register as a real change.
- **Notification deferred behind `whenReady`** if the app hasn't reached ready when the swap fires; click → `app.relaunch(); app.quit()`. Linux libnotify ignores Electron's `actions` array per the API docs, so whole-notification click is the only affordance — that matches what the macOS client does anyway.
- **Best-effort.** Stat failures and missing-asar paths (AppImage's squashfs mount, dev runs, unusual installs) silently skip arming the watcher.

## Test plan

- [ ] Build deb v(N), install, run from terminal
- [ ] `sudo apt install ./claude-desktop_v(N+1).deb` while it's running
- [ ] Notification appears within ~5s
- [ ] Click → app cleanly relaunches into v(N+1) (verify via launcher.log)
- [ ] AppImage build: confirm watcher arms but never fires (squashfs mount stays pinned to running file's contents)
- [ ] No tests added — `frame-fix-wrapper.js` has no existing test harness in this repo (the autostart shim it sits next to is also untested). Open to wiring up a node mock-test if the maintainer wants one before merge.

---
~100% AI / minimal Human
Claude: diagnosed root cause (running-process / on-disk asset version mismatch after upgrade), designed and implemented the watcher, wrote commit and PR body
Human: filed the bug report, ran the launch-context test that ruled out `CLAUDE_TITLEBAR_STYLE` as the culprit, pointed at the macOS restart-prompt UX as the model to follow

### DIFF
--- a/scripts/frame-fix-wrapper.js
+++ b/scripts/frame-fix-wrapper.js
@@ -654,6 +654,96 @@ X-GNOME-Autostart-enabled=true
         console.log('[Autostart] XDG Autostart shim installed');
       }
 
+      // Detect in-place package upgrade and prompt for restart.
+      // dpkg/rpm replace app.asar via rename(); the running main
+      // process keeps its in-memory JS, but any BrowserWindow opened
+      // after the swap reads HTML/assets fresh from disk, where the
+      // hashed asset filenames already point at v(N+1) bundles the
+      // in-memory v(N) IPC and preload no longer match. Symptoms
+      // observed: Quick Entry rendering as raw JS text, About
+      // dialog showing minified source, Ctrl+Q intermittently
+      // failing — anything where the post-swap window load crosses
+      // the version boundary. macOS / Windows handle this via
+      // Squirrel; Linux deb/RPM has no equivalent, so we watch the
+      // file ourselves and surface a notification. AppImage is
+      // unaffected (squashfs mount stays pinned to the running
+      // file's contents); Nix store paths are immutable until GC,
+      // so the running inode also stays valid until the user
+      // explicitly relaunches.
+      if (process.platform === 'linux') {
+        try {
+          const fs = require('fs');
+          const asarPath = path.join(process.resourcesPath, 'app.asar');
+          let baseline = null;
+          try { baseline = fs.statSync(asarPath); } catch { /* not present */ }
+          if (baseline) {
+            let notified = false;
+            let debounceTimer = null;
+            const promptRestart = () => {
+              if (notified) return;
+              let cur;
+              try { cur = fs.statSync(asarPath); } catch { return; }
+              // Compare both inode (rename-replace path) and mtime
+              // (in-place rewrite path) so we catch either upgrade
+              // shape. Touching the file with no real content change
+              // would still fire — tolerable; the restart is a no-op
+              // for the user, and dpkg/rpm don't do that in practice.
+              if (cur.ino === baseline.ino
+                && cur.mtimeMs === baseline.mtimeMs) {
+                return;
+              }
+              notified = true;
+              console.log(
+                '[Frame Fix] app.asar replaced — prompting restart');
+              const show = () => {
+                try {
+                  const n = new result.Notification({
+                    title: 'Claude Desktop has been updated',
+                    body: 'Click to restart and apply the update.',
+                  });
+                  // Linux libnotify ignores Electron's `actions`
+                  // (macOS-only per the Notification docs), so the
+                  // whole-notification click is the only affordance.
+                  n.on('click', () => {
+                    result.app.relaunch();
+                    result.app.quit();
+                  });
+                  n.show();
+                } catch (err) {
+                  console.warn(
+                    '[Frame Fix] Restart notification failed:',
+                    err.message);
+                }
+              };
+              if (result.app.isReady()) show();
+              else result.app.whenReady().then(show);
+            };
+            // Watch the parent dir, not the file: fs.watch on the
+            // file loses the inode across a rename-replace, but the
+            // dir watcher reports the new entry via inotify
+            // IN_MOVED_TO / IN_CREATE. Filter by filename so we
+            // ignore unrelated activity in the resources dir.
+            const watcher = fs.watch(path.dirname(asarPath),
+              (_evt, filename) => {
+                if (filename !== 'app.asar') return;
+                if (debounceTimer) clearTimeout(debounceTimer);
+                // dpkg writes app.asar.dpkg-new then renames; rpm
+                // and Nix have similar multi-stage swaps. 5s of
+                // post-event quiet covers the slowest of these
+                // without being noticeably late.
+                debounceTimer = setTimeout(promptRestart, 5000);
+              });
+            // Don't keep the event loop alive on the watcher alone;
+            // the app's other handles drive the lifetime.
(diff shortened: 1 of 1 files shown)


aaddrick:
I accidentally ran into something related to this earlier today while working on a testing suite to validate functionality between distros. Issue #567 

I'll come back with more details momentarily

aaddrick:
Hey @lizthegrey! I dug into this further after the earlier review and wanted to share the research before we lock in a direction. Trying to figure out the right shape here together rather than render a verdict.

## Context

I walked the upstream `autoUpdater` path in `build-reference/app-extracted/.vite/build/index.js`. One nuance worth being explicit about: Squirrel *stages* an update and fires `update-downloaded` once the swap is ready. dpkg and rpm *apply* the swap in-place under a running process. So the macOS/Windows signal isn't actually reproducible on Linux deb/rpm. By the time anything could fire, the file is already swapped. This isn't recreating Squirrel; it's filling a different gap. Post-hoc detection of a running stale process. The Quick Entry / About / Ctrl+Q symptom cluster matches the v(N)-process / v(N+1)-on-disk skew you described.

The implementation looks solid from static analysis. Parent-dir `fs.watch` is correct (file-level loses the inode across rename-replace), and the filename filter, 5s debounce, dual `ino` + `mtimeMs` check, `watcher.unref()`, and stat-failure early return all do the right thing. AppImage and Nix paths should be correctly inert — squashfs mount stays pinned to the running file's contents, Nix store is read-only.

I also did a quick ecosystem search to see if there's prior art. Nobody in Electron-land has a clean solution. Official `autoUpdater` is unimplemented on Linux. `electron-updater` explicitly punts on deb/rpm. VSCode doesn't auto-update on Linux at all. Discord polls a server for "update available" but doesn't detect the file swap. Slack relies entirely on apt-source integration. [electron/electron#2198](https://github.com/electron/electron/issues/2198) has been open since 2015. So this approach appears genuinely novel, and probably worth a `docs/learnings/` entry post-merge so future contributors don't go hunting for a standard library.

## Options

There's a related issue (#567) covering a separate bug: Electron's `autoUpdater` is structurally enabled on Linux because `ELECTRON_FORCE_IS_PACKAGED=true` satisfies the eligibility gate. It's only saved from firing because Linux autoUpdater is unimplemented upstream. The plan for #567 is a defensive stub in `frame-fix-wrapper.js` — no-op `setFeedURL` / `checkForUpdates`, redirect `quitAndInstall` to `app.relaunch(); app.quit()`, forward listeners to a local EventEmitter.

That stub creates a hook point. There's a spectrum of how aggressively we use it:

**Level 1 — pure stub.** What #567 proposes standalone. Defensive against future Electron bumps that implement Linux autoUpdater. Doesn't pretend to be anything more.

**Level 2 — stub plus synthetic emission.** Same stub, but this PR's inotify watcher additionally emits a synthetic `update-downloaded` after firing libnotify. The upstream state machine (`vc`) flips to `"ready"`, the in-app menu item becomes "Restart and Install", and `desktop_update_downloaded` telemetry fires. Linux installs start matching macOS/Windows on the upstream dashboards. Tiny diff over Level 1. Static analysis says no platform gating in the relevant code path.

**Level 3 — full hijack.** Substitute our own feed URL. We already run a Cloudflare Worker at `pkg.claude-desktop-debian.dev` fronting the apt/dnf metadata, and a `latest.json` endpoint would be a small addition. Implement the polling that Electron's Linux autoUpdater doesn't, so `checkForUpdates()` actually does something. Two-stage flow: `update-available` when the repo has a newer version (lights up the menu before the user runs apt upgrade), `update-downloaded` when inotify fires (after they have). Combined with a launcher-shim boot-time version check, every state of "running stale / running current / upgraded-while-closed / upgraded-while-running" has a first-class signal.

## Caveats

Level 3 has an awkward "click does what?" question for the `update-available` state. Clicking the menu item there shouldn't relaunch — nothing to load yet. Options are `shell.openExternal()` to the system update center, copying the apt command to clipboard with a follow-up notification, or staying passive until the file actually swaps. Probably the last one, but I'm not certain.

Telemetry implication for Levels 2 and 3: the upstream `update-*` handlers all call `It("desktop_update_*", ...)`. Going Level 2 or 3 means Linux installs start showing up on upstream's update dashboards. Arguably good — parity, gives them visibility — but it's a behavior change worth being explicit about. We'd be pumping events upstream currently doesn't see from Linux.

## Coverage note

Inotify covers "upgraded while running." There's a sister gap for "upgraded while closed, opened cold" that nothing upstream handles either. I plan to file separately as a launcher-shim diff against a stored last-seen-version. The launcher already has clean hook points between the cleanup functions and `setup_electron_env`. Doesn't depend on this PR.

## The ask

Which level feels right? A few possible sequences:

- Land this as-is, land #567 next, optionally amend with Level-2 emission later. Clean separation, two reviewable PRs, neither blocks the other.
- Land #567 first as the stub, then this PR amended for Level 2. Requires waiting for #567 to be written.
- Skip Level 2 entirely, file Level 3 as a "should we?" follow-up issue.

Open to whatever sequencing makes sense, and happy to write whichever piece you don't want to.

---
Written by Claude opus-4.7 via [Claude Code](https://claude.ai/code)

aaddrick:
I'd say, land this PR to shore up the immediate issue, then check in with the team about a path forward. Would be interested in your perspective as well.

aaddrick:
Going to merge for now and revisit the path forward another time. Thanks again @lizthegrey!

PR #566, opened 2026-05-03
fix(ci): force primary GPG key for repomd.xml signing

### LINKED ISSUE #213: RPM repository signature verification fails: repomd.xml signed with subkey instead of main key
## Description

The Fedora/RHEL RPM repository fails signature verification on Fedora 43 with dnf5. The repository metadata (`repomd.xml`) is signed with a GPG subkey, but RPM/DNF does not recognize subkey signatures when verifying repository metadata.

## Error Message
```
>>> repomd.xml GPG signature verification error: Signing key not found
```

## Environment

- **OS:** Fedora 43 (Forty Three)
- **Package manager:** dnf5 version 5.2.17.0
- **Architecture:** x86_64

## Steps to Reproduce

1. Add the repository:
```bash
   sudo curl -fsSL https://aaddrick.github.io/claude-desktop-debian/rpm/claude-desktop.repo -o /etc/yum.repos.d/claude-desktop.repo
```

2. Import the GPG key:
```bash
   curl -fsSL https://aaddrick.github.io/claude-desktop-debian/KEY.gpg | sudo rpm --import -
```

3. Attempt installation:
```bash
   sudo dnf install claude-desktop
```

## Root Cause Analysis

The GPG key is correctly imported into the RPM database:
```
$ rpm -qa 'gpg-pubkey*' | grep b912e0f1
gpg-pubkey-87494cf73acc0f23aa9557b86e29e413b912e0f1-6974cd1a
```

However, the `repomd.xml.asc` signature uses a **subkey** instead of the main key:
```
$ curl -s https://aaddrick.github.io/claude-desktop-debian/rpm/x86_64/repodata/repomd.xml.asc | gpg --list-packets 2>/dev/null | grep keyid
:signature packet: algo 1, keyid 78CE91D72AA9E9B9
```

- **Main key ID:** `B912E0F1` (fingerprint: `87494CF73ACC0F23AA9557B86E29E413B912E0F1`)
- **Signature key ID:** `78CE91D72AA9E9B9` (subkey)

RPM's key handling does not automatically associate subkeys with their parent key for repository signature verification, causing the "Signing key not found" error.

## Suggested Fix

Sign the repository metadata (`repomd.xml`) with the **main key** instead of the subkey. When using `gpg --detach-sign`, add the `--default-key` flag to specify the main key:
```bash
gpg --detach-sign --armor --default-key B912E0F1 repomd.xml
```

Alternatively, if using a signing script, ensure the `--local-user` or `-u` option specifies the main key fingerprint rather than allowing GPG to auto-select a subkey.

## Workaround

Users can currently work around this by disabling GPG check:
```bash
sudo dnf install claude-desktop --nogpgcheck
```

Or by setting `gpgcheck=0` and `repo_gpgcheck=0` in the repo file, though this is not recommended for security reasons.

## References

- [Red Hat: Updating GPG keys for Fedora and RHEL](https://www.redhat.com/en/blog/updating-gpg-keys-for-fedora-and-rhel) - discusses RPM's GPG key handling
- RPM does not support subkey-based repository signatures in the same way as package signatures

ProfFlow:
## Summary

- Append `!` to the keyid passed to `gpg --default-key` when signing
  `repomd.xml`, forcing gpg to use the primary key instead of silently
  selecting the most recent signing subkey.
- Re-fixes #213 (regression of #217). Without `!`, `--default-key`
  only chooses an identity; gpg still picks a subkey for the actual
  signature. rpm 4.20+ / zypper verify `repomd.xml` only against the
  primary key, so the subkey-signed `repomd.xml.asc` currently fails
  verification on Fedora 43 (dnf5) and openSUSE Tumbleweed (zypper
  1.14.96 / rpm 4.20.1) with `Signature verification failed for
  repomd.xml`.

Fixes #213

## Reproduction (current `pkg.claude-desktop-debian.dev`)

```
$ sudo zypper addrepo \
    https://pkg.claude-desktop-debian.dev/rpm/claude-desktop.repo
$ sudo zypper --gpg-auto-import-keys refresh claude-desktop
…
Repository 'Claude Desktop for Fedora/RHEL' ist ungültig.
 - Signature verification failed for repomd.xml
```

Inspecting the published signature confirms the subkey is signing:

```
$ curl -s .../repodata/repomd.xml.asc \
    | gpg --list-packets 2>/dev/null | grep keyid
:signature packet: algo 1, keyid 78CE91D72AA9E9B9   # subkey
# Primary key id: 6E29E413B912E0F1
```

## Why `!` is required

`gpg --default-key FPR` selects the **identity** to sign as; gpg then
picks any valid signing-capable key under that identity, preferring
the most recent subkey. The trailing `!` ("exact key" specifier) forces
gpg to use that exact key for the signature operation. See `gpg(1)`
"How to specify a user ID" / "exact key" notation.

## Test plan

Reproduced and validated end-to-end locally against zypper 1.14.96 /
rpm 4.20.1 / gpg 2.x with a test primary+subkey keypair, re-signing
the live `repomd.xml` from `pkg.claude-desktop-debian.dev`:

| Variant | sig keyid | `zypper refresh` |
|---|---|---|
| `--default-key FPR` (current) | subkey | fail: `Signature verification failed for repomd.xml` |
| `--default-key FPR!` (this PR) | primary | OK: `Die angegebenen Repositorys wurden aktualisiert` |

After merge:

- [ ] Next release tag triggers CI; `update-dnf-repo` job signs
      `repomd.xml` with the primary key.
- [ ] `curl -s …/repomd.xml.asc | gpg --list-packets | grep keyid`
      shows the primary key id (`6E29E413B912E0F1`).
- [ ] `sudo zypper addrepo …/claude-desktop.repo && sudo zypper
      refresh claude-desktop` succeeds on openSUSE Tumbleweed.
- [ ] `sudo dnf install claude-desktop` succeeds on Fedora 43 / dnf5.

## Note on RPM package signing

`rpmsign --addsign` (line 617, via `%_gpg_name` and `__gpg_sign_cmd
-u`) is intentionally left alone. RPM package signature verification
validates subkey-bound signatures via the primary key binding signature
imported into the rpmdb, so subkey-signed `.rpm` files verify
correctly. Only repository metadata verification in rpm 4.20+ is
strict-primary-only — that is the path this PR addresses.

---
85% AI / 15% Human
Claude: identified missing `!` on `--default-key`, set up test
primary+subkey keypair, reproduced bug 1:1 against live repomd.xml,
verified fix end-to-end with zypper 1.14.96, wrote diff, lint, commit,
PR draft.
Human: directed scope (scalpel, no adjacent refactors), rejected
out-of-scope openSUSE-detection path, ran zypper verification on
openSUSE Tumbleweed, gave explicit go/no-go at each gate.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -653,14 +653,12 @@ jobs:
               echo "Generating repodata for $arch..."
               createrepo_c --update "rpm/$arch/"
 
-              # Sign the repository metadata.
-              #   --yes: overwrite existing signature.
-              #   trailing '!' on keyid: force gpg to sign with the
-              #   primary key, not the most recent signing subkey.
-              #   rpm 4.20+ / zypper verify repomd.xml only against
-              #   the primary key; without '!', gpg silently picks
-              #   the subkey and verification fails (regression of
-              #   #213 — PR #217 added --default-key but missed '!').
+              # Sign repodata. Trailing '!' on keyid forces gpg to use
+              # the primary key; without it gpg picks the most recent
+              # signing subkey, and rpm 4.20+ / zypper reject repomd.xml
+              # signed by anything other than the primary key.
+              # Regression of #213 — PR #217 added --default-key but
+              # dropped the '!'. Do not strip it. --yes overwrites .asc.
               echo "Signing repodata for $arch..."
               gpg --batch --yes --default-key "${{ steps.import_gpg.outputs.keyid }}!" --detach-sign --armor "rpm/$arch/repodata/repomd.xml"
             fi
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -653,9 +653,16 @@ jobs:
               echo "Generating repodata for $arch..."
               createrepo_c --update "rpm/$arch/"
 
-              # Sign the repository metadata (--yes to overwrite existing signature)
+              # Sign the repository metadata.
+              #   --yes: overwrite existing signature.
+              #   trailing '!' on keyid: force gpg to sign with the
+              #   primary key, not the most recent signing subkey.
+              #   rpm 4.20+ / zypper verify repomd.xml only against
+              #   the primary key; without '!', gpg silently picks
+              #   the subkey and verification fails (regression of
+              #   #213 — PR #217 added --default-key but missed '!').
               echo "Signing repodata for $arch..."
-              gpg --batch --yes --default-key "${{ steps.import_gpg.outputs.keyid }}" --detach-sign --armor "rpm/$arch/repodata/repomd.xml"
+              gpg --batch --yes --default-key "${{ steps.import_gpg.outputs.keyid }}!" --detach-sign --armor "rpm/$arch/repodata/repomd.xml"
             fi
           done
 


aaddrick (approved):
@ProfFlow I reproduced the gpg-level claim locally. I generated a primary+subkey RSA-3072 test pair matching the production layout (primary `6E29E413B912E0F1`, subkey `78CE91D72AA9E9B9`) and signed a fresh `repomd.xml` three ways. Without `--default-key` and with `--default-key FPR` both ended up signing with the subkey. Only `--default-key FPR!` actually pinned to the primary. The trap is that gpg even prints `using "FPR" as default secret key for signing` before silently signing with the subkey anyway, which is exactly what PR #217 walked into. I also confirmed via curl that the live `pkg.claude-desktop-debian.dev` `repomd.xml.asc` is currently signed by subkey `78CE91D72AA9E9B9`, matching your diagnosis.

The `!` syntax is upstream-blessed. The GnuPG manual's "Specify a User ID" section documents it directly: an exclamation mark may be appended to force using the specified primary or secondary key and not try to calculate which one to use. Same idiom GitHub recommends for git commit signing when you want to pin exactly one key. Right tool for the job.

One honest piece of nuance on the user-visible failure mode. I couldn't fully reproduce the hard rejection on fresh containers. `dnf install claude-desktop` against the live repo on a clean `fedora:43` (dnf5 5.2.18) succeeded end-to-end, and `zypper --gpg-auto-import-keys refresh` against the live repo on `opensuse/tumbleweed:latest` (zypper 1.14.96 / rpm 4.20.1) also succeeded. Could be the dnf5 5.2.17 to 5.2.18 bump softened the path, could be locale or state differences from your host I'm not replicating. Doesn't change the call though. The gpg-level fix is correct, and primary-key signing is strictly more compatible than subkey signing for repodata across the whole rpm/zypper/dnf zoo.

Approving. The `!` is documented and harmless, and your rpmsign-untouched note is also correct since RPM package verification follows subkey binding fine and only repodata is the strict-primary path. Merging.

Issue #559, opened 2026-05-02
docs/audit: regex patch methodology — third `$`-identifier recurrence + safety conventions


aaddrick:
PR #555 (@sirfaber) is the third instance of the same class of bug. Time to write it down.

## Recurrence history

- **PR #253** (Feb 2026, @aaddrick) — `$e` electron module variable rename in 1.1.4088 broke `extract_electron_variable()`. First fix.
- **PR #421** / issue #418 (Apr 2026, @Joost-Maker) — `$e` fs reference in the startVM block on 1.3109.0 broke Patch 9 of `patch_cowork_linux()`. Introduced the `[$\w]+` pattern at `scripts/patches/cowork.sh:482-501`.
- **PR #555** (May 2026, @sirfaber) — `C$i` in the vm assignment (Patch 2b) and `g$i` in retry-delay (Patch 6 step 2) on 1.5354.0 broke two more anchors that still used `\w+`.

Three recurrences of the same root cause (`\w` excludes `$`; minified JS identifiers can contain `$`) is enough signal to systematize. It's not enough to preemptively rewrite everything.

## Deliverables

- [ ] **Add `docs/learnings/minified-identifiers.md`.** Cover why `\w` doesn't match `$`, when minifiers emit `$`-containing identifiers, the canonical `[$\w]+` (or `[\w$]+`) widening, and how to verify against real minified bytes. Include the three prior recurrences as case studies. Cross-reference the "Resilience to minifier churn" section of `tray-rebuild-race.md` and the `reference-source.tar.gz` pointer in `plugin-install.md`. Specifically:
    - **The beautified false-negative trap.** Testing a pattern against `build-reference/` is not verification. During PR #555 I tested `\w+` and `[\w$]+` against the beautified copy and both returned `false`, because the beautified file has whitespace the regex didn't account for (`await new Promise((n) => setTimeout(n, g$i))` vs minified `await new Promise(n=>setTimeout(n,g$i))`). The lesson isn't "use minified instead of beautified." It's that beautified output will silently confirm a broken hypothesis if you stop there. Always close the loop against shipped bytes.
[...]

PR #572, opened 2026-05-03
doctor: detect IBus/GTK misconfigurations that break input

aaddrick:
**Stacked on #571.** Targets `feature/549-gtk-im-module-override` because the `CLAUDE_GTK_IM_MODULE` visibility check refers to the env var introduced there. Once #571 merges I'll rebase onto main and retarget.

## Summary

- New `_doctor_check_im_modules` helper covering the four IBus/GTK failure modes from #545:
  - `ibus-gtk3` package missing while `GTK_IM_MODULE=ibus`
  - GTK immodules cache stale (`gtk-query-immodules-3.0` doesn't list the active module)
  - XWayland session routing IBus through XIM (informational, points at `CLAUDE_USE_WAYLAND=1`)
  - `CLAUDE_GTK_IM_MODULE` override visibility (informational)
- Each check is gated — package check only fires for `ibus`; cache check skipped when `gtk-query-immodules-3.0` is absent; package check returns silently on distros without dpkg/rpm/pacman.
- Hoists `_distro_id` resolution to the top of `run_doctor` so the IM check and the existing Cowork section share one `/etc/os-release` read.
- New `tests/doctor.bats` with 17 cases covering each gating branch and the `_cowork_pkg_hint` mapping for `ibus-gtk3` (Arch → `ibus`, others → `ibus-gtk3`).
- TROUBLESHOOTING entry extended with a `--doctor` callout and the diagnostics table updated to include the new check.

Closes #550. 

Refs #545, #549.

## Test plan

- [ ] `bats tests/doctor.bats` — 17 cases pass
- [ ] `bats tests/launcher-common.bats` — 68 cases still pass (no regression from `_distro_id` hoist)
- [ ] Manual on a deb/AppImage install: `claude-desktop --doctor` shows the IM section, fires expected warnings on a host with stale immodules cache or missing `ibus-gtk3`
- [ ] CI green on amd64 + arm64

---
Generated with [Claude Code](https://claude.ai/code)
Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>
70% AI / 30% Human
Claude: drafted IM check helper + gating logic, _pkg_installed cross-distro probe, BATS coverage with function-stub mocking
Human: scoped issue, decided three-PR sequence with stacking, reviewed

PR #575, opened 2026-05-04
verify(cowork): static-grep shipped asar for PR #555 markers (#559)

aaddrick:
D6 of the #559 followup. PR #555 was the third recurrence of `\w+` silently truncating a `$`-containing minified identifier, and two of the three failed gates had no `else` branch — the build log printed `Applied 10 cowork patches` while the asar shipped half-patched. A static post-build grep on the shipped bytes catches that drift no matter which gate slipped.

The check runs against the actual installed JS, not against `build-reference/`. That's the part that matters. The trap I documented in `patching-minified-js.md` is regexes that look correct against the beautified reference but fail against the minified source, so verifying the post-build asar is the only way to close the loop.

## Design

`scripts/cowork-patch-markers.tsv` is the single source of truth: tab-separated `name<TAB>pcre<TAB>sample`, 9 rows, 1:1 with the smoke-test set named in #559's body. Both `scripts/verify-cowork-patches.sh` and `tests/verify-cowork-patches.bats` parse it via a shared `load_markers` helper (the BATS file sources the script), so they can't drift apart by construction.

The verify script accepts a `.js` file, an `.asar` (extracted via `npx @electron/asar`), or a directory containing `app.asar.contents/.vite/build/index.js`. Exit 0 on full match, 1 on usage/path errors, 2 on missing markers (named on stderr). The CI step in `.github/workflows/build-amd64.yml` is gated on `inputs.artifact_suffix == 'deb'` because the patched JS is byte-identical across formats — one verification per CI run is enough.

PR 3 (silent-no-op WARNING retrofit) follows. Its empirical argument is "deliberately break each gate, confirm the WARNING surfaces it," and that experiment depends on marker-grep already running in CI.

## Fixes

Refs #559. PR 1 landed as 4bbd117.

## Test plan

- [x] `bats tests/*.bats` — 227/227 pass (9 new, 218 existing)
- [x] `shellcheck scripts/verify-cowork-patches.sh` — clean
- [x] `actionlint .github/workflows/build-amd64.yml` — only the pre-existing `$TAG_FLAG` warning in an untouched step
- [x] End-to-end against shipped bytes: pulled the pinned 1.5354.0 installer, verified SHA-256 against `claude_exe_sha256`, extracted `app.asar`, ran verify on unpatched source → 9/9 miss as expected. Ran `scripts/patches/cowork.sh` against the same bytes and re-verified → all 9 markers present, exit 0.

---
85% AI / 15% Human
Claude: drafted the #559 comment defining the 3-PR split and 9-marker set; chose TSV-as-single-source-of-truth; wrote the verify script, BATS matrix, CI step, and PR description; ran the end-to-end fixture check against the pinned 1.5354.0 installer.
Human: approved the followup plan, scoped this PR to D6, reviewed the diff.

### DIFF
--- a/.github/workflows/build-amd64.yml
+++ b/.github/workflows/build-amd64.yml
@@ -49,6 +49,29 @@ jobs:
           fi
           ./build.sh ${{ inputs.build_flags }} $TAG_FLAG
 
+      # Static-grep the shipped asar for the cowork patch markers
+      # defined in scripts/cowork-patch-markers.tsv (issue #559 D6,
+      # PR #555). Pinned to amd64-deb because the patched JS is
+      # identical across formats, so one verification per CI run is
+      # sufficient — no need to duplicate across the matrix.
+      - name: Verify cowork patches in shipped asar
+        if: inputs.artifact_suffix == 'deb'
+        run: |
+          deb_file=$(find . -maxdepth 1 -name 'claude-desktop_*amd64.deb' \
+            -print -quit)
+          if [[ -z "$deb_file" ]]; then
+            echo "verify-cowork-patches: no .deb artifact found" >&2
+            exit 1
+          fi
+          extract_dir=$(mktemp -d)
+          dpkg-deb -x "$deb_file" "$extract_dir"
+          asar_path=$(find "$extract_dir" -name app.asar -print -quit)
+          if [[ -z "$asar_path" ]]; then
+            echo "verify-cowork-patches: app.asar not found in deb" >&2
+            exit 1
+          fi
+          ./scripts/verify-cowork-patches.sh "$asar_path"
+
       - name: Upload AMD64 Artifact
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
         with:
--- a/.github/workflows/build-amd64.yml
+++ b/.github/workflows/build-amd64.yml
@@ -60,17 +60,17 @@ jobs:
           deb_file=$(find . -maxdepth 1 -name 'claude-desktop_*amd64.deb' \
             -print -quit)
           if [[ -z "$deb_file" ]]; then
-            echo "verify-cowork-patches: no .deb artifact found" >&2
+            echo "verify-patches: no .deb artifact found" >&2
             exit 1
           fi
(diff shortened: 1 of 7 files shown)


sabiut (approved):
Tests-side review (CODEOWNER for `/tests/`): tested locally end-to-end, approving.

**Local verification:**
- `bats tests/verify-cowork-patches.bats` → 9/9 pass
- `shellcheck scripts/verify-cowork-patches.sh` → clean
- `actionlint .github/workflows/build-amd64.yml` → only the pre-existing `$TAG_FLAG` SC2086 in the untouched "Run build script" step, as the PR description states
- End-to-end against real artifacts from a local build (PR #572 build tree, same patches as main):
  - Patched `index.js` (file path) → exit 0, 9 OK
  - Patched `app.asar` (exercises `npx @electron/asar extract`) → exit 0, 9 OK — the asar-extraction codepath that BATS doesn't cover, working in practice
  - Directory-shaped input → exit 0, 9 OK
  - **Upstream pre-patch `app.asar`** (negative against real bytes) → exit 2, 9/9 missing, every name surfaced in stderr — exactly the failure mode this PR exists to catch

**What I particularly like:**
- The `markers file: every regex matches its sample` test is the right reflex — that's the exact guard that would have caught the PR #555 regex truncation before merge.
- TSV-as-single-source-of-truth + the `_is_sourced` library-mode source pattern means the verify script and the BATS suite literally cannot drift — a marker-list edit hits both consumers automatically.
- Per-marker negative fixtures driven by a loop over `marker_names` — adding a marker grows the test matrix for free.

**One side observation, not blocking:** this file landed on me via `/tests/ → @sabiut` because its name is `verify-cowork-patches.bats`, not `cowork-*.bats`. The cowork override on line 94 of CODEOWNERS only matches `cowork-*` filenames, so `verify-cowork-patches.bats` slips past it. If you intend this verify infrastructure to be generic / cross-subsystem (the script+TSV are reusable for non-cowork patch sets in principle), the current name is right. If you'd rather it route alongside the other Cowork tests so @RayCharlizard sees changes, a `git mv` to `cowork-verify-patches.bats` would do it. Either is fine — flagging so the routing is intentional rather than incidental.

Issue #583, opened 2026-05-06
[bug]: GPU process FATAL on Linux: fallback exhaustion → browser SIGTRAP


aaddrick:
## Version

```
Claude Desktop Diagnostics
================================

[PASS] Installed version: 1.5354.0
[PASS] Display server: Wayland (WAYLAND_DISPLAY=wayland-0)
       Desktop: KDE
       Mode: X11 via XWayland (default, for global hotkey support)
       Tip: Set CLAUDE_USE_WAYLAND=1 for native Wayland
            (disables global hotkeys)
       Menu bar mode: auto (default, Alt toggles visibility)
       Titlebar style: hybrid (default, native frame + in-app topbar)
[PASS] Electron: v41.5.0 (/usr/lib/claude-desktop/node_modules/electron/dist/electron)
[PASS] Chrome sandbox: permissions OK (/usr/lib/claude-desktop/node_modules/electron/dist/chrome-sandbox)
[PASS] SingletonLock: no lock file (OK)
[PASS] MCP config: valid JSON (/home/aaddrick/.config/Claude/claude_desktop_config.json)
       MCP servers configured: 0
[PASS] Node.js: v22.22.0
       Path: /usr/bin/node
[PASS] Desktop entry: /usr/share/applications/claude-desktop.desktop
[PASS] Disk space: 508264MB free

Cowork Mode
----------------
[PASS] bubblewrap: found
[PASS] bubblewrap: sandbox probe succeeded
[PASS] KVM: accessible
[PASS] vsock: module loaded
[PASS] QEMU: found
[PASS] socat: found
[PASS] virtiofsd: found at /usr/libexec/virtiofsd (not on PATH)
       VM image: not downloaded yet
       Cowork isolation: bubblewrap (namespace sandbox)
[WARN] Cowork daemon: orphaned (PIDs: 4570)
       Fix: Restart Claude Desktop (daemon will be cleaned up automatically)
[PASS] Log file: 145KB (/home/aaddrick/.cache/claude-desktop-debian/launcher.log)

All checks passed.
```

Host: Fedora/Nobara, KDE Plasma 6 Wayland session, Electron running through XWayland.

## What happened

This is the tracking issue I promised in #582 for the underlying Electron crashes that surfaced as a 190GB syslog on Ubuntu 26.04.

[...]

Issue #582, opened 2026-05-06
[bug]: Electron stdout floods systemd-journal / syslog with debug logs, causing /var/log/syslog to grow unboundedly


cbnuke:
### Version (`claude-desktop --doctor` output)

```shell
Claude Desktop Diagnostics
================================

[PASS] Installed version: 1.6259.0-2.0.10
[PASS] Display server: Wayland (WAYLAND_DISPLAY=wayland-0)
       Desktop: ubuntu:GNOME
       Mode: X11 via XWayland (default, for global hotkey support)
       Tip: Set CLAUDE_USE_WAYLAND=1 for native Wayland
            (disables global hotkeys)
       IME note: Wayland session, Electron via XWayland — IBus path goes through XIM (lossy for some IMEs).
       Tip: CLAUDE_USE_WAYLAND=1 enables native Wayland IME (loses global hotkeys).
       Menu bar mode: auto (default, Alt toggles visibility)
       Titlebar style: hybrid (default, native frame + in-app topbar)
[PASS] Electron: v41.5.0 (/usr/lib/claude-desktop/node_modules/electron/dist/electron)
[PASS] Chrome sandbox: permissions OK (/usr/lib/claude-desktop/node_modules/electron/dist/chrome-sandbox)
[PASS] SingletonLock: held by running process (PID 17310)
[PASS] MCP config: valid JSON (/home/cbnuke/.config/Claude/claude_desktop_config.json)
       MCP servers configured: 3
[PASS] Node.js: v25.9.0
       Path: /home/cbnuke/.nvm/versions/node/v25.9.0/bin/node
[PASS] Desktop entry: /usr/share/applications/claude-desktop.desktop
[PASS] Disk space: 186819MB free

Cowork Mode
----------------
[PASS] bubblewrap: found
[PASS] bubblewrap: sandbox probe succeeded
[PASS] KVM: accessible
[PASS] vsock: module loaded
       QEMU: not found
[PASS] socat: found
       virtiofsd: not found
       VM image: not downloaded yet
       Cowork isolation: bubblewrap (namespace sandbox)
[PASS] Cowork daemon: running (parent alive)
[PASS] Log file: 13KB (/home/cbnuke/.cache/claude-desktop-debian/launcher.log)

All checks passed.
```

### What happened

Title: Electron stdout floods systemd-journal / syslog with debug logs, causing /var/log/syslog to grow unboundedly

**Environment**
- OS: Ubuntu 26.04
[...]

Issue #387, opened 2026-04-10
GPU process FATAL crash on local X11: "GPU process isn't usable. Goodbye." — breaks clipboard until restart


GuyMannDude:
## Description

Claude Desktop's GPU process is crashing with a FATAL error on a local X11 display (no XRDP, no VNC, no remote session). The crash happens intermittently — 5 occurrences across 7 days — and after each crash, clipboard operations (copy/paste) become unreliable until Claude Desktop is fully restarted.

This is distinct from #319 (XRDP/VNC-specific GPU crash) and #321 (GNOME quit/orphaned processes). This fires during normal use on a local X11 session with an NVIDIA laptop GPU.

## Symptoms

- `[FATAL] GPU process isn't usable. Goodbye.` in the launcher log
- After the crash, copy/paste from Claude Desktop to other applications stops working or becomes intermittent
- Restarting Claude Desktop restores clipboard functionality
- The app does not always visibly crash — sometimes it limps along with broken clipboard until manually restarted
- 5 occurrences in 7 days (April 2, 5, 5, 7, 7) plus a separate `Failed to shutdown` FATAL on April 9

## Crash log (from ~/.cache/claude-desktop-debian/launcher.log)

All 5 GPU FATAL entries follow the same pattern — multiple `error_code=1002` launch failures immediately preceding the FATAL:

```
[1225603:0407/121726.265199:ERROR:content/browser/gpu/gpu_process_host.cc:990] GPU process launch failed: error_code=1002
[1225603:0407/121726.266198:ERROR:content/browser/gpu/gpu_process_host.cc:990] GPU process launch failed: error_code=1002
[1225603:0407/121726.266608:ERROR:content/browser/gpu/gpu_process_host.cc:990] GPU process launch failed: error_code=1002
[1225603:0407/121726.266879:ERROR:content/browser/gpu/gpu_process_host.cc:990] GPU process launch failed: error_code=1002
[1225603:0407/121726.267122:ERROR:content/browser/gpu/gpu_process_host.cc:990] GPU process launch failed: error_code=1002
[1225603:0407/121726.267130:FATAL:content/browser/gpu/gpu_data_manager_impl_private.cc:417] GPU process isn't usable. Goodbye.
```

All 5 occurrences are identical in structure. Each crash is followed by a launcher restart (`--- Claude Desktop Launcher Start ---`).

There is also a separate FATAL on April 9 during shutdown:
```
[...]

PR #585, opened 2026-05-06
launcher+doctor: GPU FATAL mitigations (mitigates #583)

aaddrick:
## Summary

Two small, parallel mitigations for the Chromium GPU process FATAL exhaustion tracked in #583. Neither fixes the underlying Electron/Chromium issue — that needs symbolicated stacks and upstream — but both reduce the user-visible impact today.

- **`launcher`**: `CLAUDE_DISABLE_GPU=1` opt-in env var that adds `--disable-gpu` `--disable-software-rasterizer` to Electron's argv (`scripts/launcher-common.sh`). Mirrors the existing XRDP block; co-occurrence is deduplicated via a single `_disable_gpu` sentinel so flags aren't pushed twice. Logged in `log_session_env` so bug reports include the value.
- **`doctor`**: `_doctor_check_recent_crashes` surfaces a `[WARN]` when systemd-coredump shows ≥3 Electron crashes in the last 7 days, pointing at the upstream Settings toggle and `CLAUDE_DISABLE_GPU=1` as workarounds, plus a link to #583. Falls back to all-electron entries with a footnote when the caller's electron path doesn't match (AppImage transient paths, other Electron apps installed side-by-side). Silent on hosts without `coredumpctl`, on zero matches, or below threshold.

The motivation is empirical, not speculative: I caught a fresh repro of the FATAL today on AMD-only hardware, which (combined with @GuyMannDude's Optimus report from #387 and @cbnuke's Ubuntu 26.04 apport file from #582) gives three independent hardware paths to the same `gpu_data_manager_impl_private.cc:417` FATAL. Full evidence is on #583.

#583 stays open. These are workarounds, not fixes.

## Test plan

- [x] `bats tests/launcher-disable-gpu.bats tests/launcher-xrdp-detection.bats tests/doctor.bats` — 37/37 pass locally
- [x] `shellcheck` clean (no new warnings; only pre-existing patterns inherited from the existing test files)
- [x] Live-tested `_doctor_check_recent_crashes` against my actual coredump history — fires correctly, reports 10 crashes in last 7 days with the documented warning shape
- [x] Verified XRDP+CLAUDE_DISABLE_GPU co-occurrence test shows exactly one of each flag (no duplication)
- [ ] Doctor section visible end-to-end via `claude-desktop --doctor` post-merge on a host with recent coredumps (smoke test post-release)

## What this doesn't do

- Doesn't change behaviour for users not setting `CLAUDE_DISABLE_GPU`
- Doesn't suppress legitimate crashes (the doctor warn fires on count, but it's an info/warn — never an error)
- Doesn't address the underlying GPU process lifecycle bug — #583 stays open as the canonical tracking issue

---
90% AI / 10% Human
Claude: investigation, code, tests, docs
Human: strategic direction (consolidating into #583 not #387, keeping #583 open, scope of mitigations vs fix), course correction, final approval

### DIFF
--- a/scripts/doctor.sh
+++ b/scripts/doctor.sh
@@ -436,6 +436,66 @@ JSEOF
 	fi
 }
 
+# Surface a warning when systemd-coredump shows N+ recent Electron
+# crashes. The most common cause on Linux is the GPU process FATAL
+# exhaustion tracked in #583 — workaround for affected users is the
+# upstream Settings → disable hardware acceleration toggle, or
+# CLAUDE_DISABLE_GPU=1 in the environment for headless persistence.
+#
+# Arguments: $1 = electron path (e.g.,
+#   /usr/lib/claude-desktop/node_modules/electron/dist/electron)
+#   Used to filter results to claude-desktop's electron when possible;
+#   falls back to all-electron crashes when the path doesn't match
+#   (e.g., AppImage mount paths are transient).
+_doctor_check_recent_crashes() {
+	local electron_path="${1:-}"
+	command -v coredumpctl &>/dev/null || return 0
+
+	# `coredumpctl list electron` filters by COMM=electron. If the
+	# exact electron_path matches any entry's EXE column, prefer that
+	# tighter count; otherwise fall back to all-electron entries.
+	local listing total_count path_count
+	listing=$(coredumpctl list electron \
+		--since='7 days ago' --no-pager 2>/dev/null) || return 0
+	[[ -n $listing ]] || return 0
+
+	# Drop the header line; count remaining entries.
+	total_count=$(awk 'NR>1 && NF>0' <<< "$listing" | wc -l)
+	((total_count == 0)) && return 0
+
+	if [[ -n $electron_path ]]; then
+		path_count=$(awk -v p="$electron_path" \
+			'NR>1 && index($0, p)' <<< "$listing" | wc -l)
+	else
+		path_count=0
+	fi
+
+	# Use the path-matched count when available; else the unfiltered
+	# count with a footnote so the user knows it may include other
+	# Electron apps (Slack, VSCode, etc.).
+	local count footnote=''
+	if ((path_count > 0)); then
+		count=$path_count
+	else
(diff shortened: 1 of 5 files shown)


sabiut (approved):
## Review

Approve with comments. Verified locally: 107/107 BATS pass, the 6 new `launcher-disable-gpu` cases pass, shellcheck on `doctor.sh` and `launcher-common.sh` shows no new warnings, and live `_doctor_check_recent_crashes` exercises all four branches (≥3 → WARN, <3 → INFO, path-mismatch → footnote, no `coredumpctl` → silent rc=0).

The mitigation is well-scoped and the test shape is right — positive path, dedup, off-states pinned with strict equality, and the fallback footnote. Nicely paired with the `log_session_env` extension so the new env var shows up in session logs.

A few non-blocking observations:

### 1. `CLAUDE_DISABLE_GPU=1` isn't in `TROUBLESHOOTING.md`

The new doctor warning tells users to set `CLAUDE_DISABLE_GPU=1`, but `docs/TROUBLESHOOTING.md` doesn't document it yet. Worth either adding here or as a follow-up — that file is also where the broader #583 narrative belongs (XRDP context, when to reach for the flag, what it actually disables).

### 2. Threshold of 3 is hard-coded

A user who keeps hitting #583 until the upstream cause is fixed will see the warning every `--doctor` run. That's probably fine — it's the desired signal — but worth either:

- a brief comment explaining why 3 (e.g., "smoke-test threshold; tuned against aaddrick's 10 crashes / 7 days reproduction"), or
- a follow-up `CLAUDE_DOCTOR_CRASH_THRESHOLD` override for users who've already acted on the warning.

Not a blocker; just flagging so the magic number isn't orphaned.

### 3. `coredumpctl list` `-- Reboot --` separators

`coredumpctl list electron` should pre-filter by `COMM`, so reboot-marker rows shouldn't appear in practice. But the total-count branch (`NR>1 && NF>0`) would include them if they ever did slip through on a newer systemd. The path-matched branch is unaffected. A short comment near the awk would future-proof it.

Issue #584, opened 2026-05-06
[bug]: build.sh fails on electron@42.0.0 — "Failed to find Electron distribution directory"


BlueManCZ:
## What happened

`build.sh` fails in `setup_electron_asar` (scripts/setup/dependencies.sh) immediately after `npm install`:

```
Installing Electron and Asar locally into <work_dir>...

added 32 packages, and audited 33 packages in 2s

found 0 vulnerabilities
Electron and Asar installation command finished.
Failed to find Electron distribution directory at '<work_dir>/node_modules/electron/dist' after installation attempt.
```

`npm install` itself reports success — no errors, no warnings — but `node_modules/electron/dist/` is never created, so the subsequent existence check fails and the build aborts.

## Root cause

`scripts/setup/dependencies.sh:218` runs:

```bash
npm install --no-save electron @electron/asar
```

with no version constraint, so it resolves `electron` to whatever is currently tagged `latest`. **electron@42.0.0 was published 2026-05-06T01:28:08Z and removed the `postinstall` script** that previously downloaded the prebuilt binary into `node_modules/electron/dist/`:

```
$ npm view electron@41.5.0 scripts
{ postinstall: 'node install.js' }

$ npm view electron@42.0.0 scripts
   # (empty — no scripts field at all)
```

With electron@42, `npm install` finishes without ever populating `dist/`, and `setup_electron_asar`'s existence check at line 233 fails.

This is a fresh regression as of 2026-05-06 — the release tarball v2.0.7+claude1.5354.0 (and prior) build fine if you pin electron, since they ran against electron 41.x. Anyone building from `build.sh` today (packagers, NixOS, Gentoo overlays, AUR, anyone running it manually) is affected; users installing the prebuilt `.deb`/`.rpm`/`.AppImage` from Releases are not, since electron is already bundled into those.

## Steps to reproduce

```bash
git clone https://github.com/aaddrick/claude-desktop-debian
cd claude-desktop-debian
./build.sh --build deb    # or appimage, rpm — same failure
```

Reproduces on x86_64 Linux with Node.js 20+ and any npm version. Also reproduces if you check out tag `v2.0.10+claude1.6259.0` directly.

## Expected behavior

`build.sh` produces a working package, as it did for v2.0.7 and prior. The regression is triggered purely by the new electron release on npm — no change in this repo.

## Suggested fix

[...]

PR #587, opened 2026-05-06
fix(deps): fetch electron binary via @electron/get, drop ^41 pin

aaddrick:
## Summary

Follow-up to #586. The hotfix pinned `electron@^41` to avoid the missing-postinstall regression in `electron@42.0.0`. This PR keeps the version pin (now `41.5.0` exact, matching upstream `build-reference/app-extracted/package.json`) but stops relying on electron's postinstall to populate `dist/`. We fetch the binary ourselves so a future postinstall removal or behavior change doesn't break the build again.

- **New helper** `scripts/setup/fetch-electron-binary.js` — uses `@electron/get` to download the matching prebuilt binary for the host platform/arch, then `extract-zip` to unpack it into `node_modules/electron/dist/`. Reads the version from the just-installed electron's `package.json` so it tracks whatever the pin resolves to.
- **`setup_electron_asar`** runs the helper only when `npm install` leaves `dist/` missing. If electron ever restores the postinstall, this becomes a no-op.
- **Pins `electron@41.5.0` exact** in `scripts/setup/dependencies.sh`, hoisted to a named `electron_version` local. Upstream Claude Desktop's `app.asar` binds to a specific V8/NAPI ABI, Chromium pairing, and `node-pty` native surface. Running a different Electron major against it is unsupported. Bump it when upstream bumps.
- **Retry once on transient CDN failures.** A 503 from `github.com/electron/electron/releases` shouldn't red the whole build.
- **Arch whitelist** in `fetch-electron-binary.js` (`x64`, `arm64`, `armv7l`, `ia32`) so an exotic-arch host gets an actionable error instead of a 404 from electron's release server.

## Why two new npm deps

`@electron/get` returns a path to a downloaded zip; it doesn't extract. `extract-zip` is the same library electron's own (now-removed) `install.js` used, and it's already a transitive dep of `@electron/get`. Installing it explicitly makes the dependency surface obvious rather than relying on hoisting.

The alternative was shelling out to `unzip(1)`, which would add a system-package dep that isn't used anywhere else in this repo.

## Mirror / cache env vars

`@electron/get` reads `ELECTRON_MIRROR` and `ELECTRON_CUSTOM_DIR`; both are documented in `docs/BUILDING.md`. **`ELECTRON_CACHE` is NOT honored** — the cache lives at `~/.cache/electron/` (resolved via `envPaths`). Earlier versions of this PR body claimed `ELECTRON_CACHE` worked. That was wrong.

## Failure modes considered

| Scenario | Behavior |
|---|---|
| electron postinstall returns | Helper never runs (`dist/` already present) — no-op. |
| Transient 503 from electron CDN | Retried once; second failure exits with an `ELECTRON_MIRROR` pointer. |
| Electron release pulled from CDN | `@electron/get` throws after retry; bash propagates a clear error. |
| Exotic host arch | `fetch-electron-binary.js` throws with the supported-arch list. |
| Cross-arch build | Uses `process.arch` (mapped: `arm` → `armv7l`); CI runners build on their own arch, matching today's behavior. |
| Mirror env vars | `@electron/get` honors `ELECTRON_MIRROR` and `ELECTRON_CUSTOM_DIR`. |
| Stale `dist/` from prior build | Pre-existing `install_needed` check skips both npm install and the helper, same as before. |

## Test plan

- [ ] `./build.sh --build deb` on amd64 — fresh `work-dir` (no cached `dist/`)
- [ ] `./build.sh --build deb` on amd64 — second run reuses cached `dist/`
- [ ] `./build.sh --build appimage` on amd64
- [ ] `./build.sh --build rpm` on amd64
- [ ] CI green on this branch (covers arm64 + all three formats)
- [ ] Verify the produced binary launches and reports Electron 41.5.0

Refs #584
Follow-up to #586

---
85% AI / 15% Human
Claude: designed the helper, traced downstream consumers to confirm the `dist/` layout contract, drafted the bash wiring, wrote the script, ran shellcheck and node --check, drafted commit and PR copy, applied self-review feedback (re-pinning `electron@41.5.0`, retry wrap, arch whitelist, BUILDING.md docs, ELECTRON_CACHE correction).
Human: chose the durable-fix-as-separate-PR strategy, signed off on the `@electron/get` + `extract-zip` approach over the `unzip(1)` alternative, caught the ABI-break risk from dropping the version pin during self-review.

### DIFF
--- a/docs/BUILDING.md
+++ b/docs/BUILDING.md
@@ -41,6 +41,17 @@ The build script automatically detects your distribution and selects the appropr
 | Arch Linux | `.AppImage` (via AUR) | yay/paru |
 | Other | `.AppImage` | - |
 
+## Build Environment Variables
+
+The build pulls the Electron prebuilt binary from `github.com/electron/electron/releases` via `@electron/get`. Two upstream environment variables let you redirect that fetch:
+
+- `ELECTRON_MIRROR` — base URL to fetch Electron releases from instead of GitHub. Useful for mirrors or local proxies. Example: `ELECTRON_MIRROR=https://npmmirror.com/mirrors/electron/`.
+- `ELECTRON_CUSTOM_DIR` — overrides the path segment after the mirror. Defaults to `v{version}`.
+
+The cache location is fixed at `~/.cache/electron/` (resolved by `@electron/get` via `envPaths`) and is reused across builds. `ELECTRON_CACHE` is **not** read by `@electron/get` — set `ELECTRON_MIRROR` if you need to avoid the public CDN.
+
+The pinned Electron version lives in `scripts/setup/dependencies.sh` (`electron_version`) and must match `build-reference/app-extracted/package.json` — the upstream Claude Desktop `app.asar` is built against a specific Electron major and running a different one is unsupported.
+
 ## Installing the Built Package
 
 ### For .deb packages (Debian/Ubuntu)
--- a/scripts/setup/dependencies.sh
+++ b/scripts/setup/dependencies.sh
@@ -215,16 +215,25 @@ setup_electron_asar() {
 
 	if [[ $install_needed == true ]]; then
 		echo "Installing Electron and Asar locally into $work_dir..."
-		# Pin to electron 41.x: electron@42.0.0 (2026-05-06) dropped the
-		# postinstall that fetches the prebuilt binary into dist/, leaving
-		# node_modules/electron/dist absent and the build aborting (#584).
-		# A durable fix using @electron/get is tracked separately.
-		if ! npm install --no-save 'electron@^41' @electron/asar; then
+		if ! npm install --no-save \
+			electron @electron/asar @electron/get extract-zip; then
 			echo 'Failed to install Electron and/or Asar locally.' >&2
 			cd "$project_root" || exit 1
 			exit 1
 		fi
 		echo 'Electron and Asar installation command finished.'
+
+		# electron@42+ no longer ships a postinstall script that fetches
+		# the prebuilt binary into dist/. If npm didn't populate it, fetch
+		# the matching binary explicitly via @electron/get. See #584.
+		if [[ ! -d $electron_dist_path ]]; then
+			echo 'Electron postinstall did not populate dist/; fetching binary explicitly...'
+			if ! node "$project_root/scripts/setup/fetch-electron-binary.js"; then
+				echo 'Failed to fetch Electron binary via @electron/get.' >&2
+				cd "$project_root" || exit 1
+				exit 1
+			fi
+		fi
 	else
 		echo 'Local Electron distribution and Asar binary already present.'
 	fi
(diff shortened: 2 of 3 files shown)


aaddrick:
Have not verified locally yet. Local verification and review required

aaddrick:
Verdict: **REQUEST CHANGES**. Direction is right — fetching the binary ourselves instead of leaning on electron's postinstall. I dropped the `^41` pin without thinking through what it was doing, though, so this needs rework before I merge it.

`build-reference/app-extracted/package.json:42` pins `"electron": "41.5.0"` exact. Anthropic's `app.asar` is built against E41 (V8/NAPI ABI, Chromium pairing, `node-pty` native surface). With the pin gone, `npm install electron` resolves `latest` — currently `42.0.1`, with `43.0.0-alpha.2` already in the registry. Two `./build.sh` runs from the same git SHA on different days can ship different Electron majors against an asar built for 41.5.0. That kills #586 and makes the #583 GPU-mitigation work non-reproducible. My earlier hand-wave that "patches anchor on minified literals, an Electron major doesn't invalidate them" only covers the patch suite, not the binary-against-asar ABI. I need to re-add `electron@41.5.0` exact, and ideally hoist it to a named variable in `dependencies.sh` so the next bump is intentional. `@electron/get` still earns its keep as the postinstall-gap fix.

Retracting my `cacheRoot` suggestion from the first pass. It's wrong. `setup_work_directory` does `rm -rf "$work_dir"` at the top of every build (`dependencies.sh:103`), so anchoring the cache there forces a ~100MB CDN re-download on every clean build. That's a regression for CI, AUR `--noconfirm`, and packager chroots, and it converts every CDN hiccup into a build failure. Default `~/.cache/electron/` via `envPaths` is right — standard, shared with any other Electron tool, stable across `--clean yes`. I should document `ELECTRON_MIRROR` / `ELECTRON_CUSTOM_DIR` in the README Build section and wrap the `node fetch-electron-binary.js` call in a 2-attempt retry so a transient 503 from `github.com/electron/electron/releases` doesn't red the whole build.

The PR-body claim that `ELECTRON_CACHE` is honored is wrong. Needs fixing before merge, not after. `@electron/get@5.0.0/Cache.js:8` reads `envPaths('electron').cache` with no env override; only `ELECTRON_MIRROR` and `ELECTRON_CUSTOM_DIR` flow through `artifact-utils.js:32`. A Debian maintainer setting `ELECTRON_CACHE=` in their buildd config silently gets nothing. Squash-amend the body. Findings that still hold from the first pass: Nix path is unaffected (`build.sh:264-266` gates `setup_electron_asar` on non-Nix, derivation pulls from nixpkgs), CI workflows only install `@electron/asar` globally (`ci.yml:164`), and `fetch-electron-binary.js:82` should whitelist `['x64', 'arm64', 'armv7l', 'ia32']` with an actionable error so a 32-bit or exotic-arch host fails clearly instead of 404ing on Electron's release server.

---

PR #589, opened 2026-05-07
fix(frame-fix): skip redundant webContents.focus() under sloppy WMs (#416)

### LINKED ISSUE #416: Window raises to foreground on mouse hover with sloppy/focus-follows-mouse mode
Description:

  When using a window manager with "sloppy" focus mode (focus follows mouse without raising), the Claude Desktop window raises itself to the foreground whenever the mouse pointer enters it. This makes it difficult to work with overlapping windows, as Claude constantly covers other applications.

  Environment:
  - Linux Mint 22.3 - Cinnamon 64-bit
  - Cinnamon desktop with focus mode set to "sloppy" (auto-raise: false)
  - Claude Desktop installed via claude-desktop-debian

  Expected behavior:
  The window should receive keyboard focus without changing its z-order, like all other applications on the system.

  Actual behavior:
  The window jumps to the foreground on every mouse enter/focus event.

  Root cause:
  Claude Desktop's app code calls webContents.focus() in response to BrowserWindow focus events (~6 call sites). On Electron/X11, webContents.focus() triggers Chromium's X11Window::Activate(), which sends a _NET_ACTIVE_WINDOW client message. Per the EWMH spec, this is inherently a focus-and-raise request, so the window manager raises the window even though sloppy focus already granted keyboard focus without raising.

  This is a known upstream issue (electron/electron#38184) with no Electron flag or Chromium fix available.

  Possible fix in frame-fix-wrapper.js:
  Since frame-fix-wrapper.js already patches BrowserWindow behavior for Linux WM quirks, it could intercept webContents.focus() to skip the call when the window is already focused:

```
const originalWebContentsFocus = this.webContents.focus.bind(this.webContents);
  this.webContents.focus = () => {
    if (!this.isFocused()) {
      originalWebContentsFocus();
    }
  };
```

This would prevent the redundant _NET_ACTIVE_WINDOW raise while preserving focus behavior for cases where the window genuinely needs activation.

tkrag:
## Summary

Under sloppy / focus-follows-mouse WMs (Cinnamon Muffin, Mutter, i3 with
`focus_follows_mouse`), every BrowserWindow `focus` event triggered an
upstream `webContents.focus()` call which on Electron/X11 routes
through Chromium's `X11Window::Activate()` and sends a
`_NET_ACTIVE_WINDOW` client message — EWMH defines that as
focus-AND-raise, so the WM raised the window on every mouse-enter,
undoing the user's "no auto-raise" config.

Tracks [electron/electron#38184][upstream]; no Electron flag exposes
a focus-only path on X11.

[upstream]: https://github.com/electron/electron/issues/38184

Fixes #416.

## Approach

Hooked at `app.on('web-contents-created')` so every webContents in
the process is covered (main + child views + popups + devtools). The
upstream call site sits on a **child `WebContentsView`** (the
`claude.ai` host view), not the main BrowserWindow's own webContents,
so wrapping at the BrowserWindow-class level is insufficient — the
first version of this PR did that and missed the actual call site.

The guard runs against the **owning toplevel's** `isFocused()`, not
`wc.isFocused()`. `wc.isFocused()` returns `false` on a freshly-
attached child view even when the window is focused — exactly the
sloppy-hover state — so guarding on it would never skip.

A pass-through grace window after every `show` event (`SHOW_GRACE_MS`,
1000 ms) handles Electron's stale-`isFocused()` bug on
Cinnamon/KDE/Wayland — the same trap already addressed for KDE by
`scripts/patches/quick-window.sh`. After `hide()`, `isFocused()` can
keep returning `true` even though the WM never re-activated the
window; without the grace window a tray-restore would be silently
skipped and the window would come back visible-but-inert until
clicked.

## Scope

The guard wraps **every** Linux webContents — main, child views,
Quick Entry, About, devtools, utility processes — not just the main
window. The PR body's earlier "gated to non-popup main window" claim
was a docs lag from v1 (which had a `if (!popup)` gate that didn't
catch the child WebContentsView call site); the actual fix has to be
wide. Risk of stomping unrelated focus flows is bounded by the
grace-window logic.

## Trade-off (sloppy mode)

`webContents.focus()` wears two hats in Electron: X11 activation
(the bug) **and** renderer-focus direction. Skipping suppresses
both. So in sloppy mode, hovering over the Claude window now gives
WM focus (frame highlights, keystrokes route to our process) but
the renderer focus isn't directed to the claude.ai view until the
user clicks. Net effect: **one click per hover cycle** to start
typing, versus the constant raise that #416 reports. Documenting
this here rather than burying it — users self-selecting sloppy-focus
presumably prefer the former.

## Pre-existing issue surfaced during testing

On Cinnamon, Quick Entry submit fails to surface the main window on
attempts after the first hide → show cycle. Same root cause as the
KDE-gated patch in `scripts/patches/quick-window.sh` (stale
`isFocused()` after hide making upstream's `focusCheck() || show()`
short-circuit skip the show). **Reproduces without this PR.** Worth
a follow-up issue to widen the KDE patch to Cinnamon.

## Test plan (manual on Cinnamon 6.0 / Mint 22.3)

- [x] Sloppy mode, `auto-raise` off, hover over Claude with another
      window in front: Claude **does not raise**. (Primary bug fix.)
- [x] Sloppy mode: WM focus arrives on hover (frame highlights);
      typing into chat requires one click first — known trade-off.
- [x] Click-to-focus mode, send Claude to tray, right-click tray →
      "Show": window restores, X11 activates, first keystroke lands.
      Grace window prevents stale-`isFocused()` false-positive.
- [x] Click-to-focus mode, app launch: initial focus established
      (guard logs `PASS`, owner not yet focused).
- [x] Quick Entry first invocation works; subsequent invocations hit
      the pre-existing Cinnamon `show()` short-circuit (not this PR).

## Notes on the review

- **Scope mismatch (review point 1):** docs lag, fixed by rewriting
  the body to own the wide scope as above.
- **Timing premise (review point 2):** the regression is real — the
  click-to-focus tray-restore path *does* hit the stale-`isFocused()`
  trap on Cinnamon, captured with diagnostic logging. The grace
  window resolves it for the synchronous show → focus sequence. The
  deferred `setTimeout(() => e.webContents.focus(), 10)` site you
  flagged near `index.js:522197` (find-end / similar) is still
  vulnerable in theory — outside the grace window, with `isFocused()`
  stale-true, the call would be skipped. In practice it would only
  bite users on Cinnamon/KDE/Wayland *and* only on the brief moment
  after closing find UI; I haven't been able to reproduce a UX
  regression from it. Open to a wider guard if you'd rather see one,
  but the simplest read is "let the rare case lose renderer-focus
  re-direction for one frame; users click to recover" mirroring the
  sloppy trade-off.
- **Stray blank line (review point 3):** dropped.

---
30% AI / 70% Human
Claude: Drafted patch v1→v3 and PR text; ran diagnostic builds and
log captures end-to-end.
Human: Diagnosed the root cause from first principles in #416,
filed the issue, decided on iteration strategy (force-push v2/v3
on the fork), verified each build by running the AppImage on real
Cinnamon hardware, confirmed each behaviour change visually, made
the call on the sloppy-mode trade-off being acceptable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Tracks upstream Claude Desktop 1.8555.2.
 
 ### Fixed
 
+- Sloppy/focus-follows-mouse: suppress redundant `webContents.focus()` calls that trigger X11 `_NET_ACTIVE_WINDOW` raise-on-hover. Grace window handles stale `isFocused()` on tray-restore and minimize-restore. Thanks @tkrag. ([#589](https://github.com/aaddrick/claude-desktop-debian/pull/589), fixes [#416](https://github.com/aaddrick/claude-desktop-debian/issues/416))
 - Tray: extracted JS identifier captures now accept `$` so the 1.8089.1 minified bundle ('`i$A`' menu handler) matches. Switches `\w+` to `[\w$]+`. ([#627](https://github.com/aaddrick/claude-desktop-debian/pull/627), fixes [#625](https://github.com/aaddrick/claude-desktop-debian/issues/625))
 - RPM: silence "File listed twice" warning on `chrome-sandbox` by moving `chmod 4755` into `%install` (replaces `%attr` in `%files`). Adds regression guard that fails the build if the warning reappears. Thanks @JoshuaVlantis. ([#610](https://github.com/aaddrick/claude-desktop-debian/pull/610), fixes [#609](https://github.com/aaddrick/claude-desktop-debian/issues/609))
 - Window close with `CLAUDE_QUIT_ON_CLOSE=1` now actively quits via `app.quit()` instead of relying on the bundled handler that hardcodes hide-to-tray on Linux. Rides upstream's own quit-in-progress guard. Thanks @phelps-matthew. ([#624](https://github.com/aaddrick/claude-desktop-debian/pull/624), fixes [#623](https://github.com/aaddrick/claude-desktop-debian/issues/623))
--- a/README.md
+++ b/README.md
@@ -281,6 +281,7 @@ Special thanks to:
 - **[phelps-matthew](https://github.com/phelps-matthew)** for fixing `CLAUDE_QUIT_ON_CLOSE=1` to actively quit via `app.quit()` instead of relying on the bundled handler that hardcodes hide-to-tray on Linux, with thorough root cause analysis and alternatives evaluation (#624, #623)
 - **[dubreal](https://github.com/dubreal)** for `--password-store` keyring detection that probes D-Bus for kwallet6 / gnome-libsecret at startup, fixing session persistence on KDE Plasma and other desktops where Electron's `safeStorage` was unavailable (#611, #593)
 - **[JustinJLeopard](https://github.com/JustinJLeopard)** for detecting missing electron binaries after Node 24's `extract-zip` silently no-ops, with an `unzip` fallback that recovers from the `@electron/get` cache (#631, #584)
+- **[tkrag](https://github.com/tkrag)** for diagnosing and fixing the X11 window-raise-on-hover bug under sloppy/focus-follows-mouse WMs, tracing the upstream `webContents.focus()` → `_NET_ACTIVE_WINDOW` path through three iterations of review (#589, #416)
 
 ## Sponsorship
 
--- a/scripts/frame-fix-wrapper.js
+++ b/scripts/frame-fix-wrapper.js
@@ -246,6 +246,21 @@ Module.prototype.require = function(id) {
               this.setMenuBarVisibility(false);
             }
 
+            // Track the most recent 'show' event timestamp on the
+            // window. Read by the webContents.focus() guard below to
+            // distinguish a genuine post-show activation (which must
+            // pass through to send _NET_ACTIVE_WINDOW and actually
+            // give the window WM focus) from a sloppy-focus
+            // reassertion (which is what we want to skip). Required
+            // because Electron's isFocused() returns stale-true after
+            // hide() on Cinnamon/KDE/Wayland — a freshly-restored
+            // window reports focused=true even though the WM never
+            // activated it, and skipping the focus() call leaves the
+            // window visible-but-inert until the user clicks it.
+            // See #416 review notes.
+            this._lastShownAt = 0;
+            this.on('show', () => { this._lastShownAt = Date.now(); });
+
             // Inject CSS for Linux scrollbar styling
             this.webContents.on('did-finish-load', () => {
               this.webContents.insertCSS(LINUX_CSS).catch(() => {});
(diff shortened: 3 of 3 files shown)


tkrag:
Force-pushed a corrected version after testing the original locally — the v1 patch was hooking the wrong webContents and didn't actually fix the bug.

## What was wrong

v1 wrapped `this.webContents.focus` on the `PatchedBrowserWindow` instance — i.e. only the main BrowserWindow's own webContents. But in Claude's actual code path, `webContents.focus()` is called on a **child `WebContentsView`** (the one hosting `claude.ai`), whose webContents is a different object. v1 never ran on it.

I confirmed this by adding a diagnostic probe and hovering 3 times on Cinnamon sloppy focus:

```
[Focus Probe #1] wc=2 owner=1 wcFocused=false windows=[]
[Focus Probe #2] wc=2 owner=1 wcFocused=false windows=[]
[Focus Probe #3] wc=2 owner=1 wcFocused=false windows=[]
[Focus Probe #4] wc=2 owner=1 wcFocused=false windows=[]
[Focus Probe #5] wc=2 owner=1 wcFocused=false windows=[]
```

5 calls all on `wc=2` (a child view), `owner=1` (the main window). v1 was patching `wc.id=1`'s focus, which was never the call site. (Side note: `BrowserWindow.getAllWindows()` returns `[]` because of the wrapper's Proxy — `fromWebContents()` still works.)

## What v2 does

Hooks at `app.on('web-contents-created')` so every webContents (main + child views) gets the wrap. Skip is gated on the **owning toplevel's** `isFocused()` via `BrowserWindow.fromWebContents(wc)` — *not* on `wc.isFocused()`, which returns `false` on a freshly-attached child even when the X11 window is focused (i.e. it would never skip).

Verification log after fix:

```
[Focus Guard] wc=2 owner=1 ownerFocused=false PASS   ← initial app launch, focus established
[Focus Guard] wc=2 owner=1 ownerFocused=true SKIP    ← hover 1, raise prevented
[Focus Guard] wc=2 owner=1 ownerFocused=true SKIP    ← hover 2
[Focus Guard] wc=2 owner=1 ownerFocused=true SKIP    ← hover 3
[Focus Guard] wc=2 owner=1 ownerFocused=true SKIP
[Focus Guard] wc=2 owner=1 ownerFocused=true SKIP
```

Manually verified on Cinnamon Muffin (sloppy focus, auto-raise off):

- [x] Hover with another window in front → Claude no longer raises
- [x] Click on Claude → still focuses and raises normally
- [x] Typing in chat → input routes correctly (renderer focus established once on launch persists)
- [x] Quick Entry (Ctrl+Shift+Space) → opens, accepts input, submits

Diagnostic logging removed from the committed version; production patch is ~10 lines.

---

aaddrick (changes requested):
Requesting changes, but the diagnosis is right. `_NET_ACTIVE_WINDOW` lines up with EWMH and electron/electron#38184. Hooking at `app.on('web-contents-created')` (`scripts/frame-fix-wrapper.js:527`) instead of wrapping the BrowserWindow class is the right call — the upstream call site at `build-reference/app-extracted/.vite/build/index.js:243681` lives on a child `WebContentsView`, and its webContents isn't `this.webContents`. The prototype trick in `docs/learnings/test-harness-electron-hooks.md` wouldn't have caught that. Good find.

The first gap is scope. The PR body says the guard is gated to the non-popup main window inside the existing `if (!popup)` block, but the actual patch at `scripts/frame-fix-wrapper.js:575-582` sits in the `web-contents-created` hook with no popup gate. It wraps every Linux webContents — main WC, the claude.ai child view, Quick Entry, About, devtools, future popups, utility processes. That's a different patch than the Test Plan covers. Either restore the `if (!popup)` gate so body and code agree, or rewrite the body and Test Plan to own the actual blast radius and paste a Quick Entry launcher.log trace showing one PASS on initial show and SKIPs on subsequent focus events. It's a scope decision, not a docs refresh.

The second one is the timing premise. Your fix assumes `owner.isFocused()` is false during a genuine activation and true on a redundant sloppy reassertion. The Cinnamon log exercises the synchronous focus-event path and that holds. But there's a deferred call site upstream — `setTimeout(() => e.webContents.focus(), 10)` around index.js:522197 — where 10ms is enough for `isFocused()` to flip true before the deferred focus fires, and a genuine activation gets silently skipped. The click-to-focus case (GNOME/KDE default, not your sloppy-Cinnamon setup) is worse — X11 activates the window before the focus handler runs, so `owner.isFocused()` is already true on the first focus() of a tray-click-to-show cycle. Symptom would be "first keystroke ignored after raising from tray." Could you grab a launcher.log on GNOME or KDE default click-to-focus showing the child WebContentsView's focus() going through (not skipped) on a tray-click-to-raise? That's the one I'd want before merge.

Minor: stray blank line at `scripts/frame-fix-wrapper.js:458` inside the existing `if (!popup)` block looks like an editor artifact. Drop it. Placement near the existing Ctrl+Q `before-input-event` handler keeps the main-process plumbing together, which is the right instinct.

---

tkrag:
Force-pushed v3 (`37579c2`) addressing the review. Body rewritten end-to-end; flagging the diff here so you don't have to compare:

**Code changes since `e52c776`:**

1. **Added a `_lastShownAt` timestamp on every Linux BrowserWindow**, set in the existing `if (process.platform === 'linux')` block of the patched constructor, updated on every `show` event.
2. **Grace window in the focus guard:** within `SHOW_GRACE_MS` (1000 ms) of a `show` event, `wc.focus()` passes through unconditionally regardless of `isFocused()`. Outside the grace window, the existing `owner.isFocused()` guard applies.
3. **Dropped diagnostic `console.log`** that I'd been using for the runtime probes (`[Focus Guard] PASS/SKIP` lines).
4. **Dropped the stray blank line** at `scripts/frame-fix-wrapper.js:458`.

**Why the grace window:**

Your timing-premise concern was correct for the **click-to-focus tray-restore** path. Reproduced it on Cinnamon (with `focus-mode='click'`) using the diagnostic build:

```
PASS  visible=true   focused=false  sinceShow=inf   ← initial show
SKIP  visible=true   focused=true   sinceShow=inf   × 7
```

Right-clicking the tray icon → "Show" got the window visible but the post-show `webContents.focus()` was skipped (stale `isFocused()` from before hide), so the WM never got `_NET_ACTIVE_WINDOW` and the window sat visible-but-inert until I clicked it. Exactly the symptom you predicted.

With the grace window in v3, the same sequence shows `sinceShow≈50–100ms` on the post-show call, the guard passes through, X11 activation lands, and the first keystroke goes to chat input. Confirmed manually with two restore cycles.

**On the deferred `setTimeout(() => e.webContents.focus(), 10)` site:**

Still theoretically vulnerable outside the grace window — couldn't reproduce a user-visible regression from it on Cinnamon, but I haven't stress-tested find-bar dismissal specifically. Mentioned in the PR body. Happy to widen the guard (e.g. also track `lastFindEndAt` or just bump `SHOW_GRACE_MS` upward) if you'd rather close that gap explicitly.

**On the sloppy-mode renderer-focus trade-off:**

In sloppy mode, hover gives WM focus but typing requires one click — because `webContents.focus()` conflates X11 activation with renderer-focus direction, and skipping it loses both. No Electron API exposes a renderer-only path on X11 today. I think it's the right trade against the constant raise the issue reports, but you may disagree — it's specifically called out in the PR body's "Known trade-off" section so it doesn't slip past review.

**Pre-existing Cinnamon bug surfaced during testing:**

While testing tray-restore I noticed Quick Entry submit fails to surface the main window on attempts after the first hide → show cycle on Cinnamon. Same root cause as the KDE-only patch in `scripts/patches/quick-window.sh` — stale `isFocused()` making upstream's `focusCheck() || show()` skip the `show()`. Reproduces without this PR; I'll file a separate issue once #589 lands.

Ready for another round whenever you have time.

---

aaddrick (approved):
v3 addresses the timing and scope concerns from the first round. The grace window approach is solid. Your reproduction evidence on Cinnamon click-to-focus confirms it handles the stale `isFocused()` trap correctly. Comments are proportional to the other EWMH/platform patches in this file.

I rebased onto main, added two small things, and pushed directly to your branch:

1. `this.on('restore', ...)` alongside the `show` listener. On i3/sway (exactly the sloppy-focus crowd), minimize-to-restore fires `restore` but not `show`. A long minimize could leave `_lastShownAt` stale and the grace window wouldn't fire on the post-restore focus. One line, zero risk.
2. A one-line comment above the guard noting the deferred-setTimeout gap. Escape-key find-bar dismissal could lose renderer-focus direction outside the grace window. This way future maintainers know it's acknowledged rather than unknown.

Also added you to README Acknowledgments and a CHANGELOG entry.

The sloppy-mode trade-off (one click per hover cycle) is the right call against the constant raise. A one-line troubleshooting note in the docs would help so users don't file "typing doesn't work" issues, but that can land as a follow-up.

Good contribution. The v1-to-v3 iteration with reproduction evidence at each stage is appreciated. Approving and merging.

---

PR #592, opened 2026-05-07
test: add headless launch + --doctor smoke tests for AppImage artifact

sabiut:
## Problem

The AppImage artifact test (`tests/test-artifact-appimage.sh`) only validates package **structure** — extraction, AppDir layout, asar contents, desktop entry. It never actually executes the binary, so runtime-only regressions silently pass CI:

- `frame-fix-wrapper.js` syntax errors
- asar patches that produce invalid JS
- `frame-fix-entry.js` / `package.json` `main` field mismatches
- Electron startup crashes

The `.deb` test already runs `--doctor` as a lightweight smoke check (`tests/test-artifact-deb.sh:102-111`) — but the AppImage path had no equivalent.

## Changes

**`tests/test-artifact-appimage.sh`** — adds two new checks:

1. **`--doctor` parity test** (mirrors the deb smoke test). `--doctor` short-circuits before Electron starts (`scripts/packaging/appimage.sh:79-84`), so no display is required. Exit code `< 127` confirms the launcher dispatch + `doctor.sh` sourcing didn't crash.

2. **Headless launch test under Xvfb**. Runs the AppImage via `setsid xvfb-run dbus-run-session -- ...`, sleeps 10s, verifies the process is still alive, then tears down the whole process group. On failure, tails `launcher.log` and `xvfb-run` stderr to surface the actual error.

`setsid` is load-bearing — `xvfb-run`'s EXIT trap leaves Xvfb behind when the wrapper is killed by signal; running everything in a fresh process group lets `kill -- -PGID` reap the whole tree (xvfb-run, Xvfb, dbus, AppRun, electron, zygote children). `XDG_CACHE_HOME` is redirected to a temp dir so the test owns its own `launcher.log` without polluting the runner's cache.

**`.github/workflows/test-artifacts.yml`** — adds `xvfb`, `dbus-x11`, and `procps` to the Ubuntu apt install line (`procps` for `pkill` in the teardown sweep).

## Verification

- `shellcheck -x tests/test-artifact-appimage.sh` — clean
- `actionlint .github/workflows/test-artifacts.yml` — clean
- Local run against `claude-desktop-1.5354.0-amd64.AppImage`: **34/34 pass**, including both new tests
- `pgrep` after teardown: **zero orphan processes** (Xvfb, electron, etc. all reaped)

## Test plan

- [ ] CI `Test Build Artifacts` job for `appimage` matrix entry passes on this PR
- [ ] No orphan processes / hung jobs in the runner
- [ ] Failure path is exercised at least once in CI history before merging anything that breaks Electron startup (the test will surface `launcher.log` tail on crash)

---
80% AI / 20% Human
Claude: codebase audit, gap identification, test design (process-group teardown, log capture), implementation, local verification
Human: scope decision (full xvfb launch vs --doctor parity), review, direction

### DIFF
--- a/.github/workflows/test-artifacts.yml
+++ b/.github/workflows/test-artifacts.yml
@@ -44,7 +44,8 @@ jobs:
         if: matrix.format != 'rpm'
         run: |
           sudo apt-get update
-          sudo apt-get install -y file libfuse2 nodejs npm
+          sudo apt-get install -y file libfuse2 nodejs npm \
+            xvfb dbus-x11 procps
 
       - name: Run artifact tests
         run: |
--- a/tests/test-artifact-appimage.sh
+++ b/tests/test-artifact-appimage.sh
@@ -97,8 +97,8 @@ resources_dir="$appdir/usr/lib/node_modules/electron/dist/resources"
 validate_app_contents "$resources_dir" "${component_id}.desktop"
 
 # --- Doctor smoke test ---
-# --doctor checks system state; some checks will fail in CI (no display,
-# etc.) but the script itself should not crash with signal or 127.
+# Some --doctor checks fail in CI (no display, etc.); we only care that
+# the script itself didn't crash via signal or exec failure (>=127).
 doctor_exit=0
 "$appimage_file" --doctor >/dev/null 2>&1 || doctor_exit=$?
 if [[ $doctor_exit -lt 127 ]]; then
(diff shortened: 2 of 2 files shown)


aaddrick (commented):
Nice fill-in. The deb path has had `--doctor` smoke forever and the AppImage gap was real. `frame-fix-wrapper.js` syntax breaks and asar patch regressions would have sailed through structure-only checks. PR description is honest about scope. CI logs in run 25794059379 confirm both new tests fire: `[PASS] --doctor runs without crashing (exit: 2)` and `[PASS] AppImage stays alive under Xvfb for 10s`. Threshold choices are good. `< 127` for `--doctor` accepts real check failures without breaking on doctor output churn. 10s alive-under-Xvfb is the floor that catches startup-only crashes without false-flagging slow Electron boots. Process-group teardown via `setsid` + `kill -- -PGID`, plus the `pkill -f` zygote sweep at `tests/test-artifact-appimage.sh:168`, is the right belt-and-braces. `xvfb-run`'s EXIT trap does leak Xvfb on signal kill, so the comment is load-bearing.

One substantive ask before merge, and it's a comment, not a code change. The 10s probe proves the main process didn't `process.exit()` or segfault. It doesn't prove the app rendered. That matters because the class of crash I've been patching, #583 (commits `51e0bc7`, `368f834`, `88676f4`), is a GPU subprocess SIGTRAP. Electron's main process survives GPU child death and respawns it, so this test would PASS while users hit the bug. Xvfb has no GPU, so Electron falls back to SwiftShader and the GPU-crash path isn't exercised. A `loadURL` failure that leaves an idle main process also passes. Can you add a one-line disclaimer in the test header around `tests/test-artifact-appimage.sh:111`? Something like: "Catches main-process startup failures only; GPU/renderer-process crashes (e.g. #583-class) leave the main process alive and pass this check." Keeps a future contributor from claiming #583 coverage. An `xdotool search --name Claude` window-existence probe would tighten the signal as an optional follow-up.

Minor stuff. The `trap` at `tests/test-artifact-appimage.sh:135` installs after `launch_pid=$!` at line 131. Tiny window where an interrupt leaks Xvfb, contra the comment's "covers Ctrl-C". Negligible in practice. `unset XDG_CACHE_HOME` at line 171 clobbers any pre-existing value a local dev had set; `save_xdg=${XDG_CACHE_HOME:-}` then restore would be tidier, but it's CI-mostly anyway. Sandbox isn't a concern. `launcher-common.sh:115` already adds `--no-sandbox` unconditionally for AppImage, so the chrome-sandbox SUID dance in CI is sidestepped by design.

`procps` in the apt line is correct (`pkill` is in it), `dbus-x11` is needed for `dbus-run-session`, and `xvfb-run -a -s '-screen 0 1280x720x24'` covers the screen-query-on-startup case. Test runs in every PR build via the `test-artifacts` reusable workflow called from `ci.yml:56-59`, so the feedback loop value is there and it's not release-gated. Approve with the scope-disclaimer comment ask.

---

sabiut:
Thanks @aaddrick — scope disclaimer added in 61f938a as [`tests/test-artifact-appimage.sh:114-117`](https://github.com/aaddrick/claude-desktop-debian/blob/61f938a/tests/test-artifact-appimage.sh#L114-L117). Inlined the SwiftShader-fallback reasoning from your review so a future contributor doesn't try to fix the gap by switching Xvfb off.

Skipping the two non-blocking minors (trap-order race at line 135, `unset XDG_CACHE_HOME` at line 171) per your "negligible in practice" / "CI-mostly anyway" framing. Happy to revisit if either bites in practice.

Will open a separate issue for the `xdotool search --name Claude` follow-up — keeps it tracked without expanding this PR's scope.

PR #595, opened 2026-05-09
fix(rpm): set chrome-sandbox suid via %attr instead of %post chmod

### LINKED ISSUE #539: [bug]: rpm: chmod belongs to file section

```shell
n/a
```

### What happened

rpm postinstall includes `chmod 4755 "$SANDBOX_PATH"`. Instead, suid bit must be set in %files section.

### Steps to reproduce

n/a

### Expected behavior

n/a

### Logs / errors

```shell

```

### Anything else

```
rpm -q --scripts claude-desktop
postinstall scriptlet (using /bin/sh):
# Update desktop database for MIME types
update-desktop-database /usr/share/applications &> /dev/null || true

# Set correct permissions for chrome-sandbox
SANDBOX_PATH="/usr/lib/claude-desktop/node_modules/electron/dist/chrome-sandbox"
if [ -f "$SANDBOX_PATH" ]; then
    echo "Setting chrome-sandbox permissions..."
    chown root:root "$SANDBOX_PATH" || echo "Warning: Failed to chown chrome-sandbox"
    chmod 4755 "$SANDBOX_PATH" || echo "Warning: Failed to chmod chrome-sandbox"
fi
postuninstall scriptlet (using /bin/sh):
# Update desktop database after removal
update-desktop-database /usr/share/applications &> /dev/null || true

```

JoshuaVlantis:
## Summary

Closes #539. Move the chrome-sandbox suid bit out of the `%post` scriptlet and into the rpm spec's `%files` section via `%attr(4755, root, root)`.

The previous pattern only ran on fresh installs and silently regressed when the scriptlet was skipped (`--noscripts`, layered image builds, etc.), shipping a non-suid chrome-sandbox that breaks Electron sandboxing on every launch.

Also adds an `assert_setuid` helper to `tests/test-artifact-common.sh` and wires it up in `tests/test-artifact-rpm.sh` so a future spec regression to the old `%post` pattern fails CI rather than landing silently.

## Verification (Docker)

1. Built `.rpm` in `fedora:42` container with full build tooling.
2. Inspected rpm metadata: `rpm -qlv -p ./claude-desktop-*.rpm` shows `-rwsr-xr-x` on chrome-sandbox.
3. Confirmed `%post` no longer chmods: `rpm -q --scripts -p` shows only the `update-desktop-database` line.
4. Installed via `dnf install` in a fresh `fedora:42` container; `ls -la` and bash `[ -u <path> ]` confirm the suid bit.

## Note

`rpmbuild` emits `warning: File listed twice: /usr/lib/.../chrome-sandbox` because the file is covered by both the `/usr/lib/$package_name` directory listing and the explicit `%attr` line. This is benign — rpm uses the more specific `%attr` for the file's permissions, which is the intended behavior. The alternative (enumerating the entire `/usr/lib/$package_name` tree file-by-file in `%files`) trades the warning for a much larger and more brittle spec.

## Test plan

- [ ] CI builds rpm successfully
- [ ] `tests/test-artifact-rpm.sh` passes including the new `assert_setuid` check
- [ ] Smoke test: install via `dnf install ./claude-desktop-*.rpm` and confirm `ls -la /usr/lib/claude-desktop/.../chrome-sandbox` shows `-rwsr-xr-x`

### DIFF
--- a/scripts/packaging/rpm.sh
+++ b/scripts/packaging/rpm.sh
@@ -233,14 +233,6 @@ install -Dm 755 $staging_dir/claude-desktop %{buildroot}/usr/bin/claude-desktop
 # Update desktop database for MIME types
 update-desktop-database /usr/share/applications &> /dev/null || true
 
-# Set correct permissions for chrome-sandbox
-SANDBOX_PATH="/usr/lib/$package_name/node_modules/electron/dist/chrome-sandbox"
-if [ -f "\$SANDBOX_PATH" ]; then
-    echo "Setting chrome-sandbox permissions..."
-    chown root:root "\$SANDBOX_PATH" || echo "Warning: Failed to chown chrome-sandbox"
-    chmod 4755 "\$SANDBOX_PATH" || echo "Warning: Failed to chmod chrome-sandbox"
-fi
-
 %postun
 # Update desktop database after removal
 update-desktop-database /usr/share/applications &> /dev/null || true
@@ -248,6 +240,7 @@ update-desktop-database /usr/share/applications &> /dev/null || true
 %files
 %defattr(-, root, root, 0755)
 %attr(755, root, root) /usr/bin/claude-desktop
+%attr(4755, root, root) /usr/lib/$package_name/node_modules/electron/dist/chrome-sandbox
 /usr/lib/$package_name
 /usr/share/applications/claude-desktop.desktop
 /usr/share/icons/hicolor/*/apps/claude-desktop.png
--- a/tests/test-artifact-common.sh
+++ b/tests/test-artifact-common.sh
@@ -38,6 +38,14 @@ assert_executable() {
 	fi
 }
 
+assert_setuid() {
+	if [[ -u $1 ]]; then
+		pass "Setuid bit set: $1"
+	else
+		fail "Setuid bit not set: $1"
+	fi
+}
+
 assert_contains() {
 	local file="$1" pattern="$2" desc="${3:-}"
 	if grep -q "$pattern" "$file" 2>/dev/null; then
--- a/tests/test-artifact-rpm.sh
+++ b/tests/test-artifact-rpm.sh
@@ -41,9 +41,13 @@ electron_path='/usr/lib/claude-desktop/node_modules/electron/dist/electron'
 assert_file_exists "$electron_path"
 assert_executable "$electron_path"
 
-# chrome-sandbox
-assert_file_exists \
-	'/usr/lib/claude-desktop/node_modules/electron/dist/chrome-sandbox'
+# chrome-sandbox: setuid bit must be set by the rpm spec's %files
+# %attr(4755, ...) entry, not by a %post chmod (#539). The check
+# guards against regressing the spec to the old %post chmod pattern,
+# which leaves chrome-sandbox unsuid'd if the scriptlet is skipped.
+chrome_sandbox='/usr/lib/claude-desktop/node_modules/electron/dist/chrome-sandbox'
+assert_file_exists "$chrome_sandbox"
+assert_setuid "$chrome_sandbox"
 
 # --- Desktop entry validation ---
 desktop_file='/usr/share/applications/claude-desktop.desktop'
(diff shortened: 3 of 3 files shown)


sabiut (approved):
LGTM — verified locally and the change is the right shape.

Switching to `%attr(4755, root, root)` in `%files` makes the setuid bit
declarative and recorded in RPM's package database, so `rpm -V` will catch
drift and the bit survives `--noscripts` / layered image builds (rpm-ostree,
Silverblue) where the old `%post chmod` would silently no-op. The `%attr`
line is placed before the `/usr/lib/$package_name` directory catch-all, so
RPM applies the more-specific rule to chrome-sandbox.

**Test side**

- `assert_setuid` is a clean drop-in alongside `assert_executable` etc.,
  same tabs / `[[ ]]` / pass-fail style.
- I sanity-checked the helper against real files: 4755 → PASS, 0755 → FAIL,
  1755 (sticky-only) → FAIL. `[[ -u ... ]]` correctly discriminates setuid
  from sticky, so no false positives on that axis.
- Placement after `rpm -ivh` in `test-artifact-rpm.sh` means the assertion
  runs against the installed file mode, which is what we want.
- The explanatory comment is helpful; one small thing — it reads as if the
  guard is specific to the old `%post chmod` pattern, but really it catches
  any regression that strips the suid bit. Up to you whether that's worth
  a tweak; not a blocker.

**Other**

The "File listed twice" warning is benign — explicit `%attr` wins over the
directory listing, so the rationale in the PR description holds.

JoshuaVlantis:
Updated the comment.

aaddrick (approved):
Right fix, right shape. Moving the suid bit from a `%post` chmod to `%attr(4755, root, root)` in `%files` (`scripts/packaging/rpm.sh:243`) is the canonical RPM idiom. The mode lives in the package database, `rpm -V` catches drift, and the bit survives `--noscripts`, `rpm-ostree`/Silverblue layered builds, and any other path that skips scriptlets. The old `%post` block is fully gone. The remaining `%post` at `scripts/packaging/rpm.sh:232-234` is just `update-desktop-database`, which is correct. Cross-distro isn't a concern — `%attr` is core spec syntax, identical on Fedora, RHEL, openSUSE, Mageia. AppArmor/SELinux are orthogonal: the suid bit is a filesystem attribute, MAC confinement happens at exec time, so confinement behavior doesn't change.

Scope is correctly RPM-only. `scripts/packaging/deb.sh:211-227` still uses the postinst chmod (separate question, not this PR's problem), and the AppImage launcher runs with `--no-sandbox` unconditionally at `scripts/launcher-common.sh:115`, so chrome-sandbox suid is moot there. The "File listed twice" warning in the PR body is benign — explicit `%attr` on the file wins over the directory catch-all `/usr/lib/$package_name`, which is what you want. Enumerating the whole tree trades the warning for a much more brittle spec.

Test side is clean. `assert_setuid` in `tests/test-artifact-common.sh:41-47` uses `[[ -u $1 ]]`, which discriminates setuid from sticky (`[[ -k ]]`) and from plain executable. The check runs after `rpm -ivh` at `tests/test-artifact-rpm.sh:50`, so it asserts the installed mode the spec produced, not anything `%post` did. That's the regression guard you want for #539. If someone reverts to `%post chmod` and the install path skips scriptlets, this test fails instead of shipping a broken sandbox. The comment at `tests/test-artifact-rpm.sh:44-48` already addresses the "any regression that strips suid" framing @sabiut flagged. One gap worth a follow-up issue (not a blocker): `tests/test-artifact-rpm.sh:27` uses `rpm -ivh --nodeps` without `--noscripts`, so the test confirms `%attr` produces the right mode but doesn't exercise the scriptlet-skipped path that motivated the PR. A parallel `--noscripts` install (or a second matrix entry) would close that loop.

Pre-merge note for @aaddrick: `gh pr checks 595` reports "no checks reported" because the queued runs are sitting at `conclusion: action_required`. That's the first-time-contributor approval gate (this is @JoshuaVlantis's first PR here), not path-filtering. `workflow_dispatch` from a fork branch won't release them — you'll need to click "Approve and run workflows" on the PR's Checks tab. Once they're green, ship it.

---

PR #596, opened 2026-05-09
fix(linux): no-op autoUpdater on Linux to defend against feed activation

### LINKED ISSUE #567: Disable Electron auto-updater on Linux (currently relies on a happy accident)
The static grounding sweep on `docs/compat-matrix` flagged S26 ("auto-update doesn't fire / corrupt the install") as Missing in build 1.5354.0. Writing it up here so we can pick a fix before it bites us.

## What I found

The autoUpdater gate inside the bundled app is structurally open on Linux when `app.isPackaged` is true:

- `lii()` at `build-reference/app-extracted/.vite/build/index.js:508761-508774` returns `true` on Linux when the app is packaged.
- The truthy path sets the autoUpdater feed URL to `api.anthropic.com/api/desktop/linux/...` and registers update-check listeners.
- Our launcher exports `ELECTRON_FORCE_IS_PACKAGED=true` at `scripts/launcher-common.sh:249`, so `app.isPackaged` is true on every install we ship — deb, rpm, AppImage, AUR.

The only reason this doesn't actually try to download anything today is that Electron's Linux `autoUpdater` is unimplemented. It logs `AutoUpdater is not supported on Linux` and no-ops. That's a happy accident. If a future Electron bump implements Linux auto-update, or upstream switches to `electron-updater`, every install in the wild will start hitting that Linux feed URL — and upstream doesn't ship Linux binaries through their feed. Best case: 404s on every check. Worst case: serves something that doesn't match what we shipped, which our installs aren't prepared for.

Case doc with the full anchor list: [`docs/testing/cases/distribution.md` S26](https://github.com/aaddrick/claude-desktop-debian/blob/docs/compat-matrix/docs/testing/cases/distribution.md#s26--auto-update-is-disabled-when-installed-via-apt--dnf).

## Two fixes I see

**Option 1 — Wrap `hA.autoUpdater` in `frame-fix-wrapper.js` to no-op on Linux.**

Same pattern as the BrowserWindow defaults override. Smallest diff, lowest risk. Defends against upstream changes regardless of what `lii()` decides to gate on in a future build.

**Option 2 — Gate `ELECTRON_FORCE_IS_PACKAGED` on the package format in `scripts/launcher-common.sh`.**

Set it true only on AppImage. Leave it unset on deb/rpm where APT/DNF handle updates. More semantically correct — Electron's view of "packaged" lines up with what the user's package manager actually does. Doesn't fix the issue if `lii()` checks something other than `app.isPackaged` in a future build.

I'm leaning toward Option 1 because it's defensive and the wrapper already exists for exactly this kind of thing. Option 2 has nicer semantics but a thinner safety margin.

Anyone have a preference, or a third option I'm missing?

---
Written by Claude opus-4.7 via [Claude Code](https://claude.ai/code)

JoshuaVlantis:
## Summary

Closes #567. Wraps `require('electron').autoUpdater` on Linux with a Proxy that no-ops every method and returns chainable stubs for EventEmitter calls, installed via the existing `Module.prototype.require` interceptor in `frame-fix-wrapper.js`.

## Why

The bundled app's `lii()` gate sets a feed URL of `api.anthropic.com/api/desktop/linux/...` when `app.isPackaged` is true (which we set unconditionally via `ELECTRON_FORCE_IS_PACKAGED=true`), and registers update-check listeners. Today this is a happy accident: Electron's Linux `autoUpdater` is unimplemented and the calls log `"AutoUpdater is not supported on Linux"` and no-op.

If a future Electron implements Linux `autoUpdater`, every install in the wild would start hitting that feed and either 404 or — worse — receive content the install wasn't prepared for. `.deb`/`.rpm`/AppImage updates flow through the OS package manager (or AppImageUpdate); the Anthropic feed has no Linux artifacts.

This is the defensive option from the issue (Option 1) — the alternative (Option 2: gating `ELECTRON_FORCE_IS_PACKAGED` on package format) has nicer semantics but a thinner safety margin if `lii()` later checks something other than `app.isPackaged`.

## Verification (Docker)

1. Built `.deb` in `ubuntu:24.04` container.
2. Extracted shipped `frame-fix-wrapper.js` from `app.asar`; confirmed both the `autoUpdaterNoop` Proxy definition and the `prop === 'autoUpdater' && process.platform === 'linux'` intercept are present.
3. **Runtime test:** loaded the SHIPPED wrapper from the built `.deb` with a mock electron module whose `autoUpdater` records every method call. Replayed every bundled-app pattern observed in `index.js`:
   - `gA.autoUpdater.setFeedURL('https://api.anthropic.com/...')`
   - `gA.autoUpdater.on('error', ...)` / `on('update-available', ...)`
   - `gA.autoUpdater.checkForUpdates()`
   - `gA.autoUpdater.quitAndInstall()`
   - `gA.autoUpdater.getFeedURL()` (returns `''`)
   - Truthy check `A.autoUpdater && n(...)` (still passes — Proxy is truthy)
   - Chained `.on().on()` (returns the Proxy for chaining)

   **Zero real autoUpdater calls were recorded.** The Proxy intercepts every access path the bundled app uses.

## Test plan

- [ ] CI builds deb/rpm/AppImage successfully
- [ ] Smoke test: launch installed deb, confirm no `api.anthropic.com/api/desktop/linux` requests in `~/.cache/claude-desktop-debian/launcher.log` (or via tcpdump/strace)
- [ ] App still starts cleanly and update-related code paths don't throw

### DIFF
--- a/scripts/frame-fix-wrapper.js
+++ b/scripts/frame-fix-wrapper.js
@@ -120,12 +120,21 @@ const LINUX_CSS = `
 // autoUpdater no-op: every property access returns a chainable function
 // so `.on(...).once(...).setFeedURL(...).checkForUpdates()` is harmless.
 // `getFeedURL` returns '' so any code that inspects the URL gets a
-// well-typed empty string rather than undefined. Defined once and reused
-// across all require('electron') calls. Linux-only; macOS/Windows still
-// see the real autoUpdater. See #567.
+// well-typed empty string rather than undefined. `then`/`catch`/`finally`
+// and `Symbol.toPrimitive`/`Symbol.iterator` resolve to `undefined` so the
+// Proxy is not mistaken for a thenable (which would call chainNoop as
+// `then(resolve, reject)` and never resolve — silent await hang) or
+// asked to coerce to a primitive. Writes land on the target but are
+// shadowed by the get-trap. Defined once and reused across all
+// require('electron') calls. Linux-only; macOS/Windows still see the
+// real autoUpdater. See #567.
 const autoUpdaterNoop = new Proxy({}, {
   get(_target, prop) {
     if (prop === 'getFeedURL') return () => '';
+    if (prop === 'then' || prop === 'catch' || prop === 'finally'
+      || prop === Symbol.toPrimitive || prop === Symbol.iterator) {
+      return undefined;
+    }
     return function chainNoop() { return autoUpdaterNoop; };
   },
 });
--- a/scripts/frame-fix-wrapper.js
+++ b/scripts/frame-fix-wrapper.js
@@ -117,6 +117,19 @@ const LINUX_CSS = `
   }
 `;
 
+// autoUpdater no-op: every property access returns a chainable function
+// so `.on(...).once(...).setFeedURL(...).checkForUpdates()` is harmless.
+// `getFeedURL` returns '' so any code that inspects the URL gets a
+// well-typed empty string rather than undefined. Defined once and reused
+// across all require('electron') calls. Linux-only; macOS/Windows still
+// see the real autoUpdater. See #567.
+const autoUpdaterNoop = new Proxy({}, {
+  get(_target, prop) {
+    if (prop === 'getFeedURL') return () => '';
+    return function chainNoop() { return autoUpdaterNoop; };
+  },
+});
+
 // Build the patched BrowserWindow class and Menu interceptor once,
 // on first require('electron'), then reuse via Proxy on every access.
 let PatchedBrowserWindow = null;
@@ -740,6 +753,23 @@ X-GNOME-Autostart-enabled=true
               return Reflect.get(menuTarget, menuProp);
             }
           });
+        }
+        if (prop === 'autoUpdater' && process.platform === 'linux') {
+          // Force autoUpdater into a no-op on Linux. Upstream's bundled
+          // app code sets a feed URL of api.anthropic.com/api/desktop/linux/...
+          // when app.isPackaged is true (we set ELECTRON_FORCE_IS_PACKAGED=true
+          // unconditionally). Today this is a happy accident: Electron's Linux
+          // autoUpdater is unimplemented and logs "AutoUpdater is not supported
+          // on Linux", so the calls no-op. If a future Electron implements it,
+          // every install would start hitting that feed and would either 404
+          // or — worse — receive content the install wasn't prepared for.
+          // .deb/.rpm/AppImage updates flow through the OS package manager
+          // (or AppImageUpdate); the Anthropic feed has no Linux artifacts.
+          // We replace the entire autoUpdater object with a Proxy that
+          // no-ops every method and returns chainable stubs for EventEmitter
+          // calls so listener registration in the bundled code is harmless.
+          // See #567.
+          return autoUpdaterNoop;
         }
         return Reflect.get(target, prop, receiver);
       }


aaddrick (commented):
Thanks for picking this up @JoshuaVlantis. Option 1 is what I was leaning toward and the shape matches what I had in mind. Wrapper-based, Linux-gated, defined once outside the require closure so we're not allocating a fresh Proxy per `require('electron')`. The `getFeedURL` carve-out to return `''` instead of a chained Proxy is the right call. Anything expecting a string gets one.

I replayed the bundled patterns against the Proxy locally and coverage holds. The truthy gate at `index.js:51901` (`A.autoUpdater && n(...)`), destructure-style `n(CA.autoUpdater, ...)`, `setFeedURL({ url, serverType })` at `:511476` and `:511579`, chained `.once('error', u).once('update-downloaded', I)` at `:511452-454`, `removeListener` at `:511442-447`, the listener bank at `:511234-511288`, `checkForUpdates` / `quitAndInstall` / `downloadUpdate`. All fine. The Sentry breadcrumb wrap at `:51865-889` reads `s.emit.bind(s)` at wrap-time, which now resolves through the get-trap to `chainNoop`, but real exceptions still surface via the unhandled-error path. The `desktop_update_*` breadcrumbs vanish, which is fine for traffic that never happens. The "Check for Updates" menu item at `:526280-526300` only renders in `case "ready"` (update already downloaded) and state never advances, so it never appears.

One thing to fix before merge. `scripts/frame-fix-wrapper.js:126-131` returns `chainNoop` for every property access, including `then`. The first time anything does `await someAutoUpdaterExpr` or `Promise.resolve(autoUpdater).then(...)`, V8's thenable check calls `chainNoop(resolve, reject)`, which returns the Proxy and never invokes `resolve`. Silent startup hang, no error, no Sentry breadcrumb. The bundle-pattern replay doesn't catch it because nothing in today's bundle awaits autoUpdater, but a future re-minification could introduce it. Easiest fix is to extend the `getFeedURL`-style allowlist to also return `undefined` for `then` / `catch` / `finally` / `Symbol.toPrimitive` / `Symbol.iterator`. Two-line change and the Proxy becomes non-thenable and non-iterable by design.

While you're in there, the comment block on that same Proxy says "every property access returns a chainable function," which is accurate but understates it a bit. `getFeedURL` is the documented exception and writes land on the target while reads are masked. Something like `// Note: writes land on the target but are shadowed by the get trap.` would save the next reader five minutes. Not blocking. Approach matches `docs/learnings/patching-minified-js.md` — wrapper-over-sed, no anchoring to minified names, survives upstream re-minification. Approve once `then` is handled.

---

JoshuaVlantis:
Thanks @aaddrick — pushed `8796aa2` to handle `then`.

`scripts/frame-fix-wrapper.js:128-131` masks `then` / `catch` / `finally` / `Symbol.toPrimitive` / `Symbol.iterator` to `undefined` in the get-trap, exactly as suggested. Comment block updated to note that writes land on the target but reads are shadowed by the get-trap.

Verified against the actual Proxy in `node:20-alpine`:

- before: `await proxy`, `Promise.resolve(proxy).then(...)`, and `await proxy.checkForUpdates()` all hung the full 1500ms timeout
- after: all three resolve to the Proxy; documented chain (`.on().once().setFeedURL().checkForUpdates()`, `emit.bind(proxy)`) and `getFeedURL() === ''` carve-out unchanged

Bonus while wiring up the harness: `'' + autoUpdaterNoop` previously threw `Cannot convert object to primitive value` because `Symbol.toPrimitive` resolved to `chainNoop`. The `Symbol.toPrimitive → undefined` line you suggested fixes that too.

Re-requesting review.

---

aaddrick (approved):
Thanks @JoshuaVlantis — `8796aa2` is exactly the allowlist shape I was after. `then`/`catch`/`finally`/`Symbol.toPrimitive`/`Symbol.iterator` all return `undefined` from the get-trap, everything else still resolves to `chainNoop`. The Proxy is non-thenable and non-iterable by design now, so any future `await someAutoUpdaterExpr` path won't hang on V8's thenable check.

Nice catch on the `'' + autoUpdaterNoop` primitive-coercion path too. I hadn't flagged it, but masking `Symbol.toPrimitive` to `undefined` falls out of the same fix and kills a separate `Cannot convert object to primitive value` failure mode for free. Diff is +22/-0 on `frame-fix-wrapper.js`, all 13 checks green. Merging.

---

PR #597, opened 2026-05-09
fix(node-pty): clean upstream Windows binaries before staging Linux build

### LINKED ISSUE #401: [bug] Terminal fails to load on Fedora RPM — app.asar ships Windows node-pty binaries
## Summary

On the Fedora RPM build, the integrated terminal fails to load with the message **"Failed to load terminal backend"**. Root cause: `app.asar` ships Windows `node-pty` binaries (PE32+ DLLs) instead of — or in addition to — the Linux-compiled ones, so Electron extracts the wrong binary to `/tmp` and `dlopen()` fails with `invalid ELF header`.

## Environment

- **OS:** Fedora 44 (Workstation)
- **Architecture:** x86_64
- **Claude Desktop:** `1.2278.0-1.3.30.fc42` (RPM from `aaddrick.github.io/claude-desktop-debian/rpm`)
- **Installed via:** `sudo dnf install claude-desktop`

## Steps to reproduce

1. Install Claude Desktop RPM on Fedora 44.
2. Launch Claude Desktop.
3. Open a session and toggle the integrated Terminal panel.

**Observed:** Right panel shows `Failed to load terminal backend`.
**Expected:** A working shell PTY.

## Root cause (from `~/.config/Claude/logs/main.log`)

```
[error] Failed to load node-pty {
  error: Error: /tmp/.org.chromium.Chromium.XXXXXX: invalid ELF header
    at Module._extensions..node (node:internal/modules/cjs/loader:1980:18)
    ...
    at Object.<anonymous> (.../app.asar/node_modules/node-pty/lib/unixTerminal.js:24:11)
}
```

Inspecting the shipped archive:

```
$ npx @electron/asar list /usr/lib/claude-desktop/.../app.asar | grep node-pty/build
/node_modules/node-pty/build/Release/conpty.node
/node_modules/node-pty/build/Release/conpty_console_list.node
/node_modules/node-pty/build/Release/pty.node            # ← PE32+ Windows DLL
/node_modules/node-pty/build/Release/winpty-agent.exe
/node_modules/node-pty/build/Release/winpty.dll
```

When `lib/unixTerminal.js` requires `../build/Release/pty.node`, Node resolves through the asar to the Windows DLL (instead of falling through to the correctly-compiled Linux binary in `app.asar.unpacked/`). Electron extracts the DLL to `/tmp` and `dlopen()` rejects it.

Comparing with PR #152's `build.sh`: the PR copies only `lib/` into `app.asar.contents/node_modules/node-pty/` and places native binaries under `app.asar.unpacked/`. In the current RPM, `build/Release/` **also** ends up inside `app.asar` with the Windows binaries from the upstream macOS bundle — so the unpack-redirect never kicks in.

## Suggested fix

In `build.sh`, after copying `lib/` into `app.asar.contents`, **also strip any `build/` tree from the asar contents**:

```bash
# Remove any native binaries from the asar (they must live in app.asar.unpacked only)
rm -rf "$APP_STAGING_DIR/app.asar.contents/node_modules/node-pty/build"
```

Alternative: pack asar with `--unpack "**/*.node" --unpack "**/*.dll" --unpack "**/*.exe"` so even if files are present in the staging tree they are redirected out of the asar at pack time.

## Workaround for affected users

Until fixed upstream, users can patch their install locally:

```bash
# Backup
sudo cp /usr/lib/claude-desktop/node_modules/electron/dist/resources/app.asar{,.bak}

# Extract, replace, repack
cd /tmp && sudo npx @electron/asar extract \
  /usr/lib/claude-desktop/node_modules/electron/dist/resources/app.asar \
  claude-app-extracted

sudo cp /usr/lib/claude-desktop/node_modules/electron/dist/resources/app.asar.unpacked/node_modules/node-pty/build/Release/pty.node \
        /tmp/claude-app-extracted/node_modules/node-pty/build/Release/pty.node

sudo npx @electron/asar pack /tmp/claude-app-extracted \
  /usr/lib/claude-desktop/node_modules/electron/dist/resources/app.asar \
  --unpack "**/*.node"

# Restart Claude Desktop
```

(If `app.asar.unpacked/node_modules/node-pty/` does not exist, rebuild it first with `sudo npm install node-pty@1.1.0 --build-from-source --runtime=electron --target=41.2.0 --dist-url=https://electronjs.org/headers` in that path.)

Happy to open a PR if useful.

JoshuaVlantis:
## Summary

Closes #401. `rm -rf` the upstream-extracted `node_modules/node-pty/` in `app.asar.contents/` before staging the locally-compiled Linux build's `lib/`, `package.json`, and `build/`.

## Why

The Windows installer's `app.asar` ships node-pty with Windows binaries — `winpty.dll`, `winpty-agent.exe`, and Windows-format `build/Release/*.node` files (`pty.node`, `conpty.node`, `conpty_console_list.node`). After upstream extraction these sit in `app.asar.contents/node_modules/node-pty/`. The existing `cp -r $pty_src_dir/build` only overwrites same-named files, so:

- Orphan Windows binaries (`winpty.dll`, `*.exe`, `conpty*.node`) persist inside the asar.
- Worse, when `npm install node-pty` silently fails (e.g., missing C build tools — the rpm CI image briefly hit this during my testing) `pty.node` itself stays as a Windows DLL. `unixTerminal.js`'s `require('../build/Release/pty.node')` then resolves through the asar to a PE32+ DLL that `dlopen()` rejects, breaking the integrated terminal with the symptoms in #401.

After this change, `app.asar` only contains Linux node-pty: lib (cross-platform JS), package.json, and a Linux-ELF `pty.node` (which `--unpack '**/*.node'` redirects to `app.asar.unpacked/`).

## Verification (Docker)

Built **both** `.deb` (in `ubuntu:24.04`) and `.rpm` (in `fedora:42` with `gcc gcc-c++ make`) and inspected the shipped artifacts:

```
$ asar list app.asar | grep node-pty/build
/node_modules/node-pty/build/binding.Makefile
/node_modules/node-pty/build/config.gypi
/node_modules/node-pty/build/Makefile
/node_modules/node-pty/build/pty.target.mk
/node_modules/node-pty/build/Release
/node_modules/node-pty/build/Release/pty.node

$ file .../app.asar.unpacked/node_modules/node-pty/build/Release/pty.node
... ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked
```

No `.dll`, `.exe`, or Windows `.node` files remain inside the asar on either format.

## Note on remaining cosmetic leftovers

`app.asar.unpacked/node_modules/node-pty/build/Release/` still contains the upstream Windows installer's `winpty.dll`, `winpty-agent.exe`, and Windows `conpty*.node` files because they came in via `cp -a` of the Windows `app.asar.unpacked/`. Linux runtime never `require()`s them (they're outside `unixTerminal.js`'s code path), so they're harmless dead weight rather than a bug. A follow-up could clean those in `finalize_app_asar` (`scripts/staging/electron.sh`) before staging Linux binaries — out of scope here.

## Latent issue surfaced during testing

When `npm install node-pty` silently fails in `install_node_pty()` (because the build env lacks gcc/make), `pty_src_dir` is left empty and the entire copy block — including this `rm -rf` — is skipped. The result is a build that uses upstream Windows binaries unchanged. The fix in this PR makes that failure mode no worse than today (the bug was pre-existing) but doesn't resolve it. Worth a separate issue/PR to either fail the build loudly when node-pty install fails, or to enforce build-tool dependencies in `check_dependencies()`.

## Test plan

- [ ] CI builds both .deb and .rpm successfully
- [ ] `asar list` on shipped `app.asar` shows no Windows artifacts
- [ ] `file` on shipped `pty.node` reports ELF
- [ ] Smoke test: open the integrated terminal in Claude Desktop on a Linux install — should spawn a shell instead of "Failed to load terminal backend"

### DIFF
--- a/scripts/patches/cowork.sh
+++ b/scripts/patches/cowork.sh
@@ -832,6 +832,15 @@ install_node_pty() {
 
 	if [[ -n $pty_src_dir && -d $pty_src_dir ]]; then
 		echo 'Copying node-pty JavaScript files into app.asar.contents...'
+		# Wipe the upstream-extracted node-pty before staging the Linux
+		# build. The Windows installer's app.asar ships node-pty with
+		# Windows binaries (winpty.dll, winpty-agent.exe, Windows
+		# build/Release/*.node files). `cp -r $pty_src_dir/build` only
+		# overwrites same-named files; orphan Windows binaries persist
+		# inside the asar, surface as PE32+ when users inspect with
+		# `asar list`, and pollute /tmp via Electron's lazy-extract on
+		# any spurious require() (#401).
+		rm -rf "$app_staging_dir/app.asar.contents/node_modules/node-pty"
 		mkdir -p "$app_staging_dir/app.asar.contents/node_modules/node-pty" || exit 1
 		# --no-preserve=mode so read-only bits from the Nix store
 		# (--node-pty-dir) don't propagate into the staging tree.


aaddrick (approved):
Right shape, right placement. The `rm -rf` at `scripts/patches/cowork.sh:828` lands after `patch_app_asar` extracts the upstream Windows `app.asar` into `app.asar.contents/` (`scripts/patches/app-asar.sh:16`), and before `finalize_app_asar` packs it back with `--unpack '**/*.node'` (`scripts/staging/electron.sh:19-20`). The `patch_app_asar` → `install_node_pty` → `finalize_app_asar` ordering at `build.sh:285-287` is what makes this work, and the PR doesn't perturb it. Targeting only `app.asar.contents/node_modules/node-pty` is the conservative cut. Nothing else in `node_modules/` ships Windows binaries that matter at runtime, and the leftovers in `app.asar.unpacked/` are correctly flagged as out-of-scope.

The wipe is layout-agnostic, which is the strong property. Whole-directory `rm -rf` instead of a glob over extensions, so `.dll`, `.exe`, `.pdb`, `winpty/`, and any future `prebuilds/win32-*/` layout get caught the same way. The `mkdir -p` and copy block restages Linux content from scratch, so there's no reliance on `cp -r` overwrite behavior. Idempotency on `--clean=no` isn't a concern either: `setup_work_directory` at `build.sh:262` runs unconditionally and does `rm -rf "$work_dir"` (`scripts/setup/dependencies.sh:103`). The `cleanup_action=no` flag only short-circuits the post-build `cleanup_build` at `build.sh:175-176`. The wipe always operates on a fresh upstream tree.

Format coverage checks out. `.deb`, `.rpm`, `.AppImage`, and Nix all consume the same `$app_staging_dir/app.asar` and `app.asar.unpacked/` (verified at `scripts/packaging/{deb,rpm,appimage}.sh` and the Nix install path in `build.sh:288-298`). One fix lands on all four. The Nix path uses `--node-pty-dir` (`scripts/patches/cowork.sh:798-801`), but the `rm -rf` still runs in that branch, so Nix users get the same cleanup. The silent-failure mode where `npm install node-pty` exits non-zero and the whole block including this `rm -rf` gets skipped (`scripts/patches/cowork.sh:818`) is correctly called out in the PR body, and #598 closes it. The code paths are disjoint, so either merge order is fine.

Two non-blockers. This is a scrub rather than a structural fix. Upstream node-pty is multi-arch and we restage Linux JS on top without first removing the upstream `node_modules/node-pty/` wholesale. Dropping the upstream copy entirely before staging would be more durable, but the current diff is smaller and verified. Second, the comment at `scripts/patches/cowork.sh:820-827` frames the failure as polluting `/tmp` via Electron's lazy-extract on any spurious `require()`, which is slightly stronger than what #401's log shows. The observed failure is `unixTerminal.js`'s explicit `require('../build/Release/pty.node')` hitting a Windows DLL and `dlopen()` rejecting it with `invalid ELF header`. Doesn't change the fix.

---

Issue #548, opened 2026-05-01
launcher: log session/IME environment alongside flag decisions


aaddrick:
The launcher currently logs the Electron flag-decision path (`scripts/launcher-common.sh`) but doesn't capture the session/IME environment that caused those decisions. When users report input or display issues (e.g. #545), we have to round-trip multiple `echo $VAR` requests before we can reason about their session.

This is foundational work that makes future input/display bug reports self-contained.

## Acceptance criteria

- [ ] `setup_logging` (or a new helper invoked once per launch) writes a single `env=` block to `launcher.log` covering at minimum:
  - `XDG_SESSION_TYPE`, `WAYLAND_DISPLAY`, `DISPLAY`, `XDG_CURRENT_DESKTOP`
  - `GTK_IM_MODULE`, `XMODIFIERS`, `QT_IM_MODULE`
  - `CLAUDE_USE_WAYLAND`, `CLAUDE_TITLEBAR_STYLE`, `CLAUDE_GTK_IM_MODULE` (if-set)
- [ ] Empty/unset values are logged as `KEY=` rather than omitted, so absence is unambiguous
- [ ] No PII or paths beyond what's already in env vars
- [ ] BATS coverage in `tests/launcher-common.bats` asserts the env block is present and key vars appear

## Files

- `scripts/launcher-common.sh` (primary)
- `tests/launcher-common.bats` (coverage)

## Out of scope

- Rotating or capping `launcher.log` size — separate concern
- Adding env vars beyond observation (no behavior changes)
- Doctor command changes — tracked separately

## Notes

Small, mechanical, blocks nothing. Good first PR for this thread, and unblocks the doctor-check follow-up by ensuring relevant env state is already on disk when users report issues.

Refs: #545

---
Written by Claude Opus 4.7 via [Claude Code](https://claude.ai/code)

PR #599, opened 2026-05-10
fix(launcher): black screen and session loss on Fedora 44 KDE

### LINKED ISSUE #593: [bug]: RPM/Fedora 44 KDE: session not persisted — Cookies file always 0 bytes, ["account_profile"] data undefined in webview

```shell
Claude Desktop Diagnostics
       Desktop: KDE
[PASS] Electron: v41.5.0
[PASS] MCP config: valid JSON (~/.config/Claude/claude_desktop_config.json)
       MCP servers configured: 1
[PASS] Node.js: v22.22.0
[PASS] Disk space: 181201MB free
Cowork Mode
----------------
[PASS] bubblewrap: found
[PASS] KVM: accessible
[PASS] vsock: module loaded
       QEMU: not found
[PASS] socat: found
       virtiofsd: not found
All checks passed.
```

### What happened

Session is never persisted across launches. Every relaunch requires a fresh Google OAuth login.

Environment:
- OS: Fedora Linux 44 (KDE Plasma Desktop Edition)
- Kernel: 6.19.14-300.fc44.x86_64
- GPU: Intel Iris Xe Graphics (TigerLake-LP GT2) — integrated
- Package: RPM via DNF repo
- Claude Desktop: 1.6259.1 / Wrapper: 2.1.128

Secondary issue: default launch renders a black screen. Hardware rendering fails on Intel Iris Xe with this kernel. Workaround: MESA_LOADER_DRIVER_OVERRIDE=softpipe. App renders correctly with this flag.

With rendering working, Google OAuth completes and the app reaches claude.ai/new. But session never persists.

Root cause identified:
1. ~/.config/Claude/Cookies is created at 0 bytes on first launch and never written to. Confirmed via lsof — no file descriptor is ever opened on the Cookies database during an active logged-in session.
2. claude.ai-web.log shows this error on every launch after a successful OAuth callback: ["account_profile"] data is undefined
3. main.log shows: safeStorage not available, tokens will not persist

The account_profile IPC bridge failure means the webview never receives authenticated account state from the native layer, so cookies are never written and every session starts unauthenticated.

### Steps to reproduce

1. Install via DNF repo on Fedora 44 KDE
2. Launch with MESA_LOADER_DRIVER_OVERRIDE=softpipe to work around black screen
3. Complete Google OAuth login — app reaches claude.ai/new and appears functional
4. Close app via system tray
5. Relaunch — login screen is shown again, session not persisted
6. Check: ls -la ~/.config/Claude/Cookies → file is 0 bytes
7. Check: lsof on electron process → no file descriptor open on Cookies database

### Expected behavior

Expected: session persists across launches, no repeated login required.
Got: Cookies file remains 0 bytes, account_profile IPC bridge fails on every launch, full re-authentication required every time.

### Logs / errors

```shell
From claude.ai-web.log (repeats every launch):
[REACT_QUERY_CLIENT] QueryClient error: {"name":"Error","message":"[\"account_profile\"] data is undefined"}

From main.log (repeats every launch):
[warn] safeStorage not available, tokens will not persist
[warn] safeStorage not available, cannot decrypt tokens
[error] Electron safeStorage encryption is not available on this system

From ~/.cache/claude-desktop-debian/launcher.log:
KMS: DRM_IOCTL_MODE_CREATE_DUMB failed: Permission denied
/usr/bin/xdg-mime: line 885: qtpaths: command not found
```

### Anything else

Workarounds attempted — none resolved cookie persistence:
- --password-store=basic: safeStorage warnings gone but Cookies still 0 bytes
- --password-store=kwallet6: kwalletd6 confirmed running via DBus, no effect
- --disable-features=PartitionedCookies: no effect
- LIBGL_ALWAYS_SOFTWARE=1 vs MESA_LOADER_DRIVER_OVERRIDE=softpipe: rendering method does not affect cookie behavior
- XDG_CURRENT_DESKTOP=KDE explicitly set: already correct, no effect

The ["account_profile"] data is undefined error matches the pattern in anthropics/claude-code#56234 (macOS blank screen) — same IPC bridge failure, different platform.

qtpaths not found despite qt6-qttools installed — binary is qtpaths6 not qtpaths. Symlink workaround applied, did not affect session persistence.

Orphan process accumulation also present — closing window does not terminate process tree (related to #321).

dubreal:
Fixes #593

Two bugs affecting Fedora 44 KDE with Intel Iris Xe GPU:

1. Black screen (Intel TigerLake-LP GT2, kernel 6.x) Add apply_softpipe_if_drm_blocked() — detects TigerLake-LP GT2 via sysfs/lspci and sets MESA_LOADER_DRIVER_OVERRIDE=softpipe before the GPU process starts. Covers all seven TigerLake-LP GT2 PCI IDs. Does not clobber an existing user-set value.

2. Session not persisting (cookies always 0 bytes) Add _detect_password_store() to probe D-Bus for the best available keyring (kwallet6 -> gnome-libsecret -> basic) and inject --password-store=VALUE into electron_args before the app path so Chromium receives it as a Chromium flag, not a renderer argument.

---

Two Fedora 44 KDE-specific bugs fixed in `scripts/launcher-common.sh`. Both were diagnosed from scratch on a clean Fedora 44 KDE install with Intel Iris Xe and confirmed working.

---

## Bug 1 — Black screen on Intel Iris Xe (TigerLake-LP GT2, kernel 6.x)

**Root cause:** Mesa's iris driver calls a KMS DRM ioctl that is denied on Fedora 44 kernel 6.x, causing the GPU process to fail before anything renders.

**Fix:** Added `apply_softpipe_if_drm_blocked()` — detects TigerLake-LP GT2 hardware via three-tier sysfs/lspci detection and sets `MESA_LOADER_DRIVER_OVERRIDE=softpipe` to force Mesa's software rasterizer. Covers all seven TigerLake-LP GT2 PCI IDs (`0x9a40`, `0x9a49`, `0x9a59`, `0x9a60`, `0x9a68`, `0x9a70`, `0x9a78`). Respects any existing user-set value — no clobber. Works on first launch with no log file dependency.

---

## Bug 2 — Session not persisting (cookies always 0 bytes)

**Root cause:** `--password-store` was being passed *after* `app.asar` in the Electron exec call, so Electron received it as a renderer argument instead of a Chromium flag. `safeStorage` reported unavailable, the IPC bridge crashed, and cookies were never written to disk.

**Fix:** Added `_detect_password_store()` which probes D-Bus in order (kwallet6 → gnome-libsecret → basic) and `build_electron_args()` which injects `--password-store=VALUE` into electron args *before* the app path. `basic` is the safe fallback — no keyring required, protected by Linux file permissions.

---

## Tested on

| | |
|---|---|
| OS | Fedora Linux 44 (KDE Plasma) |
| Kernel | 6.19.14-300.fc44.x86_64 |
| GPU | Intel Iris Xe (TigerLake-LP GT2, `0x9a49`) |
| Claude Desktop | 1.6259.1 |
| Wrapper | 2.1.128 |
| Electron | v41.5.0 |

Patched source also available at [dubreal/claude-desktop-fedora](https://github.com/dubreal/claude-desktop-fedora) for anyone who needs it before this merges.

### DIFF
--- a/scripts/launcher-common.sh
+++ b/scripts/launcher-common.sh
@@ -100,6 +100,104 @@ _resolve_titlebar_style() {
 	esac
 }
 
+# Apply MESA_LOADER_DRIVER_OVERRIDE=softpipe on Intel TigerLake / Iris Xe.
+#
+# On Fedora 44 (kernel 6.x) with Intel TigerLake / Iris Xe, Chromium's GPU
+# process issues DRM_IOCTL_MODE_CREATE_DUMB (a KMS ioctl for framebuffer
+# allocation) via Mesa's iris driver. That ioctl is denied under Fedora's
+# default security policy — Mesa cannot initialise and the main window
+# renders black. MESA_LOADER_DRIVER_OVERRIDE=softpipe forces Mesa's
+# pure-software rasteriser, bypassing DRM/KMS entirely.
+#
+# Detection is hardware-level (not log-based) so the override is active
+# on the very first launch of a clean install — before any log evidence
+# exists. An existing MESA_LOADER_DRIVER_OVERRIDE value is never clobbered.
+#
+# TigerLake-LP GT2 device IDs: 0x9a40 0x9a49 0x9a59 0x9a60 0x9a68 0x9a70
+# 0x9a78. Detection order:
+#   1. /sys/class/drm/card*/device/{vendor,device} (no extra tools needed)
+#   2. lspci -nn | grep -qi "iris.xe"
+#   3. lspci -nn | grep -qiE "8086:(9a40|9a49|...)"
+apply_softpipe_if_drm_blocked() {
+	# Honour an explicit user override; do not clobber it.
+	[[ -n ${MESA_LOADER_DRIVER_OVERRIDE:-} ]] && return 0
+
+	local card_dev vendor device found=0
+
+	for card_dev in /sys/class/drm/card*/device; do
+		[[ -d $card_dev ]] || continue
+		vendor=$(cat "$card_dev/vendor" 2>/dev/null) || continue
+		[[ $vendor == '0x8086' ]] || continue
+		device=$(cat "$card_dev/device" 2>/dev/null) || continue
+		case $device in
+			0x9a40|0x9a49|0x9a59|0x9a60|0x9a68|0x9a70|0x9a78)
+				found=1; break ;;
+		esac
+	done
+
+	if [[ $found -eq 0 ]] && command -v lspci &>/dev/null; then
+		if lspci -nn 2>/dev/null | grep -qi 'iris.xe'; then
+			found=1
+		elif lspci -nn 2>/dev/null | grep -qiE \
+			'8086:(9a40|9a49|9a59|9a60|9a68|9a70|9a78)'; then
+			found=1
+		fi
+	fi
+
+	if [[ $found -eq 1 ]]; then
+		export MESA_LOADER_DRIVER_OVERRIDE=softpipe
+		log_message \
+			'Intel TigerLake/Iris Xe — MESA_LOADER_DRIVER_OVERRIDE=softpipe'
+	fi
+}
+
+# Determine the best available Chromium password-store backend.
+#
+# Electron's safeStorage API and Chromium's cookie encryption both rely
+# on the OS credential store selected by --password-store. Without a
+# working store safeStorage.isEncryptionAvailable() returns false, OAuth
+# tokens are silently discarded on exit, and users must re-authenticate
+# on every launch (Cookies file stays 0 bytes). Fixes: #593
+#
+# Detection order (first match wins):
+#   CLAUDE_PASSWORD_STORE env var  — explicit user override
+#   kwallet6                        — KDE Plasma 6 keyring
+#   gnome-libsecret                 — GNOME Keyring / libsecret bridge
+#   basic                           — fixed internal key (always works)
+#
+# With 'basic' the stored data is encrypted with a fixed key. Tokens
+# remain protected by Linux filesystem permissions on ~/.config/Claude/.
+_detect_password_store() {
+	if [[ -n ${CLAUDE_PASSWORD_STORE:-} ]]; then
+		echo "$CLAUDE_PASSWORD_STORE"
+		return
+	fi
+
+	# kwallet6: KDE Plasma 6 keyring
+	if dbus-send --session --print-reply \
+		--dest=org.kde.kwalletd6 \
+		/modules/kwalletd6 \
+		org.kde.KWallet.isEnabled 2>/dev/null \
+		| grep -q 'boolean true'
+	then
+		echo 'kwallet6'
+		return
+	fi
+
+	# gnome-libsecret: GNOME Keyring, KWallet 5 compat bridge, etc.
+	if dbus-send --session \
+		--dest=org.freedesktop.secrets \
+		/org/freedesktop/secrets \
+		org.freedesktop.DBus.Peer.Ping 2>/dev/null
+	then
+		echo 'gnome-libsecret'
+		return
+	fi
+
+	# No keyring accessible — fall back to fixed-key provider.
+	echo 'basic'
+}
+
 # Build Electron arguments array based on display backend
 # Requires: is_wayland, use_x11_on_wayland to be set
 #           (call detect_display_backend first)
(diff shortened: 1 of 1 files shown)


aaddrick (changes requested):
Thanks for the detailed write-up @dubreal. The diagnostic work in #593 is solid — lsof on the Cookies fd, bypassing the launcher to confirm `--password-store=kwallet6` fixes safeStorage, the three log warnings. Black-screen-on-launch is real user pain and I want a durable fix landed. The password-store half is well-targeted. The rendering half I'd like more evidence on before it goes in.

`--password-store` looks right. `_detect_password_store` and the injection at `scripts/launcher-common.sh:240` are correct. Electron has to receive the flag before the app path, and the launcher on main never emitted it at all. Small framing nit: the PR body says the flag "was being passed *after* `app.asar`", but on main it's not passed anywhere. Your follow-up in #593 already corrects this. Two asks: add `CLAUDE_PASSWORD_STORE` to `log_session_env` at `scripts/launcher-common.sh:35-47` so #548-style reports include the chosen backend, and surface the resolved store in `run_doctor` next to the `safeStorage` line. When this regresses on a new DE, doctor output is where the next reporter will look. `dbus-send` is universal enough that I'm not worried about the missing `command -v` guard, but a one-line comment noting the function assumes a session bus would help. Offer: this half is ~60 lines, zero shared state with the rendering half, and could land same-day as a focused split PR. Happy to fast-track that while the GPU question gets sorted.

`apply_softpipe_if_drm_blocked` I'd hold. Two concerns I'm confident on, one I want your help testing. The hardware match at `scripts/launcher-common.sh:127-145` is too broad. TigerLake-LP GT2 (`0x9a40/49/59/60/68/70/78`) is one of the most common iGPUs shipped in the last four years. Ubuntu 24.04 with `0x9a49` will land in software rendering with no log evidence the user has the underlying problem. That's a real regression for the median Iris Xe user. `CLAUDE_DISABLE_GPU=1` landed in 88676f4 for #583 as the project's existing pattern for "GPU init failed, fall back" — user-opt-in rather than auto-applied for a whole hardware generation. Less confident, and what I'd like to test: the `KMS: DRM_IOCTL_MODE_CREATE_DUMB failed: Permission denied` line is a symptom. EACCES on `MODE_CREATE_DUMB` on Fedora *can* be an SELinux denial, but I haven't run `ausearch -m AVC -ts recent` or `setenforce 0` on your machine either. That's the missing data on both sides. Could you grab `ausearch -m AVC -ts recent` from a failing launch, and try `LIBGL_ALWAYS_SOFTWARE=1 ./claude-desktop` as an A/B against `MESA_LOADER_DRIVER_OVERRIDE=softpipe`? If `llvmpipe` works on TigerLake + kernel 6.19 + Fedora 44 it'd be the more conventional fallback, but I can't claim that without seeing it run on your stack.

A few smaller things. The branch is based pre-88676f4, so `git merge-tree` says CLEAN, but the merged result puts `--password-store` injection next to the new `_disable_gpu` block from #583. Worth eyeballing post-rebase to make sure the ordering reads sensibly. No tests added. `log_session_env` already has bats coverage and `_detect_password_store` is straightforwardly mockable via the dbus-send return code. A small test pinning the fallback chain (`CLAUDE_PASSWORD_STORE` wins → kwallet6 → libsecret → basic) would be cheap insurance. In `scripts/packaging/rpm.sh:190` the `AutoReqProv: no` plus no explicit `Requires:` means a minimal Fedora KDE install can be missing libsecret entirely. Your `basic` fallback handles that gracefully, which is the right call.

Happy to land the password-store half on its own this week if you want to split it out. The rendering half can stay open while we get the `ausearch` and `LIBGL_ALWAYS_SOFTWARE=1` data.

---

dubreal:
Per review feedback, I've split the password-store fix into its own focused PR:

**→ #611**

That PR contains only the `--password-store` changes (detection, injection,
log_session_env, doctor output, bats tests) — no softpipe code.

For the rendering half: I still need to gather the `ausearch` and A/B test
data you requested before proposing a revised fix. Will run the investigation
on my Fedora 44 / Iris Xe machine and post findings here before touching
that code.

aaddrick:
Hey @dubreal, closing the loop on this one. The password-store half landed via your #611 as c81ca46 and closed #593, with the doctor and bats coverage on top. The copy still on this branch predates the `--print-reply` fix (6a7c898), so what remains here is a conflicting duplicate of code main has since improved. I'm closing this PR as superseded.

On the rendering half, my original block stands: `apply_softpipe_if_drm_blocked` keys on PCI IDs alone, with no check that DRM actually failed. Every working TigerLake-LP GT2 machine would drop to software rendering. Since then the project settled on evidence-gating for GPU fallback: the `CLAUDE_DISABLE_GPU=1` opt-in on main (#583) and the log-signature auto-fallback pattern in #666. One bad first launch is the accepted cost there, and an unconditional hardware match runs against that direction.

Your investigation carries forward. Your #593 notes show you already A/B tested `LIBGL_ALWAYS_SOFTWARE=1` against softpipe, and softpipe is a confirmed workaround on your hardware. I've opened #707 to capture it in `docs/troubleshooting.md` with credit to you, and filed #706 as a focused tracking issue for the black-screen bug since #593 closed with only half of it fixed. Your evidence is linked there.

If you want another run at auto-detection, target #706 with a fresh branch off main, gated on the GPU-fatal log signature or an opt-in env var, and attach the `ausearch -m AVC -ts recent` output from a failing launch. That data decides whether this is an SELinux denial or a Mesa bug, which changes the fix. One confirmation that would help either way: did llvmpipe alone avoid the black screen, or only softpipe? The diagnosis here was the hard part. Happy to review a follow-up.

---

PR #606, opened 2026-05-13
ci: add concurrency group to test-flags workflow

sabiut:
## Summary
- Adds a per-ref concurrency group to `.github/workflows/test-flags.yml`
- `cancel-in-progress: false` to match `ci.yml`'s release-safe posture, so a `workflow_call` invocation inside an in-flight CI run isn't killed when a subsequent push lands

## Why
`test-flags.yml` had no concurrency group, so direct `workflow_dispatch` invocations against the same ref would run in parallel. `tests.yml` already follows this pattern (`bats-${{ github.ref }}`); `test-flags.yml` was the only test workflow missing one.

The cancel posture (`false`) is deliberate: a reusable workflow's inner concurrency overrides the caller's. With `cancel-in-progress: true`, pushing to a branch while CI is running its `test-flags` step would cancel that step mid-flight, surfacing as a confusing CI failure. Queuing matches what `ci.yml` already does.

## Test plan
- [x] `actionlint .github/workflows/test-flags.yml` passes
- [x] All 9 `./build.sh --test-flags` cases run locally (9/9 pass)
- [ ] CI green on this PR
- [ ] Verify manual `gh workflow run "Test Build Script Flags (Reusable)"` shows the concurrency group in the run summary

### DIFF
--- a/.github/workflows/test-flags.yml
+++ b/.github/workflows/test-flags.yml
@@ -4,6 +4,12 @@ on:
   workflow_call: # Make this workflow reusable
   workflow_dispatch: # Allows manual triggering for testing
 
+concurrency:
+  group: test-flags-${{ github.ref }}
+  # Matches ci.yml: queue rather than cancel, so a reusable invocation
+  # from an in-flight CI run isn't killed mid-flight on the next push.
+  cancel-in-progress: false
+
 jobs:
   test-flags:
     runs-on: ubuntu-latest


aaddrick (approved):
Hey @sabiut! Clean change and the PR body has the reasoning right. A reusable workflow's `concurrency` evaluates in the caller's `github.ref` context, so a direct `workflow_dispatch` against branch X while `ci.yml` is mid-run on X lands in the same `test-flags-refs/heads/X` group as CI's inner invocation. Hardcoding the `test-flags-` prefix instead of `${{ github.workflow }}` sidesteps the inherited-name trap and matches the ref-keyed pattern at `ci.yml:23-25`. Release tags get a unique ref, so the tag path is fine.

The cancel posture is the part worth double-checking, and I think you got it right. With `false` at `test-flags.yml:7-12`, a stray dispatch queues behind CI's in-flight invocation. With `true`, the dispatch would have cancelled CI's `test-flags` step and surfaced as a confusing failure on the original CI run. That's what your comment at `test-flags.yml:9-10` flags. `tests.yml` uses `cancel-in-progress: true` because it's not `workflow_call`'d, so the difference is reusability-driven, not a repo-wide convention.

One thing worth flagging but not acting on. GitHub's concurrency queue is 1 running + 1 pending, so a burst of rapid pushes silently drops intermediate runs and only the latest enqueues. That's identical to the outer `ci.yml` posture and fine for release gating since each tag is a distinct ref.

actionlint clean, the comment at `test-flags.yml:9-10` documents the `false` choice for future readers, and CI is green. Good to merge from my side.

---

PR #610, opened 2026-05-14
fix(rpm): silence "File listed twice" warning on chrome-sandbox

### LINKED ISSUE #609: rpm: chrome-sandbox listed twice in %files emits build warning
#595 correctly fixes #539 — the SUID bit is now baked into the package payload, so it survives `--noscripts` and works in layered images. Verified locally on Nobara 43: the build succeeds, chrome-sandbox lands as `-rwsr-xr-x root:root` in the payload, and `%post` no longer carries the chmod.

The trade-off is that every rpmbuild now prints:

```
warning: File listed twice: /usr/lib/claude-desktop/node_modules/electron/dist/chrome-sandbox
```

The cause is that the new `%files` block in `scripts/packaging/rpm.sh` lists chrome-sandbox both explicitly and implicitly (lines 240-246):

```spec
%files
%defattr(-, root, root, 0755)
%attr(755, root, root) /usr/bin/claude-desktop
%attr(4755, root, root) /usr/lib/$package_name/node_modules/electron/dist/chrome-sandbox
/usr/lib/$package_name
/usr/share/applications/claude-desktop.desktop
/usr/share/icons/hicolor/*/apps/claude-desktop.png
```

The bare `/usr/lib/$package_name` entry pulls chrome-sandbox in via the directory walk, and the `%attr(4755, ...)` line names it again. The `%attr` wins for the mode bits (which is why SUID still works), but the duplicate is now permanent noise on every build.

This is a one-line spec tweak. A couple of shapes would work — excluding chrome-sandbox from the parent directory entry with `%exclude`, or restructuring so the explicit and implicit listings don't overlap. @JoshuaVlantis would know which shape fits best with the intent of #595.

---

JoshuaVlantis:
## Summary

Closes #609. Drops the explicit `%attr(4755, ...)` line on chrome-sandbox from `%files` and instead bakes the suid bit into the buildroot in `%install`, so the directory walk under `/usr/lib/$package_name` picks up the correct mode without `%files` claiming chrome-sandbox twice.

The warning was introduced in #595 — the fix for #539 — which added `%attr(4755, ...)` to bake suid into the rpm payload. Combined with the existing bare-dir listing, that produced:

```
warning: File listed twice: /usr/lib/claude-desktop/node_modules/electron/dist/chrome-sandbox
```

on every `rpmbuild` invocation.

## Why this shape

I tested four `%files` restructurings against a minimal spec on `fedora:42` and `rockylinux:9` to find a shape that silences the warning **and** keeps chrome-sandbox in the payload with suid:

| Shape | Warning? | In package? | Suid? |
|---|---|---|---|
| `%attr` + bare dir (current) | yes | yes | yes |
| `%attr` + `%exclude` + bare dir (bot draft on #609) | none | **no** | n/a |
| bare dir + `%exclude` + `%attr` (Slack/Spotify ordering) | none | **no** | n/a |
| `%attr` + bare dir + `%exclude` | yes | **no** | n/a |
| chmod 4755 in `%install`, drop `%attr` (this PR) | **none** | **yes** | **yes** |

On modern `rpmbuild`, `%exclude` is absolute — once a path is excluded, no later `%attr` line resurrects it. The chmod-in-`%install` shape is the only one I found that preserves the #539 fix and silences the warning.

## Regression guard

Captures `rpmbuild` output and fails the build if `File listed twice` appears, so any future overlap between `%files` and `%attr` surfaces at build time instead of becoming background noise. Complements `assert_setuid` in `tests/test-artifact-rpm.sh`, which catches the opposite regression (suid bit lost).

## Verification (Docker)

Ran `./build.sh --build rpm` inside `fedora:42` mirroring CI:

- `rpmbuild` log: no `File listed twice` warning, no other new warnings
- `rpm -qlvp claude-desktop-1.7196.0-1.x86_64.rpm | grep chrome-sandbox`:
  ```
  -rwsr-xr-x  1 root root 15248 May 14 11:49 /usr/lib/claude-desktop/.../chrome-sandbox
  ```
- `rpm -ivh --nodeps`; `ls -la` shows `-rwsr-xr-x root root` on disk

## Test plan

- [ ] CI builds rpm successfully on amd64 and arm64
- [ ] `tests/test-artifact-rpm.sh` `assert_setuid` still passes
- [ ] CI rpmbuild log no longer contains `File listed twice`

---
20% AI / 80% Human
Claude: empirical %exclude semantics testing on fedora:42 and rockylinux:9, draft patch, Docker build verification, regression-guard wiring
Human: directed the investigation, validated the shape selection, reviewed and trimmed comments

### DIFF
--- a/scripts/packaging/rpm.sh
+++ b/scripts/packaging/rpm.sh
@@ -229,6 +229,11 @@ install -Dm 644 $staging_dir/claude-desktop.desktop %{buildroot}/usr/share/appli
 # Install launcher script
 install -Dm 755 $staging_dir/claude-desktop %{buildroot}/usr/bin/claude-desktop
 
+# Set the chrome-sandbox suid bit in the buildroot so the /usr/lib
+# directory walk in %files records 4755 in the payload (preserves #539
+# without the "File listed twice" warning #609 — see %files block).
+chmod 4755 %{buildroot}/usr/lib/$package_name/node_modules/electron/dist/chrome-sandbox
+
 %post
 # Update desktop database for MIME types
 update-desktop-database /usr/share/applications &> /dev/null || true
@@ -240,7 +245,6 @@ update-desktop-database /usr/share/applications &> /dev/null || true
 %files
 %defattr(-, root, root, 0755)
 %attr(755, root, root) /usr/bin/claude-desktop
-%attr(4755, root, root) /usr/lib/$package_name/node_modules/electron/dist/chrome-sandbox
 /usr/lib/$package_name
 /usr/share/applications/claude-desktop.desktop
 /usr/share/icons/hicolor/*/apps/claude-desktop.png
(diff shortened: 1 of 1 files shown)


aaddrick (approved):
The investigation table in the PR body is really useful. The `%exclude` behavior on modern rpmbuild isn't obvious, and having the four shapes tested against fedora:42 and rockylinux:9 documented means nobody has to rediscover that. The `chmod`-in-`%install` approach is the right fix here.

One thing worth knowing about: the `chmod 4755` in `%install` works because `%define __strip /bin/true` prevents `brp-strip` from writing to chrome-sandbox. Writing to the file clears SUID per kernel security policy. The `__strip` override exists for a separate reason (Electron binaries break when stripped), so it's unlikely to get removed. But the dependency is real. A comment inline would be nice if you're inclined. Not a blocker.

I pushed a commit to update the test comment in `test-artifact-rpm.sh` to reflect the new `%install` chmod mechanism. The assertion was already correct, just the comment was stale. The follow-up commits for dead code removal and the shellcheck fix are clean.

LGTM. Thanks for the thorough iteration on this, @JoshuaVlantis.

---

PR #611, opened 2026-05-14
fix(launcher): inject --password-store before app path, probe keyring at startup

### LINKED ISSUE #593: [bug]: RPM/Fedora 44 KDE: session not persisted — Cookies file always 0 bytes, ["account_profile"] data undefined in webview

dubreal:
## Problem

On KDE Plasma and other desktops with a running keyring daemon,
`safeStorage.isEncryptionAvailable()` was returning false, causing OAuth
tokens to be silently discarded on exit. The Cookies file stayed at 0 bytes
and users had to re-authenticate on every launch.

Root cause: `--password-store` was being injected *after* `app.asar` in the
Electron exec call. Electron passes post-app-path arguments to the renderer,
not to Chromium — so the flag was silently ignored and Chromium fell back to
an unusable store.

Fixes #593 (password-store half — rendering investigation tracked separately)

## Changes

**`scripts/launcher-common.sh`**
- Add `_detect_password_store()`: probes D-Bus at startup and returns the
  best available backend (kwallet6 → gnome-libsecret → basic fallback)
- Respect `CLAUDE_PASSWORD_STORE` env var for explicit user override
- Inject `--password-store=VALUE` before `app.asar` in `build_electron_args()`
  so Chromium treats it as a browser flag
- Add `CLAUDE_PASSWORD_STORE` to `log_session_env` for diagnostic reports (#548)

**`scripts/doctor.sh`**
- Add `_doctor_check_password_store()`: surfaces the resolved backend in
  `--doctor` output between SingletonLock and MCP config checks
- Notes when `CLAUDE_PASSWORD_STORE` override is active
- Notes when falling back to `basic` (fixed-key, filesystem-permission-protected)

**`tests/launcher-common.bats` / `tests/doctor.bats`**
- 4 cases for `_detect_password_store` fallback chain
- 1 case for `_doctor_check_password_store` output
- Updated `log_session_env` tests for new key position

## Testing

All existing bats tests pass. New tests added and passing:

- `_detect_password_store`: CLAUDE_PASSWORD_STORE wins without calling dbus-send
- `_detect_password_store`: falls back to kwallet6 when kwallet6 probe succeeds
- `_detect_password_store`: falls back to gnome-libsecret when kwallet6 fails
- `_detect_password_store`: falls back to basic when both dbus-send calls fail
- `_doctor_check_password_store`: output contains backend name

Verified on Fedora 44 KDE Plasma — kwallet6 detected correctly, session
persists across restarts.

### DIFF
--- a/scripts/doctor.sh
+++ b/scripts/doctor.sh
@@ -1,3 +1,4 @@
+# shellcheck shell=bash
 #===============================================================================
 # Doctor Diagnostics
 #
@@ -555,6 +556,28 @@ _doctor_check_recent_crashes() {
 	fi
 }
 
+# Report the active Chromium password-store backend.
+#
+# Calls _detect_password_store() (defined in launcher-common.sh, which
+# sources this file) to surface what keyring Electron will use for
+# safeStorage / cookie encryption. 'basic' is valid but means tokens
+# rely on filesystem permissions alone, so we note it for visibility.
+# Never fails — basic is an intentional fallback, not an error.
+_doctor_check_password_store() {
+	local store
+	store=$(_detect_password_store)
+	_pass "Password store: $store"
+	if [[ $store == 'basic' ]]; then
+		_info \
+			'  → using fixed-key fallback;' \
+			'tokens are protected by filesystem permissions only'
+	fi
+	if [[ -n ${CLAUDE_PASSWORD_STORE:-} ]]; then
+		_info \
+			"  → overridden by CLAUDE_PASSWORD_STORE=${CLAUDE_PASSWORD_STORE}"
+	fi
+}
+
 # Run all diagnostic checks and print results
 # Arguments: $1 = electron path (optional, for package-specific checks)
 run_doctor() {
@@ -726,6 +749,9 @@ run_doctor() {
 	else
 		_pass 'SingletonLock: no lock file (OK)'
 	fi
+
+	# -- Password store --
+	_doctor_check_password_store
 
 	# -- MCP config --
 	local mcp_config="$config_dir/claude_desktop_config.json"
--- a/scripts/launcher-common.sh
+++ b/scripts/launcher-common.sh
@@ -128,7 +128,7 @@ _detect_password_store() {
 	fi
 
 	# kwallet6: KDE Plasma 6 keyring
-	if dbus-send --session --print-reply \
+	if dbus-send --session --print-reply --reply-timeout=1000 \
 		--dest=org.kde.kwalletd6 \
 		/modules/kwalletd6 \
 		org.kde.KWallet.isEnabled 2>/dev/null \
@@ -139,10 +139,10 @@ _detect_password_store() {
 	fi
 
 	# gnome-libsecret: GNOME Keyring, KWallet 5 compat bridge, etc.
-	if dbus-send --session \
+	if dbus-send --session --print-reply --reply-timeout=1000 \
 		--dest=org.freedesktop.secrets \
 		/org/freedesktop/secrets \
-		org.freedesktop.DBus.Peer.Ping 2>/dev/null
+		org.freedesktop.DBus.Peer.Ping >/dev/null 2>&1
 	then
 		echo 'gnome-libsecret'
 		return
--- a/scripts/launcher-common.sh
+++ b/scripts/launcher-common.sh
@@ -42,6 +42,7 @@ log_session_env() {
 		QT_IM_MODULE \
 		CLAUDE_USE_WAYLAND \
 		CLAUDE_TITLEBAR_STYLE \
+		CLAUDE_PASSWORD_STORE \
 		CLAUDE_GTK_IM_MODULE \
 		CLAUDE_DISABLE_GPU
 	do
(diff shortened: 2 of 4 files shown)


aaddrick (commented):
Thanks for splitting this out @dubreal — the diff is exactly what was asked for in the #599 review. Clean 181/-5 across `launcher-common.sh`, `doctor.sh`, and the two bats files, nothing leaking from the rendering investigation. The `CLAUDE_PASSWORD_STORE` log line and `_doctor_check_password_store` are both responsive to what I'd asked for last week, and the fallback-chain test cases (kwallet6 wins → secrets-ok → fail-fail → env override) are the right shape.

One probe correctness issue worth fixing before merge. The `gnome-libsecret` `dbus-send` at lines 107-113 is missing `--print-reply`, so it's fire-and-forget — exits 0 as long as the local session bus accepts the message, regardless of whether anything is actually listening at `org.freedesktop.secrets`. Confirmed locally: `dbus-send --session --dest=org.freedesktop.NONEXISTENT.nope /foo org.freedesktop.DBus.Peer.Ping` returns 0, same call with `--print-reply` returns 1 with `ServiceUnknown`. The `kwallet6` probe one block up gets this right. Net effect: `_detect_password_store` returns `gnome-libsecret` on any host with a session bus, so the `basic` branch this PR adds is unreachable in practice. Electron's safeStorage handles the no-secrets-service case internally, so this isn't a user-visible regression — just doctor accuracy and a dead code path. While you're in there, please also add `--reply-timeout=500` to both probes. `build_electron_args` runs `_detect_password_store` on every launch, and a wedged session bus would stall startup for the default 25s on the hot path.

Small notes, not blockers. `_doctor_check_password_store` emits `[PASS]` even when the resolved store is `basic`, which could mislead a reporter — `_info 'Password store: basic'` (no PASS prefix) would read more honestly, but I don't feel strongly. The bats coverage tests `_detect_password_store` in isolation but doesn't assert `build_electron_args` actually emits `--password-store=...`; a one-liner `has_electron_arg '--password-store=*'` in an existing case would close that cheaply. Happy to take both as follow-ups. Verdict: COMMENT — will merge same-day once `--print-reply` and `--reply-timeout=500` land.

---

sabiut (commented):
Reviewed from the test-CI perspective (CODEOWNER on `/tests/`). Pulled the branch and ran the BATS suite locally — 75/75 green, including the 4 new `_detect_password_store` cases and the new `_doctor_check_password_store` case. Test setup follows the existing patterns cleanly.

One ask in my lane, picking up @aaddrick's coverage note: none of the 11 `build_electron_args:` cases asserts `--password-store=*` is emitted, so a regression that dropped the new injection from `build_electron_args` itself wouldn't be caught here. `has_electron_arg` already supports glob patterns ([`tests/launcher-common.bats:11`](https://github.com/aaddrick/claude-desktop-debian/blob/main/tests/launcher-common.bats#L11)), so one line in any existing case closes it:

```bash
has_electron_arg '--password-store=*'
```

Since you're already in the file for @aaddrick's `--print-reply` and `--reply-timeout=500` fixes, would you mind folding this in the same revision? Keeps the regression guard with the feature.

aaddrick (approved):
Pushed the probe fixes from my earlier review. The gnome-libsecret probe needed `--print-reply` and both probes needed `--reply-timeout=1000`. I also added `CLAUDE_PASSWORD_STORE='basic'` in the `launcher-common.bats` setup so the `build_electron_args` tests don't fire real `dbus-send` calls against the host session bus. All 98 BATS tests pass.

The detection chain, env override, and flag injection position are correct. Solid fix for #593.

@sabiut's ask for a `has_electron_arg '--password-store=*'` assertion is still worth adding, but I didn't want to hold the merge for it. Happy to take that as a follow-up.

Approving.

---

PR #614, opened 2026-05-16
feat(doctor): flag eCryptfs short NAME_MAX + workaround doc (#590)

aaddrick:
## Summary

- Add `_doctor_check_filename_limit` to `scripts/doctor.sh` — probes `NAME_MAX` on the first existing ancestor of `~/.claude/projects/`, warns when `< 200`, and adds an eCryptfs-specific LUKS workaround hint when `df --output=fstype` reports the filesystem type. Wired into the Cowork Mode section of `run_doctor` so users hit the warning before their first session attempt rather than chasing an opaque `ENAMETOOLONG`.
- Add `Cowork: ENAMETOOLONG on encrypted home (eCryptfs)` section to `docs/TROUBLESHOOTING.md` walking through @proffalken's LUKS-volume + `pam_mount` workaround. The doctor's eCryptfs hint references this section by name.
- Credit @michelsfun (detailed `--doctor` bug report that pinpointed the short-`NAME_MAX` filesystem) and @proffalken (LUKS workaround) in README Acknowledgments.

Closes #590.

## Test plan

- [x] `shellcheck -x` clean (CI-style invocation against all linted shell scripts)
- [x] `bats tests/doctor.bats` — 28 tests pass (23 existing + 5 new covering: silent at `NAME_MAX >= 200`, warn at `< 200`, eCryptfs hint adds LUKS workaround, non-numeric `getconf` output silent, failing `getconf` silent)
- [x] Smoke-tested against real ext4 home (255, silent) and simulated eCryptfs via PATH shims (143 → warn + LUKS hint fires correctly)
- [x] CI shellcheck on push
- [ ] @michelsfun or @proffalken confirms the documented LUKS workaround unblocks cowork on a real eCryptfs setup

---
Generated with [Claude Code](https://claude.ai/code)
Co-Authored-By: Claude <claude-opus-4-7> <noreply@anthropic.com>
75% AI / 25% Human
Claude: drafted `_doctor_check_filename_limit` with eCryptfs detection, wrote 5 bats tests with `getconf`/`df` PATH shims, drafted the TROUBLESHOOTING.md section adapting @proffalken's workaround, drafted commit messages and this PR body
Human: directed scope (doctor check + doc + acks), did the initial issue triage that confirmed the eCryptfs hypothesis from michelsfun's `df -T` output, reviewed the approach and approved push

PR #615, opened 2026-05-16
docs: follow-ups from #585 review (#608)

### LINKED ISSUE #608: docs: follow-ups from #585 review — TROUBLESHOOTING.md, doctor threshold, awk comment
Follow-up to #585 (parent #583). @sabiut flagged three non-blocking items in their review — recording them here so they don't get lost.

### 1. Document `CLAUDE_DISABLE_GPU=1` in `docs/TROUBLESHOOTING.md`

The doctor warning in `scripts/doctor.sh:490` tells users to set `CLAUDE_DISABLE_GPU=1`, but `docs/TROUBLESHOOTING.md` has no entry for it. The flag should live there alongside the rest of the #583 narrative: XRDP / software-rendering context, when to reach for the env var vs the in-app Settings toggle, and what the launcher actually does when the variable is set (passes `--disable-gpu` to the Electron command line via `scripts/launcher-common.sh`).

### 2. Magic number `3` in `_doctor_check_recent_crashes`

`scripts/doctor.sh:484` fires the warning at `count >= 3` with no explanation. I tuned it against my own repro (~10 crashes over 7 days on the affected laptop), but that's not in the source.

Two options:

1. Add a one-line comment above the threshold check pointing at #583 and noting the tuning basis. Cheap and sufficient.
2. Add a `CLAUDE_DOCTOR_CRASH_THRESHOLD` env override so users who've already applied the workaround can suppress the nag. More flexible but more surface area.

I'd lean (1) unless someone reports the nag being noisy after acting on it.

### 3. `awk` total-count branch could miscount on newer systemd

`coredumpctl list electron` pre-filters by `COMM=electron`, so `-- Reboot --` separator rows shouldn't surface in the listing today. But `scripts/doctor.sh:463` (`NR>1 && NF>0`) would count them if a future systemd ever leaks them through. The path-matched branch at line 468 uses `index($0, p)` so it's unaffected.

A short comment near line 463 noting the assumption (COMM filter excludes reboot markers; revisit if systemd changes) is enough to future-proof.

---

Credit to @sabiut for the catch on all three.

---

sabiut:
## Summary

Picks up the three non-blocking items @aaddrick captured from my #585
review into [#608](https://github.com/aaddrick/claude-desktop-debian/issues/608):

- **`docs/TROUBLESHOOTING.md`** — new "Repeated Electron Crashes / GPU
  Process FATAL" section. Documents `CLAUDE_DISABLE_GPU=1`: when to
  reach for the env var vs the in-app Settings toggle, the
  `--disable-gpu --disable-software-rasterizer` translation in
  `scripts/launcher-common.sh`, the XRDP overlap, and the #583
  tracking link the `--doctor` warning points at.
- **`scripts/doctor.sh` (`_doctor_check_recent_crashes`)** — one-line
  comment above the `count >= 3` threshold explaining the tuning
  basis (#583 repro, ~10 crashes over 7 days on the affected laptop,
  3 chosen as the "worth surfacing the workaround" floor). Picked
  option (1) from the issue — no env override surface, can revisit
  if anyone reports the nag is noisy.
- **`scripts/doctor.sh` (awk total-count branch)** — comment noting
  the count relies on `coredumpctl list electron`'s `COMM=electron`
  filter to exclude `-- Reboot --` separator rows. The path-matched
  branch uses `index($0, p)` so it's already safe; the comment
  flags the assumption so future-us revisits if systemd ever leaks
  reboot markers through per-COMM listings.

Fixes #608.

### DIFF
--- a/docs/TROUBLESHOOTING.md
+++ b/docs/TROUBLESHOOTING.md
@@ -131,6 +131,45 @@ windows or rich compose-key sequences. Only reach for it if your
 real input method (IBus/Fcitx) is broken; if you depend on CJK or
 compose, prefer fixing the IBus/Fcitx integration instead.
 
+### Repeated Electron Crashes / GPU Process FATAL ([#583](https://github.com/aaddrick/claude-desktop-debian/issues/583))
+
+If Claude Desktop crashes repeatedly on launch or shortly after,
+the most common cause on Linux is the Chromium GPU process hitting
+a FATAL exhaustion path. `claude-desktop --doctor` surfaces this
+when `systemd-coredump` shows 3+ Electron crashes in the last 7
+days, pointing at this issue.
+
+Two ways to disable hardware acceleration as a workaround:
+
+1. **In-app:** Settings → toggle hardware acceleration off →
+   restart Claude Desktop. Persists in the upstream config.
+2. **Env var (headless / persists across reinstalls):** set
+   `CLAUDE_DISABLE_GPU=1` in the environment before launching.
+
+```bash
+# One-off:
+CLAUDE_DISABLE_GPU=1 claude-desktop
+
+# Persistent (shell profile):
+echo 'export CLAUDE_DISABLE_GPU=1' >> ~/.profile
+```
+
+When `CLAUDE_DISABLE_GPU=1` is set, the launcher passes
+`--disable-gpu --disable-software-rasterizer` to Electron (see
+`scripts/launcher-common.sh`). This is the same pair of flags
+applied automatically inside XRDP sessions, where software
+rendering is required regardless. Either signal is sufficient —
+the launcher won't stack duplicate flags.
+
+**When to prefer which:** the in-app toggle is friendlier if you
+can reach Settings without the app crashing. Reach for
+`CLAUDE_DISABLE_GPU=1` when the app crashes before you can open
+Settings, when running in environments with no GPU available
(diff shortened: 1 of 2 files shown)


aaddrick (approved):
Hey @sabiut! Thanks for picking these up from #608. Walked through all three changes against the current tree and they hold up.

The TROUBLESHOOTING.md section accurately describes what the launcher actually does. `CLAUDE_DISABLE_GPU=1` adds `--disable-gpu --disable-software-rasterizer` at `scripts/launcher-common.sh:161`, XRDP applies the same pair at lines 147-150, and the `_disable_gpu` boolean at 144-146 is what prevents duplicate flag stacking. So your "either signal is sufficient" line is correct. The doctor.sh threshold comment matches the #583 repro I was tuning against, and the `awk` comment correctly identifies that the path-matched branch is already safe via `index($0, p)` at doctor.sh:467-468.

All 23 bats tests pass on the PR head. Heads up that #614 landed yesterday and added 6 more eCryptfs NAME_MAX tests (29 on main now), but `git merge-tree` shows a clean auto-merge so you don't need to rebase.

One soft observation, not a request: the "When to prefer which" paragraph in the TROUBLESHOOTING addition slightly restates the numbered list above it. Readers skim, so the redundancy probably helps more than it hurts. Auto-merge is clean, going to merge after this.

PR #624, opened 2026-05-18
fix: actively quit on close when CLAUDE_QUIT_ON_CLOSE=1 (#623)

### LINKED ISSUE #623: [bug]: CLAUDE_QUIT_ON_CLOSE=1 leaves app alive — bundled main-process close handler still hides

```
Claude Desktop Diagnostics

[PASS] Installed version: 1.7196.1
       Desktop: ubuntu:GNOME
[PASS] Electron: v41.5.0
[PASS] SingletonLock: held by running process
[PASS] MCP config: valid JSON
       MCP servers configured: 0
[PASS] Node.js: v24.11.1
[PASS] Disk space: 258047MB free

Cowork Mode
----------------
[PASS] bubblewrap: found
[PASS] KVM: accessible
[PASS] vsock: module loaded
[PASS] Log file: 71KB

All checks passed.
```

Ubuntu 25.10, GNOME on Wayland (XWayland session). Reproduces with a clean `~/.config/Claude` moved aside.

### What happened

With `CLAUDE_QUIT_ON_CLOSE=1` exported (via XDG override or in a parent shell) the launcher.log correctly reports `[Frame Fix] Close-to-tray: off` and `/proc/<pid>/environ` confirms the env var is live in the running Electron process. But clicking X on the main window still does not quit the app: the process stays alive, the window is hidden, and subsequent launches just unhide the existing instance (single-instance lock).

The wrapper-side close-to-tray opt-out installed by #448 works as designed — it just doesn't fully address close=quit anymore, because the bundled main-process code in `app.asar` now registers its **own** `close` listener that also `preventDefault()` + `hide()`s the main window, with no setting or env var gating it on Linux.

### Steps to reproduce

1. Install 1.7196.1 (or any recent build).
2. Move `~/.config/Claude` aside; relaunch with `CLAUDE_QUIT_ON_CLOSE=1` in env (verify with `tr '\0' '\n' < /proc/<pid>/environ | grep CLAUDE_QUIT_ON_CLOSE`).
3. Confirm `~/.cache/claude-desktop-debian/launcher.log` shows `[Frame Fix] Close-to-tray: off`.
4. Click X on the main window.
5. `pgrep -af /usr/lib/claude-desktop/` — process is still running.
6. Click the launcher icon again — log shows `Not main instance, returning early from app ready` in `~/.config/Claude/logs/main.log`; the original (now-hidden) window just unhides.
7. **Expected**: process exits, next launch is a fresh process. **Got**: process persists.

Workarounds that do quit cleanly: `Ctrl+Q` (handled by the wrapper at `frame-fix-wrapper.js:309` and `:563`), right-click tray icon → Quit, or `pkill -f /usr/lib/claude-desktop/`.

### Expected behavior

`CLAUDE_QUIT_ON_CLOSE=1` should fully restore close-window=quit, as the README in this repo states. Today it only disables half of the close-to-tray pair.

### Logs / errors

```
~/.cache/claude-desktop-debian/launcher.log (excerpt):
--- Claude Desktop Launcher Start ---
[Frame Fix] Wrapper loaded
[Frame Fix] Menu bar mode: auto
[Frame Fix] Titlebar style: hybrid
[Frame Fix] Close-to-tray: off
[Frame Fix] Upgrade watcher armed: /usr/lib/claude-desktop/node_modules/electron/dist/resources/app.asar
[Frame Fix] Patches built successfully

# After clicking X: no "Electron exited with code" line; process keeps running.
# Subsequent launches from the icon:
--- Claude Desktop Launcher Start ---
[Frame Fix] Wrapper loaded
...
[Frame Fix] Close-to-tray: off
Electron exited with code: 0       <-- the second-instance launch exits
--- Claude Desktop Launcher End ---

~/.config/Claude/logs/main.log (excerpt from a second-instance launch):
[info] Starting app { appVersion: '1.7196.1', ... platform: 'linux' }
[info] Not main instance, returning early from app ready
```

#### Root cause

Anthropic's bundled main-process code in `app.asar` at `.vite/build/index.js` (offset ~12292119 in v1.7196.1) installs its own `close` listener on the main window:

```js
n.on("close", l => {
  if (lC()) return;                              // already quitting? short-circuit
  if (Tn && !Ei("menuBarEnabled")) {             // Tn = process.platform === "win32"
    yh(); return;                                 // app.quit()
  }
  l.preventDefault();
  const Q = () => { bLA(); n.hide(); };
  n.isFullScreen() ? (n.once("leave-full-screen", Q), n.setFullScreen(false)) : Q();
});
```

with `Tn = process.platform === "win32"` (~offset 5082738) and `lC = () => yNA || ltA || BtA` (~offset 6382460) where the flags are armed by the bundled before-quit / will-quit handlers. On Linux the conditional-quit branch can never fire, so the handler unconditionally `preventDefault()`s and `hide()`s the window.

This handler appears to have been added by Anthropic upstream after #448 was filed, likely for the same reason (keep the app alive for schedulers / tray persistence). The wrapper-side opt-out introduced in #448 only removes the wrapper's parallel hide-handler at `frame-fix-wrapper.js:336`; the bundled one still runs. Net effect: the env var is necessary but no longer sufficient.

#### Proposed fix (will follow up with a PR)

Extend the wrapper's `if (!popup)` block so that **when `CLAUDE_QUIT_ON_CLOSE=1`** it installs an *active* close listener (instead of just declining to install the existing one):

```js
if (!popup) {
  if (CLOSE_TO_TRAY) {
    this.on('close', (e) => { /* existing hide-on-close */ });
  } else {
    // Bypass upstream's hardcoded hide-on-close in `index.js` by
    // riding their own `lC()` quit-in-progress guard. Our listener
    // runs first (registered immediately after super()); app.quit()
    // emits 'before-quit' synchronously, which sets ltA=true inside
    // the bundled beforeQuit handler. Upstream's close listener
    // then runs second, hits `if (lC()) return;`, short-circuits
    // without calling preventDefault, and the window closes during
    // the quit flow.
    this.on('close', () => { result.app.quit(); });
  }
}
```

The contract this relies on is upstream's own `if (lC()) return;` short-circuit, which they need anyway for Ctrl+Q / tray Quit / SIGTERM. Robust to minifier renames as long as that guard exists.

Tradeoff vs alternatives:
- **Splice their listener out of `_events.close`**: no quit-flow side-effect, but requires `.toString()` matching that fights upstream re-minification (cf. `docs/learnings/patching-minified-js.md`).
- **Just additive `app.quit()` without ordering**: works but creates a brief hide-then-quit visual flash if upstream's handler races to `hide()` before before-quit propagates.
- **Patch `BrowserWindow.prototype.on`**: same heuristic problem as the splice approach.

The proposed approach intentionally avoids any string-matching against the minified bundle.

#### Related

- Follows #448 (closed; introduced the wrapper-side close-to-tray + env var). This issue is the natural follow-up: the env var's behaviour is now stacked against an upstream-bundled handler that didn't exist when #448 shipped.
- The README's `CLAUDE_QUIT_ON_CLOSE=1` description should also be updated to reflect actual behaviour once the wrapper fix lands.

phelps-matthew:
## Summary

When `CLAUDE_QUIT_ON_CLOSE=1` is set, install an *active* close listener that calls `app.quit()` instead of merely declining to install the wrapper's hide-on-close handler. This counters the bundled main-process close listener in `.vite/build/index.js` that hardcodes `preventDefault + hide` on non-Windows platforms — a handler Anthropic added upstream after #448 shipped, which the wrapper's existing opt-out doesn't reach.

Fixes #623. Refs #448.

## Why this approach

The bundled listener guards itself with `if (lC()) return;` where `lC()` ORs together flags armed by the bundled before-quit / will-quit handlers. `app.quit()` fires `before-quit` *synchronously*, which sets `ltA=true` inside that handler before returning.

So if we register our close listener first (right after `super()` in the patched `BrowserWindow`), the sequence on X-click is:

1. Electron emits `close`.
2. Our listener runs → `app.quit()` → before-quit fires → `ltA=true` → `lC()` truthy.
3. Bundled close listener runs second → `if (lC()) return;` → short-circuits, no `preventDefault`, no `hide()`.
4. `close` completes without preventDefault, the window closes in the quit flow, async cleanup proceeds, app exits.

No string-matching against the minified bundle. No `removeListener` splicing. The contract we rely on (`lC()` guard) is upstream's own quit-safety mechanism; they need it for Ctrl+Q / tray Quit / SIGTERM regardless, so it's not going away.

Alternatives considered:

| Approach | Robust to upstream re-minification | Visual flash | Touches their listener |
|---|---|---|---|
| Splice `_events.close` | No (`.toString()` heuristic) | None | Yes (fragile) |
| Additive `app.quit()` without ordering | Yes | Brief hide-then-quit | No |
| **Prepended listener riding `lC()`** | **Yes** | **None** | **No** |
| Patch `BrowserWindow.prototype.on` | No (same heuristic) | None | Yes (fragile) |

## Test plan

- [x] Manually verified on Ubuntu 25.10 (GNOME, Wayland/XWayland), 1.7196.1 base, clean `~/.config/Claude`. Before this patch: clicking X with `CLAUDE_QUIT_ON_CLOSE=1` set leaves the process alive (`pgrep -af /usr/lib/claude-desktop/` still shows the main process; relaunch from icon logs `Not main instance, returning early`). After: clicking X exits the main process; relaunch is a fresh start.
- [x] `[Frame Fix] Close-to-tray: off` still logged in launcher.log (env-var path unchanged for that message).
- [x] Without the env var (default tray behaviour): `CLOSE_TO_TRAY` path unchanged, X still hides — verified.
- [ ] CI lint / `./build.sh --build appimage` — I have not run a full local appimage build for this change because it's an in-place edit to `frame-fix-wrapper.js` (not a `scripts/patches/*.sh` patch) and the existing CI should exercise the wrapper. Happy to run a local build if you'd like.

## Scope

One file (`scripts/frame-fix-wrapper.js`), +26 / -0. Behaviour is gated entirely on the existing `CLAUDE_QUIT_ON_CLOSE=1` env var — default behaviour (close-to-tray) is unchanged.

A follow-up doc tweak to the README's `CLAUDE_QUIT_ON_CLOSE` description seems warranted once this lands (today the README implies it fully restores Electron-default close=quit); happy to do that as a separate PR if you'd prefer not to bundle.

### DIFF
--- a/scripts/frame-fix-wrapper.js
+++ b/scripts/frame-fix-wrapper.js
@@ -339,6 +339,32 @@ Module.prototype.require = function(id) {
                     this.hide();
                   }
                 });
+              } else {
+                // CLAUDE_QUIT_ON_CLOSE=1: the bundled main-process code
+                // (`.vite/build/index.js`) installs its own main-window
+                // close listener that hardcodes `preventDefault()` +
+                // `hide()` on every non-Windows platform, with no
+                // setting or env var to disable it. The wrapper's
+                // opt-out above only removes *this* file's hide handler;
+                // the bundled one still runs, so without this branch
+                // closing the window still leaves the app alive in the
+                // tray (in-app schedulers / single-instance lock /
+                // deleted-inode electron after dpkg upgrade-in-place).
+                //
+                // Approach: register a close listener that runs *first*
+                // and calls app.quit(). app.quit() emits 'before-quit'
+                // synchronously, which sets the bundled code's
+                // "quitting in progress" flag. The bundled close
+                // listener then runs second, sees that flag, and
+                // short-circuits via its own `if (lC()) return;` guard
+                // — so it never calls preventDefault, and the window
+                // closes normally during the quit flow. We ride the
+                // upstream's own quit-safety contract instead of trying
+                // to remove or splice their listener; robust to any
+                // refactor that preserves the quit-in-progress short-
+                // circuit (which they need for Ctrl+Q / tray Quit /
+                // SIGTERM anyway). Fixes: #623
+                this.on('close', () => { result.app.quit(); });
               }
 
               // Directly set child view bounds to match content size.


aaddrick (approved):
This is solid work. The root cause analysis in #623 is thorough. The bundled close handler hardcodes `preventDefault + hide` on Linux with no env var gate, and the wrapper's existing opt-out only covered half the picture.

The approach makes sense. Riding upstream's own `lC()` quit-in-progress guard avoids the fragility of string-matching against minified code or splicing `_events.close`. That contract is durable because they need that guard for their own Ctrl+Q / tray Quit / SIGTERM paths. EventEmitter FIFO ordering is guaranteed here since your handler registers inside the constructor body before `new` returns to the bundled code, and `close` must support `preventDefault()` which keeps it in JS-land dispatch.

Two minor things, non-blocking: a `console.log('[Frame Fix] Active quit-on-close triggered')` when the handler fires would help distinguish "handler never triggered" from "handler triggered but quit still failed." The install-time log is already inferrable from `Close-to-tray: off`, so this would fill the gap. Also, the double-quit comment could mention the upgrade-watcher restart path alongside Ctrl+Q and tray Quit for completeness.

Clean fix, well-scoped. Thanks @phelps-matthew.

---

PR #627, opened 2026-05-21
fix(tray): support $-containing identifiers in 1.8089.1 minified bundle (#625)

### LINKED ISSUE #625: [bug]: nix build failure on decb512

```shell
N/A
```

### What happened

On decb512, the nix flake failed to build in NixOS x86-64.

### Steps to reproduce

`$ nix build github:aaddrick/claude-desktop-debian/decb512144eb7f91a11bd1d5099eea733b615b9c`

### Expected behavior

The build should succeed and output a `result` symlink in the working directory.

### Logs / errors

```shell
Running phase: patchPhase
@nix { "action": "setPhase", "phase": "patchPhase" }
Running phase: updateAutotoolsGnuConfigScriptsPhase
@nix { "action": "setPhase", "phase": "updateAutotoolsGnuConfigScriptsPhase" }
Running phase: configurePhase
@nix { "action": "setPhase", "phase": "configurePhase" }
no configure script, doing nothing
Running phase: buildPhase
@nix { "action": "setPhase", "phase": "buildPhase" }
[1;36m--- Architecture Detection ---[0m
Detecting system architecture...
Detected machine architecture: x86_64
Configured for amd64 (x86_64) build.
Target Architecture: amd64
[1;36m--- End Architecture Detection ---[0m
[1;36m--- Distribution Detection ---[0m
Detecting Linux distribution family...
Warning: Could not detect distribution family
  AppImage build will still work, but native packages (deb/rpm) may not
Distribution: 
Distribution family: unknown
[1;36m--- End Distribution Detection ---[0m
Running as user: nixbld (Home: /build)
System Information:
Distribution: 
Distribution family: unknown
Target Architecture: amd64
[1;36m--- Argument Parsing ---[0m
Selected build format: nix
Cleanup intermediate files: no
[1;36m--- End Argument Parsing ---[0m
[1;36m--- Download the latest Claude executable ---[0m
Using local Claude installer: /build/Claude-Setup.exe
Local installer copied to build directory
Extracting resources from Claude-Setup-x64.exe into separate directory...

7-Zip [64] 17.05 : Copyright (c) 1999-2021 Igor Pavlov : 2017-08-28
p7zip Version 17.05 (locale=C,Utf16=off,HugeFiles=on,64 bits,12 CPUs x64)

Scanning the drive for archives:
  0M Scan /build/build/                       1 file, 192592544 bytes (184 MiB)

Extracting archive: /build/build/Claude-Setup-x64.exe
  0% Open         --
Path = /build/build/Claude-Setup-x64.exe
Type = PE
Physical Size = 192592544
CPU = x86
Characteristics = Executable 32-bit
Created = 2020-09-27 18:20:07
Headers Size = 1024
Checksum = 192605448
Image Size = 192598016
Section Alignment = 4096
File Alignment = 512
Code Size = 119808
Initialized Data Size = 192461312
Uninitialized Data Size = 0
Linker Version = 14.16
OS Version = 6.0
Image Version = 0.0
Subsystem Version = 6.0
Subsystem = Windows GUI
DLL Characteristics = Relocated NX-Compatible TerminalServerAware
Stack Reserve = 1048576
Stack Commit = 4096
Heap Reserve = 1048576
Heap Commit = 4096
Image Base = 4194304
Comment = FileVersion: 1.8089.1.0
ProductVersion: 1.8089.1.0
----
Path = .rsrc/DATA/131
Size = 192303426
Packed Size = 192303426
--
Path = .rsrc/DATA/131
Type = zip
Physical Size = 192303426

 97%     45% - AnthropicClaude-1.8089.1-full.nupkg                                           83% - AnthropicClaude-1.8089.1-full.nupkg                                          Everything is Ok

Files: 5
Size:       193873588
Compressed: 192592544
Found nupkg: ./AnthropicClaude-1.8089.1-full.nupkg (in /build/build/claude-extract)
Detected Claude version: 1.8089.1

7-Zip [64] 17.05 : Copyright (c) 1999-2021 Igor Pavlov : 2017-08-28
p7zip Version 17.05 (locale=C,Utf16=off,HugeFiles=on,64 bits,12 CPUs x64)

Scanning the drive for archives:
  0M Scan ./            1 file, 191857691 bytes (183 MiB)

Extracting archive: ./AnthropicClaude-1.8089.1-full.nupkg
--
Path = ./AnthropicClaude-1.8089.1-full.nupkg
Type = zip
Physical Size = 191857691

  0%      7% 5 - lib/net45/claude.exe                              13% 5 - lib/net45/claude.exe                              20% 5 - lib/net45/claude.exe                              28% 5 - lib/net45/claude.exe                              36% 5 - lib/net45/claude.exe                              45% 5 - lib/net45/claude.exe                              53% 8 - lib/net45/dxcompiler.dll                                  62% 31        74% 73 - lib/net45/resources/app.asar                                       83% 517         93% 882        Everything is Ok

Folders: 42
Files: 979
Size:       462296127
Compressed: 191857691
Resources extracted from nupkg
Processing app.asar...
Creating BrowserWindow frame fix wrapper...
Original main entry: .vite/build/index.pre.js
Modifying package.json to load frame fix and add node-pty...
Updated package.json: main entry, desktopName, and node-pty dependency
Creating stub native module...
Extracting electron module variable name...
  Found electron variable: cA
##############################################################
Fixing incorrect nativeTheme variable references...
  All nativeTheme references are correct
##############################################################
Patching tray menu handler...
Failed to extract tray menu function name
```

### Anything else

This may be related to the CI failures in the aforementioned commit.
[nix_log_:nix:store:bcbqzaisg3s8p3ywlxbz6vaf5s98zwqz-claude-desktop-1.8089.1.drv.log](https://github.com/user-attachments/files/28037897/nix_log_.nix.store.bcbqzaisg3s8p3ywlxbz6vaf5s98zwqz-claude-desktop-1.8089.1.drv.log)

typedrat:
## Summary

Three small fixes in `scripts/patches/tray.sh` that together unblock the 1.8089.1 build (#625) and restore the duplicate-icon fast-path from #515 on the new bundle.

## Root cause

The 1.8089.1 minified bundle emits the tray menu handler as `i$A` — an identifier containing `$`. `patch_tray_menu_handler` uses PCRE `\w+` to capture it; `\w` is `[A-Za-z0-9_]` and does **not** match `$`. `tray_func` comes back empty, the guard fires, and the entire `build.sh` aborts with `exit 1` — taking every downstream patch with it.

Once `tray_func=i$A` is captured correctly, a **second** latent bug surfaces: the unescaped `$` is interpolated raw into PCRE and `sed -E` patterns where it acts as an end-of-line anchor (`function i$A` reads as "function i", end of line, "A"), so `tray_var` extraction also returns empty.

And a **third** issue, surfaced only after the first two are fixed: Anthropic now ships the menu handler as `async function i$A()` upstream in 1.8089.1. `patch_tray_menu_handler`'s `s/function NAME/async function NAME/` then produces `async async function`, which breaks `patch_tray_inplace_update`'s `(?:async )?function NAME` anchor and silently loses the #515 duplicate-icon fix on this version.

## Fix

1. **Allow `$` in extracted JS identifiers** (`tray.sh:16,104`). `\w+` → `[\w$]+`, mirroring the `\$?\w+` pattern `_common.sh` already uses for `electron_var`.

2. **Escape `$` when interpolating `tray_func` into PCRE / `sed -E` patterns** (`tray.sh:25,38,72,111`). Mirror the existing `tray_var_re="\${local_tray_var//\\$/\\\\$}"` trick at line 120. BRE seds (lines 34, 49, 57) are safe with raw `$` because BRE treats `$` as an anchor only at end-of-pattern.

3. **Make the `s/function/async function/` sed idempotent** (`tray.sh:38`). Guard with `grep -q 'async function NAME'` so it no-ops when upstream already ships the function as async.

The change is a strict superset of old behavior: `[\w$]+` matches everything `\w+` did, `tray_func_re` is identical to `tray_func` when no `$` is present, and the new guard only suppresses a redundant rewrite. Older bundles continue to patch the same way.

## Test plan

Verified end-to-end against a freshly-downloaded 1.8089.1 bundle (`Claude-b98a06c70e376344e3b6938d6980242e8cc20547.exe`, sha matches upstream's `scripts/setup/detect-host.sh`):

```
patch_tray_menu_handler:
  Found tray function: i$A
  Found tray variable: Ju
  Found first const variable: e
  Added mutex guard to i$A()
  Added DBus cleanup delay after Ju.destroy()
  Added startup delay check (3 second window)
patch_tray_icon_selection:
  Patched tray icon selection for Linux theme support
patch_tray_inplace_update:
  Found tray variable: Ju
  Found menu function: QGe
  Found icon-path var: t
  Found menuBarEnabled var: e
  [OK] Fast-path injected before destroy-recreate
patch_menu_bar_default:
  Patched menuBarEnabled to default to true
```

- `node --check` on the patched bundle: parses cleanly (no syntax errors introduced)
- Post-patch grep confirms every marker: `_trayStartTime` (×1), `i$A._running` (×1), `setTimeout(r,250)` (×1), `Ju.setImage(cA.nativeImage.createFromPath(t))` (×1, the #515 fast-path), `e!==false` (×1), `TrayIconTemplate-Dark.png` (×1)
- Double-async check: `grep -c 'async async function'` → 0

`shellcheck -x scripts/patches/tray.sh` reports no new findings vs. main (the four pre-existing SC2148/SC2154 warnings about sourced globals are unchanged).

Fixes #625.

---

### DIFF
--- a/scripts/patches/tray.sh
+++ b/scripts/patches/tray.sh
@@ -11,9 +11,9 @@ patch_tray_menu_handler() {
 	echo 'Patching tray menu handler...'
 	local index_js='app.asar.contents/.vite/build/index.js'
 
-	local tray_func tray_var first_const
+	local tray_func tray_func_re tray_var first_const
 	tray_func=$(grep -oP \
-		'on\("menuBarEnabled",\(\)=>\{\K\w+(?=\(\)\})' "$index_js")
+		'on\("menuBarEnabled",\(\)=>\{\K[\w$]+(?=\(\)\})' "$index_js")
 	if [[ -z $tray_func ]]; then
 		echo 'Failed to extract tray menu function name' >&2
 		cd "$project_root" || exit 1
@@ -21,8 +21,12 @@ patch_tray_menu_handler() {
 	fi
 	echo "  Found tray function: $tray_func"
 
+	# Escape `$` for PCRE / sed -E patterns where it would otherwise act
+	# as an end-of-line anchor. Minifier emits identifiers like `i$A`.
+	tray_func_re="${tray_func//\$/\\$}"
+
 	tray_var=$(grep -oP \
-		"\}\);let \K\w+(?==null;(?:async )?function ${tray_func})" \
+		"\}\);let \K\w+(?==null;(?:async )?function ${tray_func_re})" \
 		"$index_js")
 	if [[ -z $tray_var ]]; then
 		echo 'Failed to extract tray variable name' >&2
@@ -31,11 +35,17 @@ patch_tray_menu_handler() {
 	fi
 	echo "  Found tray variable: $tray_var"
 
-	sed -i "s/function ${tray_func}(){/async function ${tray_func}(){/g" \
-		"$index_js"
+	# Idempotent: upstream may already ship the function as `async`
+	# (1.8089.1 does). Re-applying the sed would produce
+	# `async async function`, which then breaks downstream patches that
+	# match `(?:async )?function NAME`.
+	if ! grep -q "async function ${tray_func}(){" "$index_js"; then
+		sed -i "s/function ${tray_func}(){/async function ${tray_func}(){/g" \
+			"$index_js"
+	fi
 
 	first_const=$(grep -oP \
-		"async function ${tray_func}\(\)\{.*?const \K\w+(?==)" \
+		"async function ${tray_func_re}\(\)\{.*?const \K\w+(?==)" \
 		"$index_js" | head -1)
 	if [[ -z $first_const ]]; then
 		echo 'Failed to extract first const in function' >&2
@@ -69,7 +79,7 @@ patch_tray_menu_handler() {
 			"s/(${electron_var_re}\.nativeTheme\.on\(\s*\"updated\"\s*,\s*\(\)\s*=>\s*\{)/let _trayStartTime=Date.now();\1/g" \
 			"$index_js"
 		sed -i -E \
-			"s/\((\w+\([^)]*\))\s*,\s*${tray_func}\(\)\s*,/(\1,Date.now()-_trayStartTime>3e3\&\&${tray_func}(),/g" \
+			"s/\((\w+\([^)]*\))\s*,\s*${tray_func_re}\(\)\s*,/(\1,Date.now()-_trayStartTime>3e3\&\&${tray_func}(),/g" \
 			"$index_js"
 		echo '  Added startup delay check (3 second window)'
 	fi
@@ -98,17 +108,19 @@ patch_tray_inplace_update() {
 
 	# Re-extract the tray variable name — `patch_tray_menu_handler`
 	# declares it `local` so it's not visible here. Same grep pattern.
-	local tray_func local_tray_var tray_var_re
+	local tray_func tray_func_re local_tray_var tray_var_re
 	local menu_func path_var enabled_var enabled_count
 	tray_func=$(grep -oP \
-		'on\("menuBarEnabled",\(\)=>\{\K\w+(?=\(\)\})' "$index_js")
+		'on\("menuBarEnabled",\(\)=>\{\K[\w$]+(?=\(\)\})' "$index_js")
 	if [[ -z $tray_func ]]; then
 		echo '  Could not find tray function — skipping'
 		echo '##############################################################'
 		return
 	fi
+	# Escape `$` for PCRE patterns; matches the `tray_var_re` trick below.
+	tray_func_re="${tray_func//\$/\\$}"
 	local_tray_var=$(grep -oP \
-		"\}\);let \K\w+(?==null;(?:async )?function ${tray_func})" \
+		"\}\);let \K\w+(?==null;(?:async )?function ${tray_func_re})" \
 		"$index_js")
 	if [[ -z $local_tray_var ]]; then
 		echo '  Could not extract tray variable name — skipping'


aaddrick:
Poking around. Get back to you momentarily

aaddrick (approved):
Hey, this is solid work. Three independent issues, three minimal fixes, all verified end-to-end against the 1.8089.1 bundle. I walked through each claim against the real file and the diff and it holds up. Approving with a couple of non-blocking observations.

## What I checked

**Claim 1 — `\w+` vs `[\w$]+`.** Reproduced locally: `grep -oP 'on\("menuBarEnabled",\(\)=>\{\K\w+(?=\(\)\})'` against `on("menuBarEnabled",()=>{i$A()}` returns empty; `[\w$]+` returns `i$A`. PCRE `\w` is `[A-Za-z0-9_]`, no `$`. Mirroring `_common.sh`'s `\$?\w+` is the right reference.

**Claim 2 — escaping `$` for PCRE and `sed -E`.** Tested `${tray_func//\$/\\$}` (yields `i\$A`) on both sides. PCRE `grep -oP "\}\);let \K\w+(?==null;(?:async )?function i\$A)"` matches; raw `i$A` doesn't because `$` is end-of-string in default mode. `sed -E "s/...,\s*i\$A\(\)\s*,/.../g"` matches; raw `i$A` doesn't, same reason. Replacement side keeps unescaped `${tray_func}` because `sed`'s replacement language has no `$` metacharacter. Correct.

**Claim 3 — BRE seds at lines 43, 59, 67.** Verified safe with raw `$`. POSIX BRE treats `$` as end-anchor only at end-of-expression (or before `\)` / `\|`); mid-pattern `$` is literal. So the asymmetric "escape for `-E`, not for plain `s`" choice is defensible.

**Claim 4 — idempotency guard.** `grep -q "async function ${tray_func}(){"` under GNU grep BRE: `(`, `)`, `{` are literal and mid-expression `$` is literal, so it matches `async function i$A(){` exactly. The leading `async ` plus literal `(){` is specific enough that neighboring functions can't false-match. Re-run on already-async is a no-op; re-run on not-yet-async still rewrites. Both directions verified.

**Strict-superset for older bundles.** `[\w$]+` is a strict superset of `\w+`, and `tray_func_re == tray_func` when `$` isn't present. The new guard only suppresses a redundant rewrite. Older bundles patch identically. No regression risk.

**CI.** Every material job is green: deb (both arches), appimage amd64, rpm (both arches), BATS, shellcheck, spell, flag parsing, all three artifact validators. The one red job, `Build Packages (arm64 - appimage)`, is failing in `Install FUSE for AppImageTool (Ubuntu)` on `apt-get install -y libfuse2` — a runner-side infrastructure flake, not anything this PR touches. Re-running now.

## Observations

**1. Same bug class lurks in the other identifier extractors (not blocking).** This PR fixes `tray_func`'s capture, but the other `grep -oP …\K\w+…` captures in this file still use bare `\w+`:

- `tray_var` / `local_tray_var` at lines 29, 123
- `first_const` at line 48
- `menu_func` at line 134
- `path_var` at line 149
- `enabled_var` at line 171

For 1.8089.1 they all come out as `Ju` / `e` / `QGe` / `t`, no `$`. But minifier identifier choice isn't deterministic across releases, and the next time it picks a `$`-form for any of these you'll see the exact failure mode this PR is fixing — silently this time, because these extractors `echo` and `return` rather than `exit`, so `build.sh` keeps going with downstream patches broken.

The same concern hits a few other extractors elsewhere: `quick-window.sh:17` (`quick_var`), `claude-code.sh:19-20`, and several ad-hoc `\w+` captures in `cowork.sh`. Worth a follow-up PR sweeping `\w+` → `[\w$]+` plus the `${var//\$/\\$}` escape consistently, mirroring `_common.sh`'s pattern. I'd rather get the #625 unblock landed first.

**2. The `grep -c 'async async function' → 0` check is exactly the right post-condition** for the third bug. Worth pulling into a small bats test eventually so the idempotency guard is verified every build, not only when you remember to grep. Not for this PR.

Approving.

---

PR #637, opened 2026-05-24
fix(launcher): exec Electron to fix Ctrl+C / signal forwarding

### LINKED ISSUE #424: CTRL + C doesn't close the app and process

aaddrick:
## Summary
- Add `exec` before Electron invocation in deb, RPM, and Nix launchers
- Removes unreachable exit-code logging after exec (low diagnostic value — these lines only logged *after* Electron exited, not during runtime)
- Fixes signal forwarding for SIGINT, SIGTERM, SIGHUP, SIGTSTP

Without `exec`, the launcher shell stays alive as a parent process. Electron handles SIGINT internally and never dies from it, so Ctrl+C in a terminal does nothing and the process tree keeps running. With `exec`, the shell is replaced by Electron, so signals go directly to the right process.

The AppImage launcher (`scripts/packaging/appimage.sh:118`) already uses `exec` correctly and is unchanged.

Fixes #424

## Test plan
- [x] Build deb package and verify Ctrl+C terminates the app
- [x] Verify `kill <pid>` of the launcher process terminates Electron
- [x] Verify AppImage behavior unchanged (already uses exec)

---
95% AI / 5% Human
Claude: implemented the fix based on issue analysis pipeline findings
Human: identified the quick win and directed implementation

### DIFF
--- a/nix/claude-desktop.nix
+++ b/nix/claude-desktop.nix
@@ -265,12 +265,10 @@ build_electron_args 'nix'
 # Add app path
 electron_args+=("$app_path")
 
-# Execute Electron
+# Execute Electron (exec replaces the shell process so signals
+# like SIGINT, SIGTERM, and SIGHUP reach Electron directly)
 log_message "Executing: $electron_exec ''${electron_args[*]} $*"
-"$electron_exec" "''${electron_args[@]}" "$@" >> "$log_file" 2>&1
-exit_code=$?
-log_message "Electron exited with code: $exit_code"
-exit $exit_code
+exec "$electron_exec" "''${electron_args[@]}" "$@" >> "$log_file" 2>&1
 LAUNCHER
     # Substitute placeholders — electron_exec points to our custom
     # wrapper (which sets GTK/GIO env then execs our merged binary)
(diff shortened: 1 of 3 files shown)


aaddrick:
Tested on Nobara 43, KDE Plasma 6, X11, upstream v1.0.8555.2.

Built RPM from clean state. Installed and launched via `claude-desktop`.

Results:
- Launcher PID (530939) became the `electron` process directly. The exec replaced the shell as intended
- `kill 530939` (SIGTERM) terminated Electron and all child processes cleanly
- Only cowork-vm-service.js daemon remained after kill (expected, separate process)
- Verified AppImage launcher already uses exec (appimage.sh:118, unchanged by this PR)

---

typedrat (approved):

PR #646, opened 2026-05-25
test(appimage): readiness-marker poll + unified cleanup trap

sabiut:
## Summary

Two follow-ups from #592 to tighten up the AppImage headless smoke test. Test-side only — no AppImage behavior changes.

## The flat sleep

`tests/test-artifact-appimage.sh` was launching the AppImage under Xvfb, sleeping 10 seconds, then checking the main process was still alive. Worst of both worlds — burned a flat 10s on every healthy run (Electron usually reaches ready in 1–2s), and on a slow CI runner 10s wasn't always enough buffer over real startup, so we'd flake.

Replaced it with a poll loop watching `launcher.log` for `[Frame Fix] Patches built successfully`. That line is the last thing `scripts/frame-fix-wrapper.js` emits after every patch is installed and the upgrade watcher is armed, so seeing it means main-process startup finished cleanly. 30s ceiling, 0.5s tick; each tick checks the marker first, then liveness via `kill -0`, so a marker emitted right before exit still counts as a pass.

Failure output got clearer too — used to be one generic "process died" message, now it splits into `AppImage did not reach ready state within Ns` (still alive, marker never showed) vs `AppImage exited before reaching ready state (exit: N)` (died early).

## The leaky trap

The original code from #592 set the `EXIT INT TERM` trap *inside* the smoke-test block, right after launching the AppImage. That trap cleaned up `$cache_root` and `$xvfb_log` but missed `$extract_dir` — the ~190MB `squashfs-root` we'd unpacked earlier in the script. Anyone Ctrl-C'ing the script before reaching the trap, or during the smoke test, would leak the extract dir.

First instinct was to add a second trap at script scope, but bash only keeps one handler per signal — the in-block one would just override it.

So: pulled the trap up to script scope into a `_cleanup` function that handles everything (`launch_pid` process group, `cache_root`, `xvfb_log`, `extract_dir`), and dropped the in-block one. Each branch in `_cleanup` is defensive (`[[ -n ${var:-} ]] && ...`) so it can fire safely no matter how far the script got before the interruption.

### DIFF
--- a/tests/test-artifact-appimage.sh
+++ b/tests/test-artifact-appimage.sh
@@ -143,15 +143,39 @@ if command -v xvfb-run &>/dev/null \
 		rm -rf "$cache_root" "$xvfb_log"
 	' EXIT INT TERM
 
-	# CI is slow; 10s is the floor for Electron startup.
-	sleep 10
+	# Wait up to 30s for the frame-fix readiness marker, or early
+	# process death. The marker is the last log line emitted by
+	# scripts/frame-fix-wrapper.js after all patches are installed,
+	# so reaching it means main-process startup finished without
+	# crashing. Replaces a flat 10s sleep that was both slow on
+	# healthy startups and a flake risk on noisy runners.
+	readiness_marker='[Frame Fix] Patches built successfully'
+	readiness_timeout=30
+	deadline=$((SECONDS + readiness_timeout))
+	saw_marker=0
+	while ((SECONDS < deadline)); do
+		if [[ -f $launcher_log ]] \
+			&& grep -qF "$readiness_marker" \
+				"$launcher_log"; then
+			saw_marker=1
+			break
+		fi
+		if ! kill -0 "$launch_pid" 2>/dev/null; then
+			break
+		fi
+		sleep 0.5
+	done
 
-	if kill -0 "$launch_pid" 2>/dev/null; then
-		pass "AppImage stays alive under Xvfb for 10s"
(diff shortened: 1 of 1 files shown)


aaddrick (approved):
Hey! I traced the readiness marker back through the `app.whenReady()` path and it checks out. `[Frame Fix] Patches built successfully` is the last log line after all patches apply and the upgrade watcher arms (frame-fix-wrapper.js:923), so polling for it is a reliable gate. The 30s timeout with 0.5s intervals is reasonable, and checking the marker before checking liveness handles the race where the app exits right after logging success. This is a solid replacement for the flat `sleep 10`, which was slow when things worked and flaky when they didn't.

The split failure messages are a nice touch. "Did not reach ready state" vs "exited before reaching ready state" tells you immediately whether the app hung or crashed. That's going to save time on CI triage.

The unified `_cleanup` trap fixes a real leak. The old in-block trap couldn't see `$extract_dir` and would get overridden by later traps anyway. Moving it to script scope with `${var:-}` guards is the right call. One minor nit: the comment block above `_cleanup` runs about 6 lines and reads more like commit message prose than an inline comment. The first line gives enough context on its own. Not a blocker though.

shellcheck passes clean. Only pre-existing SC1091 sourcing warnings, nothing new. Merge whenever you're ready.

---

sabiut:
Good catch — shortened in 110334d. Detail kept in the commit message + PR body where it belongs.

PR #640, opened 2026-05-24
fix(patches): reject .asar paths in directory check to prevent false Cowork dispatch

aaddrick:
## Summary
- Patch the directory-check function (`wFA`) to reject `.asar` paths before calling `fs.statSync().isDirectory()`
- Electron's ASAR virtual filesystem shim makes `.asar` archives report `isDirectory() === true`, causing `app.asar` to be dispatched to Cowork as a "folder drop"
- This single patch fixes three related issues with one root cause

**Symptoms fixed:**
- Permission dialog on every launch/reopen (#383)
- Forced Cowork mode on window reopen from tray (#622)
- Fatal `--add-dir` error in Claude Code >=2.1.111 causing "No conversation found" loop (#632)

Fixes #383
Fixes #622
Fixes #632

## Test plan
- [x] Build with patch and verify no patch warnings
- [x] Launch app — no "Allow Claude to change files in app.asar" dialog
- [x] Close and reopen via tray — no forced Cowork mode
- [x] Verify local agent mode works (no "No conversation found" loop)
- [x] Verify legitimate folder drops still trigger Cowork correctly

---
Generated with [Claude Code](https://claude.ai/code)
Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>
85% AI / 15% Human
Claude: implemented the fix based on issue analysis pipeline findings
Human: identified the quick win and directed implementation

PR #650, opened 2026-05-26
fix(patches): filter .asar paths from --add-dir dispatch and session restore

### LINKED ISSUE #649: Local agent mode still broken on 2.0.13 — app.asar reaches additionalDirectories via packaged-path helper, bypassing #640 guards
Follow-up to #632, which was closed as completed by #640 but **still reproduces** on the current release. Filing fresh since #632 can't be reopened by the (read-access) author after a maintainer close.

## Summary

Local agent mode fails on every turn with `No conversation found with session ID: <uuid>` on **claude-desktop 1.8555.2-2.0.13** (the release containing #640). Identical to #632: the turn aborts before any response (`had_first_response: false`), no transcript is written, and the next turn's `--resume` loops forever.

## #640 is present but insufficient

Both guards added by #640 are confirmed present in the installed `app.asar`:

```
function wFA(e){try{return!e.endsWith(".asar")&&ee.statSync(e).isDirectory()}catch{return!1}}
...addTrustedFolder: ${i}`);if(i.endsWith(".asar"))return;...
```

Yet the runtime still forwards the asar as `--add-dir`:

```
[HostLoop] sdkOptions after patch: { cwd: '.../local_<id>/outputs',
  executable: '~/.config/Claude/claude-code/2.1.111/claude',
  additionalDirectories: [ '/usr/lib/claude-desktop/node_modules/electron/dist/resources/app.asar' ], ... }
[HostLoop] cli.js stderr: .../resources/app.asar is not a directory. Did you mean to add the parent directory .../resources?
[HostLoop] cli.js stderr: No conversation found with session ID: 9cfe78e0-...
[Result] Turn failed ... subtype=error_during_execution, is_error=true
[Result] Clearing stale cliSessionId ... — transcript was unresumable (queue-ops only)
[CycleHealth] unhealthy_reason: 'api_error', error_category: 'session_not_found', had_first_response: false
```

## Root cause #640 missed

The asar path that reaches `additionalDirectories` does **not** pass through `wFA` (the `statSync().isDirectory()` check) or `addTrustedFolder` — the two spots #640 hardened. It's built directly by the packaged-path helper:

```js
const t = app.isPackaged ? path.join(process.resourcesPath, "app.asar") : app.getAppPath()
```

That value flows into `additionalDirectories` unconditionally, so the `.endsWith(".asar")` filters never see it. Bundled claude-code **2.1.111** treats a non-directory `--add-dir` as fatal, which is what aborts the turn.

## Suggested fix

Filter/validate at the point where `additionalDirectories` is assembled (drop non-directory or `.asar` entries before forwarding as `--add-dir`), rather than only at the `isDirectory` helper. Alternatively, point this entry at the resources **directory** (`.../resources` or `app.asar.unpacked`) instead of the `app.asar` archive file.

## Environment

- claude-desktop: **1.8555.2-2.0.13** (deb, amd64) — contains #640
- Bundled Claude Code: **2.1.111** (`~/.config/Claude/claude-code/2.1.111/claude`)
- OS: Linux Mint (Ubuntu noble base), kernel 6.17

## Workaround

Use the standalone Claude Code CLI for agent work; the desktop's plain cloud chat is unaffected.

Related: #632 (original, closed), #640 (partial fix), #383 / #622 (sibling symptoms of the same asar path).

aaddrick:
## Summary

- Filter `.asar` entries at the `--add-dir` CLI dispatch loop — the single convergence point where ALL `additionalDirectories` paths become CLI arguments to Claude Code
- Filter `.asar` entries at session restore to self-heal corrupted pre-#640 persisted state
- Add fallback regex for `.forEach` minifier variant, with count assertion for safety

Fixes #649

## Context

PR #640 patched `wFA()` and `addTrustedFolder` to reject `.asar` paths, but corrupted sessions persisted before #640 survive restore (Electron's ASAR VFS makes `existsSync` return true for `.asar` archives). The `.asar` path flows through `resolvedFolders` → `getFolderPermissionPaths` → `additionalDirectories` → `--add-dir`, where Claude Code >=2.1.111 fatally rejects it.

Rather than patching intermediate extraction functions (`dI()`, `cTe()`), this fix targets the final dispatch point where all code paths converge — including `MV()` managed settings, `computeSpawnAdditionalDirectories`, and the HostLoop merge.

## Test plan

- [x] Both sub-patches apply cleanly against 1.8555.2 reference source
- [x] Marker regex matches the patched output
- [x] Idempotency guard correctly prevents double-application
- [x] `shellcheck` passes (no new warnings)
- [x] `bats tests/verify-patches.bats` — all 9 tests pass (marker count bumped to >= 10)
- [ ] Full build (`./build.sh --build appimage --clean no`)
- [ ] Manual test: inject corrupted session with `.asar` in `userSelectedFolders`, verify no "No conversation found" error

## Open question

Need to confirm whether the bug reproduces after clearing session state (`~/.config/Claude/local-agent-mode-sessions/`). If it does, there's an active injection vector beyond stale persistence that warrants a separate follow-up. The primary patch covers both scenarios regardless.

---
80% AI / 20% Human
Claude: root-cause analysis, implementation, testing against reference source
Human: directed investigation, contrarian review iterations, plan approval

### DIFF
--- a/scripts/cowork-patch-markers.tsv
+++ b/scripts/cowork-patch-markers.tsv
@@ -4,11 +4,10 @@
 #     <name><TAB><pcre_pattern><TAB><sample>
 # Lines starting with '#' and blank lines are ignored.
 #
-# Each row names a post-patch fingerprint of patch_cowork_linux() in
-# scripts/patches/cowork.sh. Both verify-patches.sh and
-# tests/verify-patches.bats consume this file, so adding a marker
-# here adds it to the runtime check and the test matrix at the same
-# time.
+# Each row names a post-patch fingerprint from the patch suite in
+# scripts/patches/. Both verify-patches.sh and tests/verify-patches.bats
+# consume this file, so adding a marker here adds it to the runtime
+# check and the test matrix at the same time.
 #
 # Columns:
 #   name     — kebab-case id; surfaces in verify output and BATS names.
@@ -16,8 +15,9 @@
 #   sample   — concrete string the pattern matches; BATS uses it to
 #              build positive and per-marker negative fixtures.
 #
-# The 9 markers below correspond 1:1 with the smoke-test set defined
-# in issue #559 (PR #555 retrofit, deliverable D6).
+# The first 9 markers correspond to the smoke-test set defined in
+# issue #559 (PR #555 retrofit, deliverable D6). Additional markers
+# cover other critical patches (e.g., .asar guards).
 vmclient-log-gate	process\.platform==="linux"\)\s*\?\s*"vmClient \(TypeScript\)"	(F||process.platform==="linux")?"vmClient (TypeScript)"
 vm-assignment-linux-gate	process\.platform==="linux"\)\?\(?[\w$]+=\{vm:[\w$]+\}	(F||process.platform==="linux")?N={vm:M}
 unix-socket-path	process\.platform==="linux"\?\(process\.env\.XDG_RUNTIME_DIR\|\|"/tmp"\)\+"/cowork-vm-service\.sock"	process.platform==="linux"?(process.env.XDG_RUNTIME_DIR||"/tmp")+"/cowork-vm-service.sock"
@@ -27,3 +27,4 @@ econnrefused-on-linux	process\.platform==="linux"&&[\w$]+\.code==="ECONNREFUSED"
 cowork-daemon-pid	global\.__coworkDaemonPid	global.__coworkDaemonPid=_c.pid
 cowork-linux-daemon-shutdown	cowork-linux-daemon-shutdown	name:"cowork-linux-daemon-shutdown"
 sharedcwdpath-threadthrough	sharedCwdPath:this\.sessions\.get\(	sharedCwdPath:this.sessions.get(t)?.userSelectedFolders?.[0]
+asar-adddir-filter	\.filter\(_d=>!_d\.endsWith\("\.asar"\)\).*"--add-dir"	.filter(_d=>!_d.endsWith(".asar")))Y.push("--add-dir"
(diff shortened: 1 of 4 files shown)


sabiut (approved):
**Verdict:** Approve

**Validation as `/tests/` owner:**

Pulled `fix/649-asar-additional-dirs` and ran the test surface locally:

- `bats tests/verify-patches.bats` — all 9 tests pass; "regex matches own sample" succeeds for the new `asar-adddir-filter` row; the bumped `≥ 10 markers` threshold check passes.
- `shellcheck scripts/patches/{config,app-asar}.sh` — only the pre-existing baseline noise (`SC2148` / `SC2154` for sourced globals like `$project_root`). No new warnings introduced.
- CI rollup is green across all 15 checks (BATS, shellcheck, codespell, test-flags, all 6 build matrices, all 3 artifact validations).

**Substance:**

- Targeting the `--add-dir` dispatch loop is the right call — it's the single convergence point for `MV()` managed settings, `computeSpawnAdditionalDirectories`, the HostLoop merge, and the `path.join(process.resourcesPath, "app.asar")` packaged-path helper that #649 identified as the active vector. Catches all of them without chasing intermediate functions.
- Failure modes are well-allocated: primary sub-patch is fail-fast (`exit 1` if anchor disappears), secondary sub-patch is warn-only via `warn()` (correct — it's belt-and-suspenders for stale persisted state, and the comment explicitly notes the primary filter still protects).
- Marker pattern `\.filter\(_d=>!_d\.endsWith\("\.asar"\)\).*"--add-dir"` is a self-fingerprint on the literal `_d` chosen by the patch itself, not on upstream minified identifiers — survives re-minification. Matches both for-of and forEach variants.
- Idempotency guard via `grep -qF` on a unique substring; theoretical false-positive risk if upstream introduces an unrelated `_d`-named filter is negligible.

Issue #651, opened 2026-05-26
Model stream stalls mid-response; UI timer keeps running, no tokens arrive (LocalSessions IPC query teardown)


LiukScot:
### Version (`claude-desktop --doctor` output)

```
Claude Desktop Diagnostics
================================

[PASS] Display server: Wayland (WAYLAND_DISPLAY=wayland-0)
       Desktop: GNOME
       Mode: X11 via XWayland (default, for global hotkey support)
[PASS] Electron: v41.5.0
[PASS] Chrome sandbox: permissions OK
[PASS] SingletonLock: held by running process (PID 330222)
[PASS] MCP config: valid JSON (0 servers configured)
[PASS] Node.js: v22.22.2
[PASS] Desktop entry: /usr/share/applications/claude-desktop.desktop
[PASS] Disk space: 36001MB free

Cowork Mode
----------------
[PASS] bubblewrap, KVM, vsock, QEMU, virtiofsd: all found
       socat: not found
[PASS] Cowork daemon: running (parent alive)
[PASS] Log file: 1566KB (/home/luca/.cache/claude-desktop-debian/launcher.log)

All checks passed.
```

Package: `claude-desktop-1.7196.3-2.0.12.fc42.x86_64`
Host: Fedora 44, GNOME Wayland session, Electron via XWayland (default).
Network: Tailscale VPN active.

### What happened

Model response stream stalls mid-generation. UI stays alive (elapsed-time counter keeps incrementing, app responsive), but no further tokens arrive from the model. Only recovery: click Stop and resend the message.

Affects both main chat and spawned subagents. Higher frequency with subagents — spawn ~10 subagents and at least one will stall. Random, not tied to message length or content.

Onset: ~2 weeks ago (around 2026-05-12).

### Log evidence (~/.cache/claude-desktop-debian/launcher.log)

Two distinct error patterns observed around stall events.

**1. IPC query torn down before stream response is consumed** (occurred 2× in this excerpt, channel `LocalSessions_$_setFastMode`):

```
Error occurred in handler for '$eipc_message$_b21ff65e-f132-4859-9a4f-2b7581a1bec5_$_claude.web_$_LocalSessions_$_setFastMode': Error: Query closed before response received
    at mPi.performCleanup (.../app.asar/.vite/build/index.js:406:16463)
    at mPi.cleanup       (.../app.asar/.vite/build/index.js:406:16117)
[...]

PR #665, opened 2026-05-29
fix: shorten Claude Code local-session stall watchdog (10min → 5min)

LiukScot:
## Problem

On Linux, **Claude Code local agent sessions** (and spawned subagents) occasionally **stall mid-stream**: the elapsed-time counter keeps running, no tokens arrive, and **no error is ever shown**. The only recovery is to manually click Stop and resend. Windows users hit this far less. Tracked in #651.

> **Naming note:** the bug is in the Claude Code local-session surface, *not* the Cowork VM sandbox. This patch happens to live in `cowork.sh` because that is where the `index.js` bundle string-patches live in this repo, and the watchdog's config key is literally named `coworkMessageTimeoutMs` by upstream (it governs the local agent/session timer). No Cowork-VM behaviour is changed.

## Root cause (investigated against the installed 1.8555.2 build)

Two distinct layers:

1. **The actual stall is upstream, in the Claude Code CLI — not this repackaging.** The bundled Claude Code CLI subprocess (`/$bunfs/.../cli.js`) throws an `AbortError` inside a hook callback / `control_request` (`sendRequest`) path after a tool use, then stops emitting stdout. The model stream goes silent with `hadFirstResponse=true`. Reported upstream: anthropics/claude-code#63584 (see also anthropics/claude-code#51851, the same `sendRequest` wedge).

2. **The recovery watchdog already exists and works — but is gated at 10 minutes.** The local-session health monitor (`wRr.startTimeoutDetection`) marks a session timed out after `coworkMessageTimeoutMs` of stdout inactivity (default `6e5` = 10 min), and on timeout the manager emits:

   ```
   {type:"error", error:"The session stopped responding. Send your message again to resume with a fresh process."}
   ```

   and tears the query down. That is exactly the desired UX — but at a 10-minute threshold, users hard-stop manually long before it fires. In a 100k-line `main.log` it fired **once**.

Evidence (`~/.config/Claude/logs/main.log`, 1.8555.2):

```
[CCD] Session local_… timed out after 652s of inactivity (hadFirstResponse=true, last_tool_name=Write, seconds_since_stderr=53)
[CCD] Session … stderr tail:
  AbortError:
      at A (/$bunfs/root/src/entrypoints/cli.js:9486:473)
  Error in hook callback hook_1:
```

## Fix

Lower the `coworkMessageTimeoutMs` default from `6e5` (10 min) to `3e5` (5 min) via a new bundle patch (Patch 13 in `scripts/patches/cowork.sh`). This makes the **existing, already-working** recovery surface the error ~2× sooner. 5 minutes stays well above any legitimate single-tool silence (a long Bash/build emits no stdout messages while it runs), so it should not falsely time out long tool turns.

- Anchored on the stable config-key string `"coworkMessageTimeoutMs",` (whitespace-tolerant after the comma per `CONTRIBUTING.md`) so it survives Claude version bumps; the numeric default is matched generically.
- Asserts a single match site; only ever **shortens** the value (never lengthens); skips if already low → idempotent, safe to re-run.
- Registers a `verify-patches` marker (`cowork-message-timeout-shortened`) in `scripts/cowork-patch-markers.tsv` so CI flags a silent no-op if the anchor moves upstream.

## Caveats

- This is a **stopgap**, not a cure. The real fix is upstream (the Claude Code CLI hook-abort that wedges the stream) — anthropics/claude-code#63584.
- `coworkMessageTimeoutMs` is read through a remote-config getter; a remote value, if upstream ever pushes one, would override this local default. None is observed in current builds.

## Test

- `bash -n scripts/patches/cowork.sh` passes.
- Patch 13 logic dry-run against the extracted 1.8555.2 `index.js`: rewrites `6e5`→`3e5`, second run is a no-op (idempotent).
- New marker validated: pattern matches the patched output and its own sample; all existing marker rows still pass.
- [ ] Full build + manual confirmation that a stalled session now surfaces "stopped responding" at ~5 min (needs a maintainer build).

### DIFF
--- a/scripts/cowork-patch-markers.tsv
+++ b/scripts/cowork-patch-markers.tsv
@@ -28,3 +28,4 @@ cowork-daemon-pid	global\.__coworkDaemonPid	global.__coworkDaemonPid=_c.pid
 cowork-linux-daemon-shutdown	cowork-linux-daemon-shutdown	name:"cowork-linux-daemon-shutdown"
 sharedcwdpath-threadthrough	sharedCwdPath:this\.sessions\.get\(	sharedCwdPath:this.sessions.get(t)?.userSelectedFolders?.[0]
 asar-adddir-filter	\.filter\(_d=>!_d\.endsWith\("\.asar"\)\).*"--add-dir"	.filter(_d=>!_d.endsWith(".asar")))Y.push("--add-dir"
+cowork-message-timeout-shortened	"coworkMessageTimeoutMs",\s*3e5	"coworkMessageTimeoutMs",3e5
--- a/scripts/patches/cowork.sh
+++ b/scripts/patches/cowork.sh
@@ -882,7 +882,7 @@ if (serviceErrorIdx !== -1) {
 }
 
 // ============================================================
-// Patch 11: Shorten local-session "no-stream" watchdog 10min -> 5min
+// Patch 13: Shorten local-session "no-stream" watchdog 10min -> 5min
 //
 // The local-agent health monitor (wRr.startTimeoutDetection) marks
 // a session "timed out" only after coworkMessageTimeoutMs of stdout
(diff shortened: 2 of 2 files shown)


LiukScot:
this is not the root cause of the problem

PR #666, opened 2026-05-29
Fix Fedora launch regression and KDE tray quit handling

DhanushSantosh:
## Summary
- fix the tray patch matcher for the current upstream menu/context-menu shape so KDE tray updates stop rebuilding the StatusNotifierItem unnecessarily
- fix the trusted-folder `.asar` guard injection so newer minified bundles do not produce a malformed `index.js`
- auto-disable GPU on the next launch after the known Chromium GPU-process fatal path, with docs and launcher tests

## Problem
On Fedora RPM installs we hit two separate regressions while testing the current upstream bundle:

1. The rebuilt app could fail immediately with a JavaScript `SyntaxError` inside `addTrustedFolder()` because the `.asar` guard patch in `scripts/patches/config.sh` was inserting into the wrong expression shape in the newer minified bundle.
2. After fixing that parse failure, the app still failed to stay open on Fedora KDE/Wayland because Electron 41 was hitting the known `GPU process isn't usable. Goodbye.` crash path unless `CLAUDE_DISABLE_GPU=1` was set.
3. Separately, KDE tray `Quit` was inert because the tray fast-path patch no longer matched the current upstream tray menu shape, so the app fell back to the destroy/recreate SNI path that races on Plasma.

## Fix
- update `patch_tray_inplace_update()` to support the current `menuVar = menuFunc(); tray.setContextMenu(menuVar)` pattern as well as the older direct-call shape
- retarget the trusted-folder injection to the `async addTrustedFolder(...) {` function body so the `.asar` guard lands in valid JavaScript again
- teach the launcher to automatically apply `--disable-gpu --disable-software-rasterizer` on the next launch after the prior log section shows the GPU-fatal signature, while preserving explicit env-var override behavior
- document the auto-recovery path and cover it with `launcher-disable-gpu.bats`

## Validation
- verified the rebuilt `app.asar.contents/.vite/build/index.js` passes `node --check` after the config patch fix
- verified on the Fedora install that the app progressed past the prior `SyntaxError` and that launches with GPU disabled stayed alive
- verified the launcher log shows the expected GPU-disable path once the workaround is active
- could not run `bats` or `shellcheck` locally on this Fedora machine because those tools were not installed in the environment

---
Generated with Codex (GPT-5)
Co-Authored-By: OpenAI Codex <noreply@openai.com>
90% AI / 10% Human
Codex: investigated the Fedora RPM failure, patched the tray/config/launcher logic, updated docs/tests, and verified the installed app behaviour.
Human: reproduced the Fedora install issues, requested the fix direction, and validated the real-machine packaging flow.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,7 @@ result-*
 
 # Wrangler (Cloudflare Worker dev/deploy cache)
 worker/.wrangler/
+
+# Graphify outputs and temporary files
+graphify-out/
+.graphify_*
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,15 @@ worker/.wrangler/
 # Graphify outputs and temporary files
 graphify-out/
 .graphify_*
+
+# Local agent/editor state and helper bins
+.agents/
+.codex/
+.tmpbin/
+
+# Local package artifacts
+*.rpm
+
+# Root-level scratch extracts from app inspection
+/frame-fix-wrapper.js
+/index.js
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -163,8 +163,10 @@ the launcher won't stack duplicate flags.
 
 If the previous launch already died with the GPU-process FATAL
 signature and `CLAUDE_DISABLE_GPU` is unset, the next launch
-auto-applies the same flags. `CLAUDE_DISABLE_GPU=0` suppresses that
-auto-fallback if you need to retest hardware acceleration.
+auto-applies the same flags and keeps them applied on subsequent
+launches. Set `CLAUDE_DISABLE_GPU=0` to suppress the auto-fallback
+when retesting hardware acceleration after a driver fix — any
+explicitly set value suppresses it; only `1` forces the flags on.
 
 **When to prefer which:** the in-app toggle is friendlier if you
 can reach Settings without the app crashing. Reach for
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -161,6 +161,11 @@ applied automatically inside XRDP sessions, where software
 rendering is required regardless. Either signal is sufficient —
 the launcher won't stack duplicate flags.
 
+If the previous launch already died with the GPU-process FATAL
+signature and `CLAUDE_DISABLE_GPU` is unset, the next launch
+auto-applies the same flags. `CLAUDE_DISABLE_GPU=0` suppresses that
+auto-fallback if you need to retest hardware acceleration.
+
 **When to prefer which:** the in-app toggle is friendlier if you
 can reach Settings without the app crashing. Reach for
 `CLAUDE_DISABLE_GPU=1` when the app crashes before you can open
--- a/scripts/launcher-common.sh
+++ b/scripts/launcher-common.sh
@@ -179,10 +179,10 @@ _previous_launch_hit_gpu_fatal() {
 				exit 1
 			}
 			text = sections[target]
-			if (
-				index(text, "GPU process launch failed: error_code=") &&
-				index(text, "GPU process isn'\''t usable. Goodbye.")
-			) {
+			if (index(text,
+				"GPU process launch failed: error_code=") &&
+				index(text,
+				"GPU process isn'\''t usable. Goodbye.")) {
 				exit 0
 			}
 			exit 1
--- a/scripts/launcher-common.sh
+++ b/scripts/launcher-common.sh
@@ -156,6 +156,40 @@ _detect_password_store() {
 	echo 'basic'
 }
 
+# Detect whether the previous launch ended in Chromium's
+# "GPU process isn't usable" crash signature (#583).
+#
+# setup_logging() must have run first so $log_file is available. The
+# launcher writes the current session header before build_electron_args()
+# runs, so the previous launch lives in the penultimate log section.
+_previous_launch_hit_gpu_fatal() {
+	[[ -f ${log_file:-} ]] || return 1
+
+	awk '
+		/^--- Claude Desktop (Launcher|AppImage) Start ---$/ {
+			section++
+			next
+		}
+		{
+			sections[section] = sections[section] $0 "\n"
+		}
+		END {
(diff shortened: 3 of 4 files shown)


sabiut (changes requested):
One blocker before this can merge: the new GPU-fatal detection doesn't run. `_previous_launch_hit_gpu_fatal()` is invalid awk and aborts at parse time:

```
awk: line 15: missing ) near end of line
awk: line 18: extra ')'
```

awk doesn't allow a newline immediately after the opening `(` of an `if`. As written, the function always errors out, so the auto-recovery never triggers — the core feature of the PR is a no-op.

Your own test catches this: `tests/launcher-disable-gpu.bats` test 7 ("prior GPU fatal auto-disables on next launch") fails locally. I ran the suite — 7/8 pass, only that one is red. (Makes sense given the note that bats wasn't available on the Fedora box.)

The fix is to pull the condition onto the `if (` line — continuation after `&&` is fine:

```awk
if (index(text, "GPU process launch failed: error_code=") &&
    index(text, "GPU process isn't usable. Goodbye.")) {
	exit 0
}
```

I verified the corrected form returns 0 when the fatal signature is in the previous log section and 1 otherwise, and test 7 then passes.

DhanushSantosh:
Addressed the GPU-fatal auto-recovery blocker.

I reproduced the issue locally exactly as reported:
- `tests/launcher-disable-gpu.bats` was failing on test 7
- the failing path emitted the same AWK parse error from `_previous_launch_hit_gpu_fatal()`

The fix is now pushed in `eff4a1d` (`Fix GPU fatal awk detection`). I changed the AWK condition so the `if (` starts on the same line as the condition expression instead of breaking immediately after the opening parenthesis.

Validation after the change:
- `bats tests/launcher-disable-gpu.bats` → 8/8 passing
- `shellcheck scripts/launcher-common.sh` → only the existing `SC1091` sourced-file info note

Thanks for catching that one.

---
Written by Codex via ChatGPT

sabiut (approved):
Verified locally on `eff4a1d`:

- `bats tests/launcher-disable-gpu.bats` → **8/8** (test 7, the one that was failing, now passes)
- the awk in `_previous_launch_hit_gpu_fatal()` parses standalone and detects the GPU-fatal signature correctly
- `launcher-common.bats` + `launcher-xrdp-detection.bats` green — no regression
- `shellcheck scripts/launcher-common.sh` → only the pre-existing SC1091 sourced-file note (resolved by `-x` in CI)

The GPU-fatal auto-recovery blocker is resolved — thanks for the fast turnaround. Approving from the test-ownership side.

Note: the `config.sh` / `tray.sh` patch changes are outside my lane and still want @aaddrick's review. The `config.sh` `addTrustedFolder` fix here is also what unblocks the currently-red package builds on `main`, so it'd be good to get this in.

DhanushSantosh:
@sabiut okay let's wait for @aaddrick response

aaddrick:
@DhanushSantosh this is approved and the tray + GPU work is wanted. One thing changed under you: the addTrustedFolder anchor fix landed on main via #685, so the config.sh hunk here is now redundant and will conflict.

Could you rebase on main and drop the config.sh changes? Keep the tray.sh menu/context-menu matcher fix and the GPU auto-disable work (launcher + docs + tests) as-is. Those are the parts we still need.

Minor while you're in there: the `if ! FOLDER_PARAM=...` line lost its leading tab somewhere in the edit, so it's sitting at no indent. The bash style guide wants tabs, so worth fixing on the same pass.

aaddrick:
Hey @DhanushSantosh! Thanks again for this, and for the fast awk turnaround with @sabiut. The GPU auto-recovery (launcher + bats + docs) is the piece we want, and I'm going to finish it on this branch directly: rebase onto main, drop the superseded hunks, fix two small things, push. You stay the author on the recovery work.

One correction to my June 4 comment. I asked you to keep the tray.sh fix, but #680 merged 43 minutes after I wrote that, and main's tray.sh now has its own `menu_var` fallback for the same prebuilt-menu shape. Main's version resolves the builder name and emits `setContextMenu(builder())` on purpose: the captured var is a `Menu` instance, so calling it would throw (see `docs/learnings/tray-rebuild-race.md`). So the rebase takes main's `config.sh` and `tray.sh` wholesale. That also makes the lost-tab nit disappear.

Final review turned up two real bugs in the recovery logic, both tiny, and I'll fix them in the same push. First: a recovered launch runs with `--disable-gpu` and writes no GPU output, so launch N+2 sees a clean penultimate section and re-enables GPU. On the permanently broken hardware #583 describes, that means crash, work, crash, forever. The fix is a sticky marker: the awk also accepts the launcher's own "disabling GPU" line as a trigger, with `CLAUDE_DISABLE_GPU=0` as the escape hatch for retesting after a driver fix. Second: the section-header regex misses Nix, since `nix/claude-desktop.nix` writes `Launcher Start (NixOS)`, so the helper silently no-ops there. One regex alternation fixes it. A bats case each will lock both in.

Push coming shortly. If anything in the plan looks off, flag it here and I'll adjust.

---

aaddrick:
Pushed as `29fa6da`. The branch is rebased onto main with your four commits and authorship intact. `config.sh` and `tray.sh` are now byte-identical to main (the #685/#680 supersession), so the net diff is just the GPU auto-recovery: `launcher-common.sh`, its bats tests, the troubleshooting docs, and `.gitignore`.

The two fixes from my last comment are in. Recovery is now sticky: the awk also accepts the launcher's own FATAL marker in the penultimate section, so a recovered launch no longer re-enables GPU and re-crashes on launch N+2. `CLAUDE_DISABLE_GPU=0` stays the escape hatch. And the header regex now matches `Launcher Start (NixOS)`, so Nix users get recovery at all. 282/282 bats green, shellcheck clean.

One honest caveat: the sticky behavior is verified against simulated log sections, not on real #583 hardware. If you (or anyone reading this with affected hardware) can confirm the crash → recover → stays-recovered sequence across three launches, that would close the loop. Merging once CI is green.

---

Issue #622, opened 2026-05-17
[bug]: Each time i close Claude windows and reopen it claude start on "Cowork Mode"


mathys-lopinto:
### Version (`claude-desktop --doctor` output)

```shell
❯ claude-desktop --doctor
Claude Desktop Diagnostics
================================

[PASS] Display server: X11 (DISPLAY=:0)
       Menu bar mode: auto (default, Alt toggles visibility)
       Titlebar style: hybrid (default, native frame + in-app topbar)
[PASS] Electron: v41.5.0 (/tmp/.mount_clauden6tLrj/usr/lib/node_modules/electron/dist/electron)
[FAIL] Chrome sandbox: perms=755, owner=root
       Fix: sudo chown root:root /tmp/.mount_clauden6tLrj/usr/lib/node_modules/electron/dist/chrome-sandbox
            sudo chmod 4755 /tmp/.mount_clauden6tLrj/usr/lib/node_modules/electron/dist/chrome-sandbox
[PASS] SingletonLock: held by running process (PID 152191)
[PASS] MCP config: valid JSON (/home/mathys/.config/Claude/claude_desktop_config.json)
       MCP servers configured: 0
[PASS] Node.js: v26.1.0
       Path: /usr/bin/node
[PASS] Desktop entry: /usr/share/applications/claude-desktop.desktop
[PASS] Disk space: 555552MB free

Cowork Mode
----------------
[PASS] bubblewrap: found
[PASS] bubblewrap: sandbox probe succeeded
[PASS] KVM: accessible
[PASS] vsock: module loaded
[PASS] QEMU: found
[PASS] socat: found
[PASS] virtiofsd: found at /usr/lib/virtiofsd (not on PATH)
       VM image: not downloaded yet
       Cowork isolation: bubblewrap (namespace sandbox)
[PASS] Cowork daemon: running (parent alive)
[PASS] Log file: 4925KB (/home/mathys/.cache/claude-desktop-debian/launcher.log)

1 check(s) failed.
See above for fixes.

❯ yay -Q claude-desktop-appimage 
claude-desktop-appimage 2.0.11+claude1.7196.1-1
```

### What happened

Each time i close the claude window and reopen it, it always reopen on Cowork New Task page.

### Steps to reproduce

1. Talk in a chat with Claude
2. Close the window
3. Reopen Claude

https://github.com/user-attachments/assets/b1b9db1a-753a-425f-a519-f360607b2302
[...]

Issue #383, opened 2026-04-06
app.asar permission.


awake4real:
When invoking Claude Desktop in LM22.1 XIA, this always appears in the  default start page in Cowork:
Allow Claude to change files in "app.asar"?
This includes all files and subfolders. Claude will be able to read, edit, and permanently delete—and may share file contents with third-party tools it connects to. Be careful about exposing sensitive information. 

Included are 3 buttons,  Cancel, Always. Yes

I don't want to use Cowork, and since I only click Cancel, it comes up every time I start Claude.

ramonsmits:
Just installed it using Fedora instructions in the README:

```
# Add the repository
sudo curl -fsSL https://aaddrick.github.io/claude-desktop-debian/rpm/claude-desktop.repo -o /etc/yum.repos.d/claude-desktop.repo

# Install
sudo dnf install claude-desktop
```

I'm experiencing this same issue on Fedora 43:

## Software Information:
- **Firmware Version:**                            3602
- **OS Name:**                                     Fedora Linux 43 (Workstation Edition)
- **OS Build:**                                    (null)
- **OS Type:**                                     64-bit
- **GNOME Version:**                               49
- **Windowing System:**                            Wayland
- **Kernel Version:**                              Linux 6.19.11-200.fc43.x86_64

RayCharlizard:
Good news — this looks like two bugs in one, both addressed by fixes that landed this week and first shipped in `v2.0.0+claude1.3109.0`, released earlier today:

- **Lands on Cowork on every launch** → fixed by #433. An empty Linux bundle manifest was making `getDownloadStatus()` vacuously report `Ready` every startup, tripping the same "download just finished" auto-navigation that macOS/Windows get once after install. Linux users hit it on every boot.
- **Permission dialog names `app.asar`** → fixed by #392 and #436. Cowork's host-side cwd resolution was failing (the app sent `cwd=/sessions/{name}`, which didn't match the `/mnt/{mount}` shape HostBackend expected) and falling back to Electron's install directory — which surfaces as `app.asar` in the filesystem permission UI. The fixes derive cwd from the primary user mount and thread `userSelectedFolders[0]` through to the daemon.

Neither of your installs would have had these — @awake4real your `1.1.5368-1.3.17` predates them by a lot, and @ramonsmits your Apr 14 `dnf install` pulled `v1.3.30+claude1.2278.0`, still too early. `v2.0.0+claude1.3109.0` is the first release bundling all three.

[...]

PR #669, opened 2026-05-29
fix(patches): reject .asar paths in argv file-drop collector (#668)

### LINKED ISSUE #668: [bug]: app.asar still reaches file drop handler on every cowork screen focus — incomplete fix after #640/#650 (v2.0.16, KDE, X11)

```shell
Claude Desktop Diagnostics

[PASS] Installed version: 1.9255.2-2.0.16
       Desktop: KDE
            (disables global hotkeys)
[PASS] SingletonLock: held by running process (PID 20215)
[PASS] MCP config: valid JSON (/home/mitch/.config/Claude/claude_desktop_config.json)
       MCP servers configured: 1
[PASS] Node.js: v22.22.1
       Path: /home/mitch/.nvm/versions/node/v22.22.1/bin/node
[PASS] Disk space: 513410MB free

Cowork Mode
----------------
[PASS] bubblewrap: found
[PASS] KVM: accessible
[PASS] vsock: module loaded
       QEMU: not found
[PASS] socat: found
       virtiofsd: not found
       VM image: not downloaded yet
[PASS] Log file: 2386KB (/home/mitch/.cache/claude-desktop-debian/launcher.log)
```

### What happened

After #640 and #650, the folder drop path for app.asar appears to be patched, but a separate file drop handler is still receiving the path. The prompt to attach app.asar appears every time the cowork/landing screen is shown — not just at startup, but throughout the session.

Version: claude-desktop 1.9255.2-2.0.16 (deb, amd64)
KDE Plasma 5.27.12, Ubuntu 24.04.4 LTS 

Prior to the #640 fix (April logs), this was Handling folder drop — so the folder-drop path was fixed, but the file-drop path was not. The .asar path is reaching setDraftSessionFolders and the user is prompted on every cowork screen focus.

### Steps to reproduce

1. Install claude-desktop deb on Ubuntu 24.04 with KDE Plasma 5.27
2. Launch the app — it opens to the cowork screen by default
3. Prompt immediately appears asking to attach app.asar
4. Close the window (app remains in system tray)
5. Re-open from the taskbar
6. Prompt reappears

### Expected behavior

Expected: app.asar is never surfaced to the user in any context
Actual: Prompt appears every time the window is closed and re-opened from the taskbar

### Logs / errors

```shell
Log evidence from today (~/.config/Claude/logs/main.log), firing repeatedly:


Handling file drop: /usr/lib/claude-desktop/node_modules/electron/dist/resources/app.asar
LocalAgentModeSessions.setDraftSessionFolders: 0 folders
setDraftSessionFolders: setting 0 folders for draft session
```

### Anything else

Note: This is compounded by #622 (also regressed in v2.0.16) — the app always reopens to the cowork screen on window close+reopen, meaning there's no way to avoid the prompt on every launch.

MitchSchwartz:
## Summary

Fixes the permission prompt that appears on every window close+reopen from the taskbar (#668, regression of #383 and #622).

- PR #640 patched `isDirectory()` in the directory-check helper, blocking the **folder-drop** path for `app.asar`
- The argv collector function (`lKr`) has a separate **file-drop** branch using `existsSync()` that was never patched
- Electron's ASAR VFS shim makes `existsSync()` return `true` for `.asar` paths, so `app.asar` passes the check and is dispatched to `cCA()` as a file drop every time the second-instance handler fires (i.e. every taskbar reopen)
- The startup code path already excludes the app bundle via `tA.resolve(n) !== appPath`; `lKr()` — used only for second-instance relaunches — has no equivalent guard

**Fix:** inject `!PARAM.endsWith(".asar")&&` before the `existsSync()` call in `lKr()`:

```
// Before
!i.startsWith("-") && ee.existsSync(i)

// After
!i.startsWith("-") && !i.endsWith(".asar") && ee.existsSync(i)
```

## Changes

- `patch_asar_argv_file_drop_guard()` added to `scripts/patches/cowork.sh`
- Wired into `patch_app_asar()` in `scripts/patches/app-asar.sh` immediately after `patch_asar_path_filter`
- `asar-file-drop-guard` marker added to `scripts/cowork-patch-markers.tsv`

## Test plan

- [ ] Build deb package and verify `verify-patches.sh` passes the new `asar-file-drop-guard` marker
- [ ] Close Claude window, reopen from taskbar — confirm no `app.asar` attach prompt
- [ ] Confirm `Handling file drop: .../app.asar` no longer appears in `~/.config/Claude/logs/main.log` on reopen

Verified anchor uniqueness (1 match) and idempotency against extracted `index.js` from v2.0.16 (1.9255.2). Idempotency check is context-anchored to `startsWith("-")` to avoid false-positives from other `.asar` guards (statSync patch, `--add-dir` filter).

Closes #668
Related: #383, #622, #640

---
Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>
95% AI / 5% Human
Claude: root cause investigation, patch implementation, issue analysis, code review
Human: bug report, reproduction details, direction

### DIFF
--- a/scripts/cowork-patch-markers.tsv
+++ b/scripts/cowork-patch-markers.tsv
@@ -28,3 +28,4 @@ cowork-daemon-pid	global\.__coworkDaemonPid	global.__coworkDaemonPid=_c.pid
 cowork-linux-daemon-shutdown	cowork-linux-daemon-shutdown	name:"cowork-linux-daemon-shutdown"
 sharedcwdpath-threadthrough	sharedCwdPath:this\.sessions\.get\(	sharedCwdPath:this.sessions.get(t)?.userSelectedFolders?.[0]
 asar-adddir-filter	\.filter\(_d=>!_d\.endsWith\("\.asar"\)\).*"--add-dir"	.filter(_d=>!_d.endsWith(".asar")))Y.push("--add-dir"
+asar-file-drop-guard	\.startsWith\("-"\)&&![\w$]+\.endsWith\("\.asar"\)	.startsWith("-")&&!i.endsWith(".asar")
--- a/scripts/cowork-patch-markers.tsv
+++ b/scripts/cowork-patch-markers.tsv
@@ -28,4 +28,4 @@ cowork-daemon-pid	global\.__coworkDaemonPid	global.__coworkDaemonPid=_c.pid
 cowork-linux-daemon-shutdown	cowork-linux-daemon-shutdown	name:"cowork-linux-daemon-shutdown"
 sharedcwdpath-threadthrough	sharedCwdPath:this\.sessions\.get\(	sharedCwdPath:this.sessions.get(t)?.userSelectedFolders?.[0]
 asar-adddir-filter	\.filter\(_d=>!_d\.endsWith\("\.asar"\)\).*"--add-dir"	.filter(_d=>!_d.endsWith(".asar")))Y.push("--add-dir"
-asar-file-drop-guard	\.startsWith\("-"\)&&![\w$]+\.endsWith\("\.asar"\)	.startsWith("-")&&!i.endsWith(".asar")
+asar-file-drop-guard	\.startsWith\("-"\)\s*&&\s*![\w$]+\.endsWith\("\.asar"\)	.startsWith("-")&&!i.endsWith(".asar")
--- a/scripts/patches/app-asar.sh
+++ b/scripts/patches/app-asar.sh
@@ -108,6 +108,11 @@ console.log('Updated package.json: main entry, desktopName, and node-pty depende
 	# trigger false Cowork dispatch (#383, #622, #632).
 	patch_asar_path_filter
 
+	# Reject .asar paths in the argv file-drop collector so the
+	# existsSync branch doesn't dispatch app.asar as a file drop,
+	# triggering a permission prompt on every window reopen (#383, #622).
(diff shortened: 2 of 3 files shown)


aaddrick (commented):
@MitchSchwartz — solid work, and the writeup made this easy to follow. You found the second `app.asar` path: the `existsSync()` branch in the argv collector that #640 never touched. The fix mirrors the `statSync` guard exactly. I dry-ran it against the real minified bundle. It patches, verifies, and re-runs idempotently, so the reopen-prompt bug is fixed.

A few small things I pushed to the branch rather than round-tripping them through you (pushed as `5772cc1`):

The match regex tolerates whitespace, but the node post-verify, the idempotency grep, and the TSV marker all hardcode `startsWith("-")&&` with no `\s*`. It can't break a real build, since the build only ever sees minified input and packs only on success. But it'd miss against beautified source, so I added `\s*` around the `&&` in all three for the both-forms rule.

I also dropped the dead `cd "$project_root"` before the unconditional `exit 1` to match the sibling guard.

Last one: I added a threat-model comment that wasn't there before. The argv here can actually carry user-launched paths via `Exec=... %u`, not just the app's own relaunch. But the only sink is attach-to-draft, the same as a manual drag, so exact-suffix on the app's lowercase `app.asar` is the right call. I wrote that down so nobody later "hardens" it with a `toLowerCase()` that'd diverge from the sibling guards for zero gain.

Thanks for chasing this one all the way down.

---

PR #671, opened 2026-05-29
test(artifacts): headless launch smoke test for deb and rpm

### LINKED ISSUE #670: test: headless launch smoke test for deb and rpm artifacts
## Gap

The artifact smoke tests cover three formats, but only AppImage actually launches the app:

| Format | Install check | Launch check |
|---|---|---|
| AppImage | ✅ | ✅ (xvfb readiness-marker poll, #646) |
| deb | ✅ | ❌ static file checks only |
| rpm | ✅ | ❌ static file checks only |

`test-artifact-deb.sh` and `test-artifact-rpm.sh` install the package and assert that files like `launcher-common.sh` exist, but never run `/usr/bin/claude-desktop`. Runtime-only regressions are invisible on these two formats.

## Why now

PR #666 fixed a Fedora **launch** regression — a JS `SyntaxError` from a bad patch anchor that killed the app on startup. The rpm artifact test would not have caught it: it only checks that files are present, never that the app boots. The exact regression class #666 addressed is currently untested on the two formats where it occurred.

## Proposed work

Port the AppImage launch pattern (setsid process-group isolation, `[Frame Fix] Patches built successfully` readiness poll with a 30s deadline, EXIT-trap cleanup) to:

- **deb** — installs to a predictable `/usr/bin/claude-desktop`; mostly a lift-and-shift, and turns the currently-wasted `xvfb dbus-x11 procps` install on the deb job into a used dependency.
- **rpm** — runs in the `fedora:42` container, so it needs xvfb deps added there and headless handling inside the container.

Both formats in one PR.

sabiut:
Closes #670.

## Problem

The artifact smoke tests cover three formats, but only AppImage actually launches the app. `test-artifact-deb.sh` and `test-artifact-rpm.sh` install the package and assert files exist, then run `--doctor` — but never boot `/usr/bin/claude-desktop`. A startup-only regression like **#666**'s Fedora `SyntaxError` (bad patch anchor → app dies on launch) was invisible to CI on the two formats where it occurred: the rpm test would have stayed green.

## Change

Generalize the AppImage launch smoke test (the #646 readiness-marker poll) into a shared `run_launch_smoke_test` helper, and apply it to all three formats.

- **`test-artifact-common.sh`** — new `run_launch_smoke_test` + `_launch_smoke_cleanup`: boot under `xvfb-run` + `dbus-run-session` in a fresh process group, poll up to 30s for `[Frame Fix] Patches built successfully` (emitted by `frame-fix-wrapper.js` once patches install → main-process startup survived), dump logs on failure, reap the process group. Tool absence is a skip, matching `validate_app_contents`.
- **appimage** — refactored to call the helper; the ~90-line inline copy is gone, so the two can't drift.
- **deb** — launches as the non-root `ubuntu-latest` runner.
- **rpm** — drops to a throwaway unprivileged user. Electron hard-aborts as root without `--no-sandbox` (which the launcher only adds on Wayland/deb), so the root `fedora:42` container must drop privileges; this also exercises the real setuid-`chrome-sandbox` path a Fedora user hits.
- **`test-artifacts.yml`** — install Xvfb/dbus/util-linux/procps on the Fedora job. The deb job already installed the Ubuntu equivalents, which were previously unused — now they're load-bearing.

## Scope / limits

Main-process startup only. GPU/renderer crashes (#583-class) leave the main process alive and pass — Xvfb has no GPU, so Electron falls back to SwiftShader and that path isn't exercised here.

## Validation

- `shellcheck -x` clean on all three artifact scripts; `actionlint` clean on the workflow.
- Full BATS suite green (274/274).
- The launch tests themselves can only run against built packages in CI — this PR's own CI run is the first real exercise. The rpm Fedora package names (`xorg-x11-server-Xvfb`, `dbus-daemon`) were verified against Fedora package metadata.

### DIFF
--- a/.github/workflows/test-artifacts.yml
+++ b/.github/workflows/test-artifacts.yml
@@ -38,7 +38,9 @@ jobs:
 
       - name: Install test dependencies (Fedora)
         if: matrix.format == 'rpm'
-        run: dnf install -y findutils file nodejs npm
+        run: |
+          dnf install -y findutils file nodejs npm \
+            xorg-x11-server-Xvfb dbus-daemon util-linux procps-ng
 
       - name: Install test dependencies (Ubuntu)
         if: matrix.format != 'rpm'
--- a/.github/workflows/test-artifacts.yml
+++ b/.github/workflows/test-artifacts.yml
@@ -49,6 +49,23 @@ jobs:
           sudo apt-get install -y file libfuse2 nodejs npm \
             xvfb dbus-x11 procps
 
+      # Fail loud if a smoke-test tool is missing. Without this guard a
+      # missing/renamed tool turns run_launch_smoke_test into a silent
+      # green skip (it does `pass "$skip"; return`), masking the test.
+      - name: Verify smoke-test tools are present (Ubuntu)
+        if: matrix.format != 'rpm'
+        run: |
+          for t in xvfb-run dbus-run-session setsid; do
+            command -v "$t" >/dev/null || { echo "::error::missing $t"; exit 1; }
+          done
+
+      - name: Verify smoke-test tools are present (Fedora)
+        if: matrix.format == 'rpm'
+        run: |
+          for t in xvfb-run dbus-run-session setsid runuser; do
+            command -v "$t" >/dev/null || { echo "::error::missing $t"; exit 1; }
+          done
+
       - name: Run artifact tests
         run: |
           chmod +x tests/test-artifact-${{ matrix.format }}.sh
--- a/.github/workflows/test-artifacts.yml
+++ b/.github/workflows/test-artifacts.yml
@@ -38,9 +38,19 @@ jobs:
 
       - name: Install test dependencies (Fedora)
         if: matrix.format == 'rpm'
+        # Electron's shared libraries (nss/nspr/gtk3/X11/etc.) must be
+        # installed explicitly: the rpm is installed with `rpm -ivh --nodeps`
+        # and its spec sets `AutoReqProv: no`, so the package declares no
+        # runtime Requires and nothing pulls these in. Without them the
+        # launch smoke test dies with `libnspr4.so: cannot open shared
+        # object file` (exit 127). The Ubuntu runner already carries them.
         run: |
           dnf install -y findutils file nodejs npm \
-            xorg-x11-server-Xvfb dbus-daemon util-linux procps-ng
+            xorg-x11-server-Xvfb dbus-daemon util-linux procps-ng \
+            nss nspr atk at-spi2-atk at-spi2-core cups-libs gtk3 \
+            libdrm mesa-libgbm alsa-lib libX11 libXcomposite libXdamage \
+            libXext libXfixes libXrandr libxcb libxkbcommon pango cairo \
+            libXScrnSaver libXtst libxshmfence
 
       - name: Install test dependencies (Ubuntu)
         if: matrix.format != 'rpm'
--- a/tests/test-artifact-appimage.sh
+++ b/tests/test-artifact-appimage.sh
@@ -8,14 +8,11 @@ script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 source "$script_dir/test-artifact-common.sh"
 
 # Single point of cleanup, set at script scope so any interruption
-# between resource alloc and normal exit is covered.
+# between resource alloc and normal exit is covered. _launch_smoke_cleanup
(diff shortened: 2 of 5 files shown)


sabiut:
Note on CI: the 6 **Build Packages** failures here are **not from this PR** — they're inherited from `main`, which is currently red after the 1.9659.2 upstream bump. The builds abort in `scripts/patches/config.sh`:

```
Patching addTrustedFolder to reject .asar paths...
  [FAIL] addTrustedFolder anchor not found
Failed to inject .asar trusted folder guard
```

The `.asar`-guard patch anchors on the volatile `addTrustedFolder` log string, which changed shape in 1.9659.2. This is exactly what **#666** fixes (retargeting to the stable `async addTrustedFolder(...)` function anchor). `main`'s own latest CI run fails the same way.

This PR touches only the artifact **tests** and the test workflow — no build/patch code. Shellcheck/Codespell/BATS/test-flags are green. Once #666 (or an equivalent `config.sh` fix) lands on `main`, I'll rebase and the package builds + the new deb/rpm launch smoke tests will run for real.

aaddrick (commented):
Hey @sabiut — strong work, and the part I like most isn't the new coverage, it's the refactor. Collapsing ~90 lines of inline AppImage logic into one shared `run_launch_smoke_test` means the three formats can't drift, and the header comment scoping the boundary (main-process only, #583/GPU out) is exactly what the next maintainer needs. Lint's clean across the board and the Fedora package names check out on a real fc42, including the `dbus-daemon`-not-`dbus-x11` detail.

I pushed a few changes rather than round-tripping them (pushed as `9dc4314`):

- **Workflow tool guard.** Right now if `xvfb-run`/`dbus-run-session`/`setsid` (or `runuser` on rpm) ever goes missing — install typo, a future package rename — the test hits the skip path and silently passes green, which defeats the whole point of the PR. Your comment says loud failure belongs at the workflow layer, so I added a `command -v … || exit 1` step after the install to make that real.
- **pkill-reaper comment.** Under the rpm `runuser` path, PAM re-`setsid()`s the app into its own group, so the negative-PID `kill` misses it and the `pkill -f /usr/lib/claude-desktop` sweep is the actual reaper. I confirmed a `sleep` under `setsid runuser` survives the group kill. Added a comment so nobody later "simplifies" the sweep away and starts leaking electron / breaking `userdel -r`.
- **Softened the marker comment.** The marker fires on the first `require('electron')`, before `app.whenReady()`, so it proves "asar loaded, wrapper ran without a SyntaxError" — the #666 class exactly — but a hang after the marker still passes. Reworded the comment to match what it actually proves. A later post-`whenReady` marker would be a nice follow-up.

One thing I checked and it's fine: a failed smoke test does go red. `fail` bumps `_fail_count` and `print_summary` exits 1.

The only real blocker is process, not code. This has never run CI once, so I'd want to see both the deb and rpm jobs actually reach the marker on a real run before stamping it. The sandbox init inside the fedora container is the one link that's runner-dependent, and it's the same dependency on the deb side. Nice refactor — yes from me once it goes green.

---

aaddrick:
CI ran on the pushed commit (`9dc4314`) and we got the green-run answer: deb and appimage validate pass, **rpm validate fails** — and it's a real one the smoke test just earned its keep on.

```
electron: error while loading shared libraries: libnspr4.so: cannot open shared object file
[FAIL] rpm package exited before reaching ready state (exit: 127)
Results: 32 passed, 1 failed
```

The Fedora job installs only the test tools (`dnf install -y findutils file nodejs npm`) into a bare `fedora:42` container and never installs Electron's runtime libs. Before this PR the rpm test never launched Electron, so it never mattered; now that we actually boot it, it dies on `libnspr4.so` (nspr/nss and friends). The deb/appimage jobs pass because the Ubuntu runner already carries those.

So the test is doing exactly what it should — it's the rpm job's environment that's incomplete. Two ways to close it, and I'd lean on the first because it also validates the package's own metadata:

1. Install the built rpm with dep resolution — `dnf install -y ./artifacts/*.rpm` — so the package's auto-generated `Requires` (the ELF `libnspr4.so()` / `libnss3.so()` deps) actually pull the runtime libs. If that *doesn't* pull them, that's a packaging gap worth knowing about on its own.
2. Or explicitly `dnf install -y nss nspr` (plus whatever else the launch needs — `at-spi2-atk`, `libdrm`, `mesa-libgbm`, `gtk3`, `alsa-lib`, `libXtst`, `libXScrnSaver`, …) before the smoke test.

I'd try (1) first and let the failure tell you which deps are missing. Holding my approval until the rpm job goes green — which is the whole point of having added it. Good first catch from the new test.

---

Issue #672, opened 2026-05-29
[bug]: nix build failure on 2ae2172


fdnt7:
### Version (`claude-desktop --doctor` output)

```shell
N/A
```

### What happened

On 2ae2172, the nix flake failed to build in NixOS x86-64.

### Steps to reproduce

```console
$ nix build github:aaddrick/claude-desktop-debian/2ae2172a6007f70e5499b65900db1b7b9a493e42 
```

### Expected behavior

The build should succeed and output a result `symlink` in the working directory.

### Logs / errors

```shell
Running phase: patchPhase
@nix { "action": "setPhase", "phase": "patchPhase" }
Running phase: updateAutotoolsGnuConfigScriptsPhase
@nix { "action": "setPhase", "phase": "updateAutotoolsGnuConfigScriptsPhase" }
Running phase: configurePhase
@nix { "action": "setPhase", "phase": "configurePhase" }
no configure script, doing nothing
Running phase: buildPhase
@nix { "action": "setPhase", "phase": "buildPhase" }
[1;36m--- Architecture Detection ---[0m
Detecting system architecture...
Detected machine architecture: x86_64
Configured for amd64 (x86_64) build.
Target Architecture: amd64
[1;36m--- End Architecture Detection ---[0m
[1;36m--- Distribution Detection ---[0m
Detecting Linux distribution family...
Warning: Could not detect distribution family
  AppImage build will still work, but native packages (deb/rpm) may not
Distribution: 
Distribution family: unknown
[1;36m--- End Distribution Detection ---[0m
Running as user: nixbld (Home: /build)
System Information:
Distribution: 
Distribution family: unknown
Target Architecture: amd64
[1;36m--- Argument Parsing ---[0m
Selected build format: nix
Cleanup intermediate files: no
[1;36m--- End Argument Parsing ---[0m
[1;36m--- Download the latest Claude executable ---[0m
Using local Claude installer: /build/Claude-Setup.exe
Local installer copied to build directory
Extracting resources from Claude-Setup-x64.exe into separate directory...

7-Zip [64] 17.05 : Copyright (c) 1999-2021 Igor Pavlov : 2017-08-28
p7zip Version 17.05 (locale=C,Utf16=off,HugeFiles=on,64 bits,12 CPUs x64)

Scanning the drive for archives:
  0M Scan /build/build/                       1 file, 193374368 bytes (185 MiB)

Extracting archive: /build/build/Claude-Setup-x64.exe
[...]

Issue #673, opened 2026-05-30
[bug]: Not building in ubuntu 24.04.4


rajatpage:
### Version (`claude-desktop --doctor` output)

```shell
none
```

### What happened

Getting following error on running ./build.sh thus blocking me from installing claude in Ubuntu:
Patching addTrustedFolder to reject .asar paths...
  Found folder parameter: i
  [FAIL] addTrustedFolder anchor not found
Failed to inject .asar trusted folder guard

### Steps to reproduce

clone the repo and try to build using build.sh in Ubuntu 24.04.4 LTS

### Expected behavior

It should build and install claude in Ubuntu

### Logs / errors

```shell

```

aaddrick:
Same root cause as #677, and nothing Ubuntu-specific here. The `[FAIL] addTrustedFolder anchor not found` abort came from upstream 1.9659.2 re-minifying the log call the config.sh patch anchored on. It was purely patch revision against bundle bytes, so 24.04.4 wasn't the variable.

Fixed by PR #685, shipped in v2.0.17+claude1.10628.2 and every release since. `git pull` and rebuild from main, or install from the apt repo — either should clear it.

Closing as a duplicate of #677 / fixed.

---

Written by Claude Fable 5 via [Claude Code](https://claude.ai/code)

PR #674, opened 2026-05-31
fix(patches): anchor addTrustedFolder guard on function definition

mhentschke:
The `addTrustedFolder` log call moved from a standalone statement in
1.9255.x to a comma-expression inside an `if` condition in 1.9659.2:

// 1.9255.x — standalone statement, anchor ended with );
ie.debug(LocalAgentModeSessions.addTrustedFolder: ${i});

// 1.9659.2 — log call inside if-condition comma expression
if(S.info(LocalAgentModeSessions.addTrustedFolder: ${i}),await NUe(i)===null){

The old anchor `LocalAgentModeSessions.addTrustedFolder: \${F}\);`
no longer exists, causing `[FAIL] addTrustedFolder anchor not found`
and aborting the build entirely.

Replace with `async addTrustedFolder(F){` — the function definition
itself — which is unique (one occurrence in the bundle), stable across
refactors of the function body, and is the correct semantic injection
point (function entry). Per docs/learnings/patching-minified-js.md §
Anchor selection.

Fixes:
https://github.com/aaddrick/claude-desktop-debian/issues/672
https://github.com/aaddrick/claude-desktop-debian/issues/673

### DIFF
--- a/scripts/patches/config.sh
+++ b/scripts/patches/config.sh
@@ -121,7 +121,7 @@ const p = 'app.asar.contents/.vite/build/index.js';
 const F = process.env.FOLDER_PARAM;
 let code = fs.readFileSync(p, 'utf8');
 
-const anchor = 'LocalAgentModeSessions.addTrustedFolder: \${' + F + '}\`);';
+const anchor = 'async addTrustedFolder(' + F + '){';
 const idx = code.indexOf(anchor);
 if (idx === -1) {
   console.error('  [FAIL] addTrustedFolder anchor not found');


Ondra-Zik:
Confirming that this resolved the build failure with my NixOS Nix flake.

FloLangenfeld:
This fix works for me as well (thanks @mhentschke).

aaddrick:
@mhentschke you had the correct read first. The old anchor matched the trailing log-string `` `); ``, and 1.10628.x folded that log call into a comma-expression inside the `if`, so it disappeared. Re-anchoring on `async addTrustedFolder(` is exactly the right fix, and @Ondra-Zik and @FloLangenfeld both confirmed it unblocks their builds.

We ended up going with #685 because it also moves the parameter-extraction grep off the `LocalAgentModeSessions.addTrustedFolder:` log literal onto the method name. #674 works today, but leaving extraction on the log string is the same class of incidental anchor that just broke on us. #685 keys both the grep and the injection point off the same `addTrustedFolder` literal, so they can't drift apart if upstream drops the log line entirely later.

Closing this as superseded by #685. Credit for the first correct diagnosis is yours though. Want me to add you to the README acknowledgments?

Issue #604, opened 2026-05-12
[bug]: tray icon invisible on LMDE 7 / Cinnamon — Mint-Y-Dark-Aqua dark panel + light colour-scheme


IvanTheGeek:
### Version (`claude-desktop --doctor` output)

```shell
Claude Desktop Diagnostics
================================

[PASS] Installed version: 1.6608.2-2.0.10
[PASS] Display server: X11 (DISPLAY=:0)
       Menu bar mode: auto (default, Alt toggles visibility)
       Titlebar style: hybrid (default, native frame + in-app topbar)
[PASS] Electron: v41.5.1 (/usr/lib/claude-desktop/node_modules/electron/dist/electron)
[PASS] Chrome sandbox: permissions OK (/usr/lib/claude-desktop/node_modules/electron/dist/chrome-sandbox)
[PASS] SingletonLock: held by running process (PID 19667)
[PASS] Node.js: v24.15.0
[PASS] Desktop entry: /usr/share/applications/claude-desktop.desktop

All checks passed.
```

### What happened

On LMDE 7 (gigi) with Cinnamon 6.6.7, the tray icon renders as the black `TrayIconTemplate.png` against the panel's dark background, making it effectively invisible / mis-rendering as a dark blob next to my other tray icons.

This is the same failure pattern described in #588 (XFCE), #557 (Linux Mint Cinnamon X11), #563 (KDE Wayland) and #157 (closed, Fedora KDE 6) — `nativeTheme.shouldUseDarkColors` returning `false` while the panel is dark by design.

### Steps to reproduce

1. Install LMDE 7 with defaults (or Linux Mint Cinnamon with the default Mint-Y-Dark-Aqua Cinnamon theme + Mint-Y-Aqua GTK theme)
2. Install `claude-desktop` (1.6608.2-2.0.10 in my case)
3. Launch — observe the tray icon disappears into the panel

### Expected behavior

A visible tray icon: the white `TrayIconTemplate-Dark.png` should be used on this configuration because the panel is dark.

### Logs / errors

Not log-relevant — this is detection-heuristic behaviour rather than an error.

### Anything else

**Why I'm filing instead of just submitting a PR**

PR #491 already explored a user-facing Auto/Black/White submenu for this exact class of bug and was closed; the maintainer's stated preference is to keep the choice automatic, not give users a manual override. Before writing a patch I'd like to confirm direction.

**Why LMDE/Cinnamon is solvable separately from the general case**

[...]

PR #680, opened 2026-06-02
fix(tray): startup icon stuck black — make rebuild mutex trailing-edge (#679)

### LINKED ISSUE #679: [bug]: tray icon stuck black at startup on dark GNOME — corrective nativeTheme "updated" dropped by the rebuild mutex (distinct from #604/#563)
## Summary

On GNOME (dark mode), the tray icon comes up as the black `TrayIconTemplate.png` on the dark panel and stays invisible until the user manually toggles Claude's in-app appearance. Unlike the existing "dark panel" reports, here the OS theme **is** dark and `shouldUseDarkColors` **does** resolve to `true` within ~50 ms of startup — the correct value is just being **dropped** before it reaches the tray.

## Why this is NOT a duplicate of #604 / #563 / #588 / #157

The open "black tray icon" reports share a symptom but have a different root cause. #604 groups #588, #557, #563, #157 under one pattern: *"`shouldUseDarkColors` returning `false` while the panel is dark **by design**"* — i.e. the OS colour scheme is **light** but the panel is dark (KDE per-panel themes, Cinnamon Mint-Y-Dark-Aqua + light GTK). There `shouldUseDarkColors` is **persistently** `false`; panel-colour detection or a manual override is the fix.

This report is the opposite, and was **measured**, not inferred:

| | #604 family (divergence) | This issue (startup race) |
|---|---|---|
| OS `color-scheme` | `light` (panel dark by design) | **`prefer-dark`** (system genuinely dark) |
| `shouldUseDarkColors` | stays `false` forever | **`false` for ~50 ms, then `true`** |
| Nature | persistent, theme-dependent | transient, **startup-timing only** |
| Fixed by in-app toggle? | no (value still wrong) | yes (forces `themeSource`) |
| Fix | detect panel colour / override | don't drop the corrective rebuild |

A standalone Electron probe in this exact session:

```
[ready+0ms]     shouldUseDarkColors=false   <- tray created here -> black
[UPDATED-EVENT] shouldUseDarkColors=false
[UPDATED-EVENT] shouldUseDarkColors=true    <- correct value arrives ~50-100 ms in
[ready+500ms]   shouldUseDarkColors=true     (stays true)
```

So the signal is correct almost immediately. The icon is wrong because the `tray_func._running` mutex (added in `scripts/patches/tray.sh`) is a *leading-edge* throttle: the first `"updated"` (still `false`) takes the 1500 ms lock and renders black, and the follow-up `"updated"` events (`true`) all arrive inside that window and are **dropped**. No further event fires on its own, so the icon stays black until a manual theme change forces a new `"updated"`.

A 4-second simulation of the exact throttle confirms it: leading-edge -> final icon BLACK; trailing-edge (re-apply the last value) -> final icon WHITE.

**Why a separate issue rather than a comment on #604:** different root cause -> different fix. The panel-detection / manual-override work proposed for the divergence family does not address this (the value here is already correct), and this fix (making the rebuild throttle trailing-edge) does not resolve the divergence cases (where the value is genuinely `false`). Filing separately keeps the two fixes traceable and avoids symptom-based triage merging them.

Secondary finding surfaced while debugging: on this upstream version `patch_tray_inplace_update` silently skips (its `setContextMenu(BUILDER())` extraction no longer matches upstream's `M=BUILDER(); setContextMenu(M)` shape), so the #515 in-place fast-path is absent — the duplicate-icon race it fixed has quietly regressed. The accompanying PR restores it.

## Steps to reproduce

1. GNOME with dark mode (`color-scheme = prefer-dark`), default `gtk-theme` Adwaita
2. Launch Claude Desktop
3. Tray icon renders black on the dark panel and stays so until you change Claude's in-app appearance

## Expected behavior

The white `TrayIconTemplate-Dark.png` at startup, since the system is dark — with no manual toggle.

## Environment

- Claude Desktop 1.9255.2, Electron 41.5.0
- Fedora Linux 44 (Workstation Edition), GNOME, Wayland session (Electron via XWayland, default `--ozone-platform=x11`); also reproduced under `--ozone-platform=wayland`
- `color-scheme=prefer-dark`, `gtk-theme=Adwaita`

---

LiukScot:
## Summary

Fixes the tray icon coming up **black (invisible) on a dark panel at startup** and staying that way until the user manually toggles Claude's in-app theme. Fixes #679.

This is **not** the #604/#563 "panel dark by design / `shouldUseDarkColors` stuck false" family — here the system genuinely is dark and the signal *does* become correct within ~50 ms; the correct value is just dropped before it reaches the tray. See #679 for the full distinction.

## Root cause (measured, not inferred)

A standalone Electron probe in the affected session:

```
[ready+0ms]     shouldUseDarkColors=false   <- tray created here -> black
[UPDATED-EVENT] shouldUseDarkColors=false
[UPDATED-EVENT] shouldUseDarkColors=true    <- correct value ~50-100 ms in
[ready+500ms]   shouldUseDarkColors=true     (stays true)
```

`shouldUseDarkColors` reads `false` for the first ~50 ms, then a burst of `nativeTheme "updated"` events flips it `true`. The tray is created with the transient `false` (black). The rebuild function's mutex was a **leading-edge** throttle: the first `"updated"` (false) takes the 1500 ms lock and renders black; the corrective `"updated"` (true) events all arrive inside that window and are **dropped**. No further event fires on its own, so the icon stays black until a manual theme change.

A 4-second simulation of the exact throttle confirmed it: leading-edge → final icon BLACK; trailing-edge → final icon WHITE.

## The fix (`scripts/patches/tray.sh`)

1. **Trailing-edge mutex** (`patch_tray_menu_handler`) — a request arriving while a rebuild is in flight is remembered (`_pending`) and re-run once when the window clears, so the final `nativeTheme` value wins instead of being dropped.
2. **Removed the `_trayStartTime > 3e3` startup-suppression window** — it gated the very `"updated"` → rebuild call the correction now relies on.
3. **Restored the in-place `setImage` fast-path** (#515): on current upstream `patch_tray_inplace_update` silently skipped because its `menu_func` extraction assumed `setContextMenu(BUILDER())`, but upstream now does `M=BUILDER(); setContextMenu(M)`. The fallback resolves the builder one hop back (so the injected code stays `setContextMenu(BUILDER())`, never `setContextMenu(OBJECT())`), using a word-boundary lookbehind that tolerates separator **and** `let`/`const`/`var` declarator shapes. Without this the #515 duplicate-icon race was quietly regressed on current releases.

Docs: `docs/learnings/tray-rebuild-race.md` gains a "Startup icon-colour race" section and its stale 3 s-window pitfall is corrected. CHANGELOG `[Unreleased]` updated.

## Trade-off

The trailing re-run lands when the mutex window clears, so there is a **~1.5 s black flash** at startup before the icon corrects (vs. permanently black before). Happy to tighten the window if you'd prefer a shorter flash — left at the existing 1500 ms here to keep the change minimal.

## Test plan

- [x] Built an AppImage from this branch on Fedora 44 / GNOME / Wayland (Electron via XWayland) and confirmed the tray icon now comes up **white/visible** on the dark panel without any manual toggle (previously permanently black).
- [x] Verified the patched `app.asar`: no `_trayStartTime` gate, `setImage` fast-path present, trailing-edge mutex (`G9A._pending`) present.
- [x] `bash -n` clean; `shellcheck -x` shows only the pre-existing sourced-fragment notes (no shebang / injected globals) that the CI's `git grep` filter excludes by design — no new findings.
- [x] Word-boundary fallback regex validated against `,;({` separators and `let`/`const`/`var` declarators.
- [ ] Maintainer: full pipeline / CI (note: the current upstream `1.9659.2` bundle trips an **unrelated** pre-existing `addTrustedFolder` patch drift in `config.sh`, independent of this change).

---
85% AI / 15% Human
Claude: measured the root cause (Electron probe + throttle simulation), wrote the three patch changes and the docs, built and runtime-verified the AppImage, drafted the issue and PR.
Human: reported the bug, supplied the decisive observations (manual-toggle-fixes-it; "still black" feedback that twice corrected the diagnosis), and approved the approach and scope.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,8 @@ The format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) — 
 
 ### Fixed
 
-- Tray icon no longer stuck black at startup on dark desktops. `nativeTheme.shouldUseDarkColors` reads `false` for the first ~50 ms then flips `true`, but the leading-edge rebuild mutex latched the transient `false` and dropped the corrective `"updated"` events; the mutex is now trailing-edge (re-applies the final value) and the obsolete 3 s startup-suppression window was removed. (__PRURL__, fixes [#679](https://github.com/aaddrick/claude-desktop-debian/issues/679))
-- Restored the in-place tray `setImage` fast-path ([#515](https://github.com/aaddrick/claude-desktop-debian/pull/515)), which silently stopped applying after upstream changed the context-menu wiring from `setContextMenu(BUILDER())` to a prebuilt `setContextMenu(MENU)` object — `patch_tray_inplace_update` now resolves the builder in both shapes, so the duplicate-icon SNI race no longer regresses. (__PRURL__)
+- Tray icon no longer stuck black at startup on dark desktops. `nativeTheme.shouldUseDarkColors` reads `false` for the first ~50 ms then flips `true`, but the leading-edge rebuild mutex latched the transient `false` and dropped the corrective `"updated"` events; the mutex is now trailing-edge (re-applies the final value) and the obsolete 3 s startup-suppression window was removed. ([#680](https://github.com/aaddrick/claude-desktop-debian/pull/680), fixes [#679](https://github.com/aaddrick/claude-desktop-debian/issues/679))
+- Restored the in-place tray `setImage` fast-path ([#515](https://github.com/aaddrick/claude-desktop-debian/pull/515)), which silently stopped applying after upstream changed the context-menu wiring from `setContextMenu(BUILDER())` to a prebuilt `setContextMenu(MENU)` object — `patch_tray_inplace_update` now resolves the builder in both shapes, so the duplicate-icon SNI race no longer regresses. ([#680](https://github.com/aaddrick/claude-desktop-debian/pull/680))
 
 ## [v2.0.16] — 2026-05-27
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ The format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) — 
 
 <!-- Updated automatically by check-claude-version; will be current at release time. -->
 
+### Fixed
+
+- Tray icon no longer stuck black at startup on dark desktops. `nativeTheme.shouldUseDarkColors` reads `false` for the first ~50 ms then flips `true`, but the leading-edge rebuild mutex latched the transient `false` and dropped the corrective `"updated"` events; the mutex is now trailing-edge (re-applies the final value) and the obsolete 3 s startup-suppression window was removed. (__PRURL__, fixes [#679](https://github.com/aaddrick/claude-desktop-debian/issues/679))
+- Restored the in-place tray `setImage` fast-path ([#515](https://github.com/aaddrick/claude-desktop-debian/pull/515)), which silently stopped applying after upstream changed the context-menu wiring from `setContextMenu(BUILDER())` to a prebuilt `setContextMenu(MENU)` object — `patch_tray_inplace_update` now resolves the builder in both shapes, so the duplicate-icon SNI race no longer regresses. (__PRURL__)
+
 ## [v2.0.16] — 2026-05-27
 
 Tracks upstream Claude Desktop 1.9255.0.
--- a/docs/learnings/tray-rebuild-race.md
+++ b/docs/learnings/tray-rebuild-race.md
@@ -80,7 +80,7 @@ releases. All five are extracted dynamically in `tray.sh`:
 | `tray_func` | `on("menuBarEnabled",()=>{ … })` |
 | `tray_var` | `});let X=null;(async )?function ${tray_func}` |
 | `electron_var` | already extracted earlier in `_common.sh` |
-| `menu_func` | `${tray_var}.setContextMenu(X(`  |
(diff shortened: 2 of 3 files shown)


aaddrick (commented):
Solid work, and I appreciate that you measured the race instead of inferring it. The leading-vs-trailing simulation made the diagnosis easy to verify. The logic holds up when I traced `E6A()`. It's synchronous upstream, so `_running` really does latch the transient black value, and the trailing re-run re-reads the theme fresh.

Two things I'd want confirmed before merge. Both because the testing so far was on hand-built fixtures and a 1.9255.2 build, not current upstream.

1. Run the full patch against the real current `index.js` and confirm `_trayStartTime` is fully gone and the fast-path marker lands once. The reference bundle doesn't even contain the prebuilt `setContextMenu(M)` shape the fallback targets, so "passes on fixtures" isn't quite "passes on upstream."
2. Flip the KDE theme a handful of times quickly and confirm the icon settles and no duplicate appears. The trailing re-run arms a fresh 1.5 s window each time, so steady-state bursts can lag the final icon up to ~1.5 s. That's great at startup, but it's unverified under active toggling.

One small change I pushed (pushed as `2ad33d3`): when both the inline `menu_func` grep and the new fallback come up empty, it now warns on stderr instead of a quiet skip. A silent skip there is how the #515 race regressed last time, so it's worth making loud in CI.

Heads-up that `main` cut v2.0.17 today. Give the CHANGELOG a glance on rebase to keep the entry under `[Unreleased]`. Nice fix overall.

---

PR #682, opened 2026-06-03
Clean up Desktop helpers after explicit quit

svankirk:
## Summary

- keep the Linux launcher alive while Electron runs, so explicit quit can run post-exit cleanup
- add stale-helper cleanup for known Claude Desktop-owned Cowork, user-data, installed Electron helper, and Claude Extension processes when no live UI exists
- wire deb, RPM, AppImage, and Nix launchers through the shared cleanup runner
- add BATS coverage for safe helper matching and post-exit cleanup preserving Electron exit status

## Notes

This preserves close-to-tray semantics: cleanup only runs after the Electron process exits. Closing the window to tray keeps the app and helpers running.

## Verification

- `npx --yes bats tests/launcher-common.bats`
- `npx --yes shellcheck -x scripts/launcher-common.sh scripts/doctor.sh scripts/packaging/deb.sh scripts/packaging/rpm.sh scripts/packaging/appimage.sh`
- `git diff --check`

---
Generated with OpenAI Codex via Codex CLI
Co-Authored-By: OpenAI Codex <codex@openai.com>
85% AI / 15% Human
Codex: investigation, patch, tests, verification, PR preparation
Human: reported Linux Desktop quit cleanup behavior and clarified close-to-tray vs explicit quit semantics

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ The format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) — 
 
 <!-- Updated automatically by check-claude-version; will be current at release time. -->
 
+### Fixed
+
+- Explicit quit now keeps the launcher alive until Electron exits, then runs
+  stale-helper cleanup for Desktop-owned Cowork, Claude config, and extension
+  helpers. Close-to-tray still leaves the app and helpers running.
+
 ### Changed
 
 - The native-Wayland launch path now routes Quick Entry's global shortcut (`Ctrl+Alt+Space`) through the XDG GlobalShortcuts portal: `GlobalShortcutsPortal` is added to the `--enable-features` set, and all Chromium feature requests are merged into a single `--enable-features=` switch (Chromium honours only the last one, so the previous code could silently clobber features). GNOME Wayland users can opt into the portal route with `CLAUDE_USE_WAYLAND=1`, which works on GNOME ≤ 49 after a one-time portal permission dialog and fixes the focus-bound hotkey from [#404](https://github.com/aaddrick/claude-desktop-debian/issues/404). The default GNOME session stays on XWayland (no rendering/IME regression risk); auto-selecting native Wayland on GNOME is deferred until it can be gated on a real render check. **On GNOME 50 / xdg-desktop-portal ≥ 1.20 the portal route is currently a no-op** — Electron/Chromium doesn't perform the portal's new host `Registry.Register` app-id handshake (filed upstream as [electron/electron#51875](https://github.com/electron/electron/issues/51875)). `CLAUDE_USE_WAYLAND` is now tri-state: `1` native Wayland, `0` force XWayland, unset auto-detects. ([#404](https://github.com/aaddrick/claude-desktop-debian/issues/404))
--- a/nix/claude-desktop.nix
+++ b/nix/claude-desktop.nix
@@ -239,6 +239,7 @@ fi
 setup_logging || exit 1
 setup_electron_env
 cleanup_orphaned_cowork_daemon
+cleanup_stale_desktop_helpers
 cleanup_stale_lock
 cleanup_stale_cowork_socket
 
@@ -265,10 +266,11 @@ build_electron_args 'nix'
 # Add app path
 electron_args+=("$app_path")
 
-# Execute Electron (exec replaces the shell process so signals
-# like SIGINT, SIGTERM, and SIGHUP reach Electron directly)
+# Execute Electron and keep the launcher alive so explicit quit can
+# clean up Desktop-owned helpers that outlive the Electron main process.
 log_message "Executing: $electron_exec ''${electron_args[*]} $*"
-exec "$electron_exec" "''${electron_args[@]}" "$@" >> "$log_file" 2>&1
+run_electron_and_cleanup "$electron_exec" "''${electron_args[@]}" "$@"
+exit $?
 LAUNCHER
     # Substitute placeholders — electron_exec points to our custom
     # wrapper (which sets GTK/GIO env then execs our merged binary)
--- a/nix/claude-desktop.nix
+++ b/nix/claude-desktop.nix
@@ -226,6 +226,7 @@ stdenvNoCC.mkDerivation {
 
 electron_exec="ELECTRON_PLACEHOLDER"
 app_path="RESOURCES_PLACEHOLDER/app.asar"
+claude_desktop_app_path="$app_path"
 
 source "LAUNCHER_LIB_PLACEHOLDER"
 
--- a/scripts/launcher-common.sh
+++ b/scripts/launcher-common.sh
@@ -302,6 +302,21 @@ build_electron_args() {
 	fi
 }
 
+_claude_desktop_ui_cmdline_matches() {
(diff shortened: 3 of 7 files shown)


svankirk:
Follow-up hardening after reviewing cross-app behavior: the live-UI detector now uses the exact Claude `app.asar` path exported by each launcher, instead of treating any Electron `app.asar` process as a live Claude UI. Added BATS coverage that a non-Claude Electron `app.asar` is ignored.

This prevents other Electron apps from causing stale Claude helper cleanup to be skipped.

aaddrick (commented):
Hey @svankirk — this is genuinely good work and the testing writeup made it easy to follow. I pulled the branch: shellcheck `-x` clean, 80/80 bats green, heredoc escaping correct. The dedup of the live-UI check keyed on the exact `app.asar` path is a real correctness win.

Two things before this lands, and one scope question for @aaddrick.

The `wait` in `run_electron_and_cleanup` returns early when a forwarded signal fires. Bash hands back `128+signum` while the child's still alive (reproduced with a TERM-ignoring stub). That undercuts the "launcher outlives and reaps Electron" guarantee. I pushed a fix, but with one deliberate choice worth calling out: I used the signal-status-preserving loop, capturing status on the first `wait` and re-`wait`ing only to reap, so the launcher still exits 143 on TERM like the old `exec` did rather than overwriting it with Electron's real code. I also pushed a small reorder so `_electron_child_pid` is cleared after cleanup. Otherwise the self-kill guard in the cleanup functions is comparing against an already-empty var and can never match. (Both pushed as `45f612c`.)

The bigger one I did not push: the MCP cgroup gate greps `/app-claude-desktop-*.scope`, which matches KDE-style scope names but not GNOME's `app-gnome-claude-desktop-*.scope`. So the whole MCP-cleanup slice is effectively dead code on GNOME. That gate is the only thing keeping the broad `*-mcp*` match safe, so it needs to be DE-agnostic and tested on both GNOME and KDE before the MCP slice can be trusted.

Which ties into scope. The helper cleanup (cowork, user-data, extensions, Chromium helpers) is a clean Linux-process-model fix and I think it's in scope. The MCP-process slice is the one genuinely new thing, since the launcher's never reached into user-spawned MCP servers, and it's both net-new and currently mis-gated. I'd rather see it split out and landed on its own once the gate is proven. @aaddrick, that MCP slice is your call. Nice work on the rest.

---

aaddrick (changes requested):
Marking this Request Changes to track the split — detail is in my review above, short version here.

The helper-cleanup half (cowork, user-data, extensions, `/usr/lib/claude-desktop` Chromium helpers) is in scope and close to done; I pushed the `wait` reap fix and the self-kill-guard reorder (`45f612c`).

Two things block the MCP-process slice as written:

1. The cgroup gate greps `/app-claude-desktop-*.scope`, which doesn't match GNOME's `app-gnome-claude-desktop-*.scope` — so the slice is dead code on GNOME, and that gate is the only thing keeping the broad `*-mcp*` match safe. It needs to be DE-agnostic and tested on GNOME + KDE.
2. Reaching into user-spawned MCP servers is net-new scope for the launcher. I'd rather land the helper cleanup on its own and split the MCP slice out until the gate is proven.

Happy to take the helper half as soon as it's carved out. Nice work on the rest.

---

sabiut:
Reviewed from the test side (`tests/launcher-common.bats` is the part in my lane). Pulled the branch — 80/80 bats green, `shellcheck -x` clean, and the matcher coverage is good: `_desktop_helper_cmdline_matches` correctly leaves the live-UI `app.asar` process unmatched, and `_claude_desktop_ui_cmdline_matches` covers the exact-path dedup nicely.

Two coverage gaps I'd want closed, both tied to things already raised above:

**1. The signal-path reap fix (45f612c) isn't tested.** The only `run_electron_and_cleanup` test uses a stub that exits immediately, so it never exercises the `while kill -0; do wait; done` loop — i.e. the exact behavior that fix corrects. I ran a TERM-ignoring stub locally and confirmed the fix holds (launcher exits 143, child gets reaped, cleanup still runs), but there's nothing in the suite pinning it, so it can regress silently. Could we add a test with a TERM-ignoring Electron stub asserting exit status 143 + that the child is reaped + cleanup ran?

**2. `_pid_in_claude_desktop_scope` has no test.** That gate is the only thing keeping the broad `*-mcp*` match safe, and it's the same gate flagged as GNOME-broken above. The matcher tests pass in isolation, which gives the MCP slice a false green. If/when the MCP slice is split out and the gate is made DE-agnostic, it'll need coverage for both `app-claude-desktop-*.scope` (KDE) and `app-gnome-claude-desktop-*.scope` (GNOME).

Agree with splitting the MCP slice out — the helper-cleanup half looks close once the 143-reap test lands. Nice work.

sabiut:
Took a look from the test side before the respin. Suite is green at `45f612c` (80/80), but I poked at the matchers with some fixture cmdlines and found a problem in the helper-cleanup half too.

`_desktop_helper_cmdline_matches` substring-matches anywhere in argv, with no scope gate on that branch. So all of these come back as kill targets:

```
vim ~/.config/Claude/Claude Extensions/foo/manifest.json
tail -f ~/claude-desktop-debian/scripts/cowork-vm-service.js
grep -r pattern ~/.config/Claude/Claude Extensions/
rsync -a ~/.config/Claude/Claude Extensions/ /backup/
```

Quit the app while editing one of those files and your editor gets SIGTERM, then SIGKILL two seconds later. It's the same broadness issue as #592, just with `kill` instead of `pkill`. The match needs to anchor on what the process is (exe/argv[0]), not what its arguments happen to mention.

Also checked the cgroup gate against real systemd naming, and it's worse than the GNOME miss: KDE registers apps as `app-<id>@<random>.service` (`.service`, not `.scope`), and terminal launches land in `vte-spawn-*.scope`. So the gate is dead on GNOME, probably dead on KDE, and dead for terminal launches.

For the respin, the tests should pin both: table-driven fixtures for the scope-name variants, and negative cases asserting the bystanders above are spared.

aaddrick (approved):
Hey @svankirk! Closing this out rather than asking for a third round. You addressed the earlier code points, so the only open items were the GNOME gate and the split. I pushed the split myself (head `9648416`). The PR now ships the solid half: the quit wrapper, signal forwarding, the bundle-keyed live-UI matcher, and the four safe helper arms, all `pgrep -u "$(id -u)"` scoped. That half is genuinely yours and lands with your credit.

The MCP slice moved to #709. The gate turned out worse than "dead on GNOME": GNOME prefixes units with `app-gnome-`, and systemd escapes the hyphens, so KDE renders the unit as `app-claude\x2ddesktop...` with `.service` variants. The literal `app-claude-desktop-*.scope` grep matches neither. Safe by construction since it's an AND, but invisible. It needs real cgroup captures from both desktops before any regex gets trusted, and #709 spells out the acceptance criteria.

Housekeeping while I was in there: rebased twice over today's merge train (#690 through #695 landed underneath) and reconciled #693's reaper tests with the new matcher. 299/299 BATS green, shellcheck clean.

Approving; merge follows green CI. Thanks for sticking with this across the rounds. If you want the MCP slice, it's welcome as a follow-up PR against #709.

---

PR #687, opened 2026-06-04
fix(deb): auto-install AppArmor userns profile to stop Ubuntu 24.04+ launch crash

diarized:
## Problem

Ubuntu 24.04+ sets `apparmor_restrict_unprivileged_userns=1` (distro default), blocking the unprivileged user namespaces Chromium's sandbox needs.

- **Affected:** deb on **X11 only.** `scripts/launcher-common.sh` adds `--no-sandbox` for AppImage and any Wayland session, so deb-on-Wayland never reaches this path. Real-world slice ≈ NVIDIA-on-Ubuntu (still defaults to Xorg).
- **Symptom:**
  ```
  FATAL:sandbox/linux/services/credentials.cc:131] Check failed: . : Permission denied (13)
  Trace/breakpoint trap (core dumped)   # exit 133
  ```
- **Not the cause:** the SUID `chrome-sandbox` helper (4755). Chromium still spins an unprivileged userns zygote that AppArmor denies.

## Fix

Mirror the `google-chrome`/`code`/`slack` debs (`/etc/apparmor.d/*` all use `flags=(unconfined) { userns }`): grant `userns` to our Electron binary via a scoped profile. Sandbox stays **enabled** — never `--no-sandbox` on the deb.

- **`scripts/packaging/deb.sh`**
  - `postinst`: generate + load `/etc/apparmor.d/<pkg>`. Gated on the `apparmor_restrict_unprivileged_userns` kernel knob (existence — 0 may flip to 1) so non-Ubuntu systems that ship `apparmor_parser` but never restrict get nothing. Write is abort-proofed: a read-only/atypical `/etc` cannot fail the install under `set -e`.
  - `postrm`: unload + delete on `remove`/`purge`/`abort-install` (generated file, not dpkg-tracked; skips upgrade, where the incoming `postinst` re-applies).
- **`scripts/doctor.sh`** — **User namespaces** check. Silent unless the restriction is in force. Reports the **loaded** state (reads the kernel profile set when root) rather than mere file presence; non-root reports on-disk presence only, never a false PASS.
- **`docs/troubleshooting.md`** — automatic behavior + manual fallback; distro-agnostic fix hints.
- **`CHANGELOG.md`** — `[Unreleased] → Fixed`.

## Why a profile, not `--no-sandbox`

`--no-sandbox` permanently disables the Chromium renderer sandbox for remote web content — the larger attack surface. The scoped profile re-permits `userns` for one trusted binary; every other program on the host stays restricted. Matches the repo's intent (deb keeps the sandbox) and Ubuntu's per-binary profile design.

## Verification (Ubuntu 24.04.4 LTS, Electron 41.5.0, knob=1)

End-to-end on real hardware: build → install (profile loaded) → `--doctor` (all branches, root + non-root) → `remove`/`purge` (profile unloaded + deleted) → reinstall → **GUI launch, no `credentials.cc` crash.** `shellcheck` clean on both scripts; `dash -n` + `shellcheck -s sh` clean on the generated `postinst`/`postrm`.

## Scope / not touched

- **RPM/Fedora** — SELinux, not AppArmor; no such restriction.
- **AppImage / deb-on-Wayland** — already `--no-sandbox`; unaffected.
- **Nix** — separate packaging path; not touched.

---
Generated with everybody knows who.


sabiut:
Reviewed the parts in my lane — `scripts/doctor.sh` and `docs/troubleshooting.md`.

The doctor probe is good and mirrors the existing bwrap probe nicely. Two things:

1. The check is on the profile **file existing** (`[[ -e $_aa_profile ]]`), not on it being **loaded** into the kernel. A staged-but-unloaded or stale profile would report `PASS` while the app still crashes on launch. I think file-existence is a reasonable proxy for a doctor probe (loaded-state really needs root / `aa-status`), but a one-line comment saying so would stop it being mistaken for a liveness check.

2. The fix hint says "reinstall the `.deb`" — which ties into the bigger thing below.

**Coverage gap:** the auto-install is `.deb`-only, since this only touches `deb.sh`. RPM and Nix users hit the same launch crash with no automated profile, and the doctor hint pointing them at "reinstall the `.deb`" doesn't apply. The manual docs do cover them, so it's not blocking — but I'd suggest (a) softening the doctor hint to not assume `.deb` (e.g. "reinstall the package, or apply the profile manually — see docs/troubleshooting.md"), and (b) a follow-up to add the same profile install to `rpm.sh`.

The deb postinst/prerm logic reads correctly to me — `$package_name` expands at build time, `\$APPARMOR_PROFILE` stays literal for install time, prerm correctly skips upgrades, and the `-Q` parse-check before load is a nice touch. Profile content + revert path look right.

aaddrick:
@diarized This is a clean diagnosis. The repro is solid, and granting userns to our own binary through a scoped profile is the right call. It matches exactly what google-chrome, code, and slack ship, and it keeps the sandbox on instead of reaching for `--no-sandbox`. Good instinct.

One scope note worth folding into the framing. The crash only reaches deb users on an X11/Xorg session. On Wayland the deb already launches with `--no-sandbox` (launcher-common.sh adds it for deb/nix on any Wayland session), so deb-on-Wayland never hits this path. The real-world slice is mostly NVIDIA-on-Ubuntu-24.04 folks, who still tend to land on Xorg by default. So it's narrower than "crashes the main app on launch," and the PR description should say so.

Before I merge, a few things need addressing:

1. **The postinst can fail the install.** It runs `set -e`, but the new `mkdir -p /etc/apparmor.d` and the `cat > "$APPARMOR_PROFILE"` heredoc are unguarded, while every other fallible op in that script is `|| echo`. On a read-only or atypical `/etc`, either one aborts the whole postinst and dpkg reports failed-to-configure. Wrap the block so it can't take down the install.
2. **Scope the install to the kernel knob.** Gate on `[[ -e /proc/sys/kernel/apparmor_restrict_unprivileged_userns ]]` instead of just `command -v apparmor_parser`. That knob is Ubuntu-family-specific. As written, stock Debian ships apparmor_parser but never imposes the restriction, so it gets an unconfined-grant profile plus a per-boot AppArmor reload it never needs. Gating on knob-existence still covers the "knob is 0 now, flips on later" case.
3. **Drop the doctor probe, or make it non-prescriptive.** It PASSes on the profile file *existing*, not on it being loaded, so it hands a false PASS to the exact person debugging a launch crash. And it WARNs "reinstall the .deb" for AppImage and deb-on-Wayland users who can't hit this crash at all. Computing "is this crash reachable" after launch is genuinely hard, and I'd rather ship no check than one that misleads.
4. **Move cleanup from prerm to postrm.** That way it also fires on `purge` and `abort-install`, which is the idiomatic shape.

The minimal version I'd merge is the abort-proof, knob-scoped profile install plus postrm cleanup plus a short troubleshooting paragraph. That last part absorbs #686, so once this lands I'll close #686 and credit you on both. Happy to take it from there if you'd rather hand it off.

---

diarized:
@sabiut:
* "the check is on the profile file existing ([[ -e $_aa_profile ]]), not on it being loaded into the kernel" - definitely something to improve
* "the fix hint says "reinstall the .deb" - doctor should be distro-agnostic, sure thing

@aaddrick - valid points, on my way to punish my Claude ;-)

diarized:
Addressed in `af05e3d` (pushed). Per item:

**@aaddrick**
1. **postinst can abort install** — fixed. `mkdir` + heredoc moved into an `if`-condition (commands there are exempt from `set -e`); a failed write hits an `else` warning and the script still reaches `exit 0`. Verified by forcing a write failure → "handled gracefully", exit 0.
2. **Gate on the knob** — done. Now requires `[ -e /proc/sys/kernel/apparmor_restrict_unprivileged_userns ]` alongside `apparmor_parser`. Stock Debian (knob absent) gets no profile and no per-boot reload. Gates on *existence* so a later 0→1 flip is still covered.
3. **Doctor false PASS** — rebuilt. Silent unless the knob reads `1`; reports the actual **loaded** state, never file-existence; no `.deb`-specific advice (see sabiut #2).
4. **prerm → postrm** — moved. `case remove|purge|abort-install`; `upgrade` falls through (incoming postinst reloads). `disappear` intentionally unhandled — matching it would also clean during overwrite-by-another-package.

**Scope framing** — description updated: the crash reaches deb on **X11 only**. `launcher-common.sh` already adds `--no-sandbox` for AppImage + any Wayland session, so the real slice is ~NVIDIA-on-Ubuntu-Xorg.

**@sabiut**
1. **Loaded vs on-disk** — doctor now reads `/sys/kernel/security/apparmor/profiles` and greps the loaded set (root path); non-root can't read it, so it reports on-disk presence only as a neutral info line, never a PASS.
2. **Distro-agnostic hint** — dropped "reinstall the `.deb`"; all hints point at `docs/troubleshooting.md` → "Claude Desktop crashes immediately on launch", which also serves AppImage/Nix-on-Ubuntu.
3. **deb-only / rpm follow-up** — kept out of scope deliberately: the Ubuntu fix must stay invisible to RHEL/Nix, and a stacked PR builds on this branch. Knob-gating makes it a no-op everywhere else.

**Extra fix found while testing on real hardware (same commit):** the first loaded-state check gated on `[[ -r $profiles ]]`. securityfs marks that file `0444`, so `-r` passes for non-root — but the kernel still denies the `read()` without `CAP_MAC_ADMIN`, so the probe falsely reported a *loaded* profile as "not loaded". Switched to judging by whether the read actually returns data.

**Validated end-to-end** (Ubuntu 24.04.4, knob=1): build → install (profile loaded) → `--doctor` all four branches, root + non-root → `remove`/`purge` (unloaded + deleted) → reinstall → GUI launch, no `credentials.cc` crash.

---

sabiut (approved):
Approving for my CODEOWNERS files (`scripts/doctor.sh` + `docs/troubleshooting.md`).

The doctor probe now addresses both review points: it reads the kernel's loaded-profile set and judges by whether the read returns data (correctly handling the securityfs `0444` / `CAP_MAC_ADMIN` trap), reports the real load state for root and a neutral on-disk info line for non-root (never a false PASS), and the hints are all distro-agnostic. Knob-gating keeps it silent off Ubuntu. I drove all the probe branches with redirected fixtures (knob absent/0/1 × loaded/not-loaded/unreadable × profile present/absent) and each produces the intended PASS/WARN/INFO/silent result; shellcheck -x clean. Nice catch on the securityfs read trap.

Scoping note: I'm not reviewing the `deb.sh` postinst/prerm behavior — that's @aaddrick's call. Two non-blockers: the new probe has no `doctor.bats` coverage (the `run_doctor` body is untested generally), and a heads-up that this and #692 both touch `doctor.sh`, so whichever lands second should rebase.

aaddrick (approved):
Approving. The root fix here is @diarized's: a scoped AppArmor profile that grants userns to our own electron binary, sandbox left on, same shape chrome/code/slack ship. The securityfs `0444`/`CAP_MAC_ADMIN` catch was a genuine find on top of that. Thanks also to @sabiut for two solid review rounds on `doctor.sh` and `docs/troubleshooting.md`; the loaded-vs-on-disk probe and the distro-agnostic hints came out of those.

I pushed two commits on top rather than bounce this back. `6b02fbb` gates the doctor userns WARN on the deb electron path and scopes its text to X11. The knob alone was firing for AppImage/Nix users and for deb-on-Wayland sessions, all of which run `--no-sandbox` per `launcher-common.sh` and can't hit the crash, so they were getting a false "app crashes on launch" warning.

`133ffce` brings the profile lifecycle in line with Debian Policy 10.7.3: a marker header in the generated profile, postinst preserves an admin's unmarked profile instead of stomping it, postrm deletes only on purge (the `-R` unload still runs on remove), and the truncated-profile cleanup is errexit-safe. I ran the behavioral checks in a sandboxed harness: fresh install, unmarked-profile preserve, upgrade overwrite, remove-keeps-file, purge-deletes all verified, plus 105/105 BATS.

Merge order: this lands first, then #694 rebases. @diarized, since #694 is stacked on this branch, after the rebase please mirror the same lifecycle shape on the bwrap profile: marker header, purge-only delete, and the errexit-safe cleanup in the write-failure branch.

---

PR #690, opened 2026-06-04
Route GNOME Wayland global shortcuts through the XDG GlobalShortcuts portal (#404)

### LINKED ISSUE #404: Quick Entry feature does not work properly
## The Problem

On Fedora 43, with desktop environment: Gnome 49.5, the Quick Entry feature (that can be set under Settings -> General -> Quick Entry keyboard shortcut) does not work properly (*i.e.* most of the times it doesn't work at all). I suspect this to be a general issue, at least under the Gnome desktop environment.

## Issues

The keyboard shortcut only works if the claude-desktop app window is in focus, if instead another window is in focus, or the user is active on another Gnome desktop, the shortcut does nothing. This is clearly not intended behavior since the feature description is *"Quickly open Claude from everywhere"*.

## Steps to reproduce

On Fedora 43 (Gnome), download the current version of `claude-desktop` via `dnf`, following the provided repo instructions, and then try opening the Quick Entry chat bar when the claude-desktop app window is not in focus. Nothing will happen, and this is the anomalous behavior.

jerem:
## What & why

Quick Entry's global shortcut (`Ctrl+Alt+Space`) is focus-bound on modern GNOME Wayland. mutter (GNOME ≥ 49) no longer honours XWayland-side global key grabs, so upstream's `globalShortcut.register()` X11 grab only fires when Claude already has focus — the #404 symptom.

This wires up the XDG GlobalShortcuts portal path in the launcher (`scripts/launcher-common.sh`):

- **`detect_display_backend`** auto-forces GNOME Wayland to native Wayland (joining Niri). KDE/Sway/Hyprland stay on XWayland (their grabs still work).
- **`build_electron_args`** adds `GlobalShortcutsPortal` to the native-Wayland feature set, so `globalShortcut` routes through `org.freedesktop.portal.GlobalShortcuts` instead of an X11 grab.
- **`CLAUDE_USE_WAYLAND` is now tri-state**: `1` = force native Wayland, `0` = force XWayland (escape hatch if native Wayland regresses rendering), unset = auto-detect.
- **Feature flags merged into one `--enable-features=` switch.** Chromium honours only the *last* such switch, so the previous code (up to two separate ones) had a latent clobber bug this fixes.

## Status — important caveat

This is the **correct, required prerequisite** and it closes #404 on **GNOME ≤ 49** (the current mainstream releases). It is verified that the launcher delivers `--ozone-platform=wayland` + the merged `--enable-features=…,GlobalShortcutsPortal`.

On **GNOME 50 / xdg-desktop-portal ≥ 1.20** it does **not** yet work, through no fault of this change: the newer portal requires apps to declare identity via `org.freedesktop.host.portal.Registry.Register`, which Electron/Chromium does not perform — so `globalShortcut.register()` returns `false` and the portal is never contacted. This was proven by D-Bus capture plus a minimal Python portal client (which succeeds end-to-end the moment `Registry.Register` is called), across Electron 40.6.1 / 41.5.0 / 41.7.1 / 42.3.3. Filed upstream: electron/electron#51875.

Full diagnosis and proof: `docs/learnings/wayland-global-shortcuts-portal.md`.

## Testing

- `tests/launcher-common.bats` — 8 new cases (GNOME → native, `ubuntu:GNOME`, case-insensitive, `CLAUDE_USE_WAYLAND=0` escape hatch on GNOME/Niri, single merged `--enable-features` carrying `GlobalShortcutsPortal`, portal absent on XWayland). All 85 pass.
- `shellcheck` clean.
- `S12` (`--enable-features=GlobalShortcutsPortal` present in argv on GNOME Wayland) passes.

Refs #404. Upstream blocker: electron/electron#51875.


sabiut:
Reviewed from the test side (`tests/launcher-common.bats`). Pulled the branch — 85/85 green, `shellcheck -x` clean.

Coverage looks great and the assertions are solid, not surface-level:

- The `--enable-features` clobber fix is properly pinned — `count_enable_features -eq 1` confirms only one switch survives, and the `--enable-features=*UseOzonePlatform*GlobalShortcutsPortal*` glob confirms the features actually merge into it. The hidden-titlebar variant checks `WindowControlsOverlay` + `WaylandWindowDecorations` + `GlobalShortcutsPortal` all coexist. That's exactly the test that latent bug needed.
- The tri-state `CLAUDE_USE_WAYLAND` is fully covered: `=1` force-native, `=0` force-XWayland (on both GNOME and Niri), and unset→auto.
- GNOME detection matrix is thorough: plain `GNOME`, colon-separated `ubuntu:GNOME`, case-insensitive, GNOME-on-X11 stays X11, and the portal correctly absent on XWayland.

One optional nit: the merged-features check uses an ordered glob (`*UseOzonePlatform*GlobalShortcutsPortal*`). Order inside `--enable-features` doesn't matter to Chromium, so a future reorder would fail the test for no real reason — two separate `has_electron_arg` checks would be slightly more robust. Non-blocking.

Tests LGTM. Deferring the launcher-logic / GNOME 50 portal-blocker side to @aaddrick.

sabiut (approved):
Approving for the test coverage (`tests/launcher-common.bats`) — details in my comment above. The launcher logic and GNOME 50 portal blocker are @aaddrick's call.

aaddrick:
Excellent diagnosis. The D-Bus capture, the Python portal client that actually proves Electron never fires the `Registry.Register` app-id handshake, and the upstream filing at electron/electron#51875 are the kind of evidence most of these never get. The learnings doc stands on its own. And the `--enable-features=` merge is a genuine latent bug — Chromium honouring only the last switch means any second `--enable-features` we ever added would have silently clobbered the first. That fix is correct regardless of the portal work.

My proposal is to split this. The decoupled, zero-regression parts I'd take as-is: the single merged `--enable-features`, adding `GlobalShortcutsPortal` to the native-Wayland feature set, and the tri-state `CLAUDE_USE_WAYLAND`. Those are clean and they help every backend.

The piece I want to hold is the GNOME auto-flip. It moves every GNOME Wayland user off the boring, mature XWayland path onto native Wayland, and that's a rendering / IME / HiDPI / NVIDIA-and-fractional-scaling change, not just a hotkey change. GNOME Wayland is the *default* session, so that's a large slice of the base, not an edge case. We'd be shipping it on argv-only verification, and on GNOME 50 — what most new installs land on now — it's a documented no-op per your own findings, so those users eat the rendering risk for none of the hotkey benefit. For now I'd make GNOME native+portal opt-in via `CLAUDE_USE_WAYLAND=1`, which works fully on GNOME ≤49 after the one-time portal dialog. The auto-flip-by-default can be a follow-up gated on one real "app still renders correctly on GNOME Wayland native" check, not just "the flag made it into argv."

One small doc note: the `configuration.md` portal-support line asserts a fixed "GNOME and KDE yes, wlroots no" allowlist. That omits COSMIC, which also ships no GlobalShortcuts backend. I'd soften it to "per-compositor and currently uneven," and lead the GNOME section with the current GNOME 50 reality rather than the ≤49 success path. Happy to help with the descope if that's useful.

---

jerem:
Thanks both — pushed the descope in 9762043.

@aaddrick agreed on all counts; the auto-flip was the risky part. Changes:

- **Dropped the GNOME auto-flip.** `detect_display_backend` now only auto-forces Niri (no XWayland at all). The default GNOME session stays on XWayland.
- **Kept the zero-regression parts:** single merged `--enable-features`, `GlobalShortcutsPortal` in the native-Wayland feature set, tri-state `CLAUDE_USE_WAYLAND`.
- **GNOME portal route is now opt-in** via `CLAUDE_USE_WAYLAND=1` (works on GNOME ≤49 after the one-time portal dialog). Left a comment marking the auto-flip-by-default as a follow-up gated on a real "renders correctly on GNOME native" check, not argv-only.
- **Docs:** `configuration.md` now says GlobalShortcuts-backend support is "per-compositor and currently uneven" and lists COSMIC alongside wlroots as having no backend; GNOME section leads with the GNOME 50 reality. CHANGELOG/learnings reframed from "forces" to "opt-in."

@sabiut nit applied — the merged-features assertion is now two order-independent `has_electron_arg '--enable-features=*UseOzonePlatform*'` / `*GlobalShortcutsPortal*` checks instead of the ordered glob. BATS green (84).

One thing to confirm: **S12**. With GNOME no longer auto-flipping, a default launch won't carry the portal flag, so I changed the spec to launch with `CLAUDE_USE_WAYLAND=1` — i.e. it now verifies the flag wiring on the opt-in path. If you'd rather S12 stay a default-launch detector that goes red until the auto-flip follow-up lands, I can flip it back — let me know which contract you want.

sabiut:
Keep S12 as you've changed it — opt-in path. The "stays red until the fix lands" contract only made sense while the auto-flip was in scope; now that the default is *deliberately* XWayland, a red S12 wouldn't be detecting a regression, it'd be alarming on a decision we made on purpose. Permanently-red cells just teach people to ignore that column.

The default-launch contract isn't lost either — it's pinned at the launcher level by the new "GNOME Wayland keeps XWayland default" BATS case, which is the right layer for it. When the auto-flip follow-up lands, that PR should flip S12's contract in the same change that flips the behavior.

Ran the suite locally at `2b332b9`: 84/84 green. Also sourced the launcher directly — opt-in yields a single merged `--enable-features` with the portal flag, default launch carries none. Both sides of the descope behave as documented. Approval stands.

aaddrick (approved):
Approving. @jerem, the portal diagnosis is the standout here: the `Registry.Register` root cause pinned to `portal.cc`, the D-Bus captures across four Electron versions, and the Python client that proves the portal itself works. The descope landed cleanly too: tri-state `CLAUDE_USE_WAYLAND`, Niri-only auto-detect, GNOME opt-in via `=1`. @sabiut already covered the test side in depth and re-confirmed at head, so I'll skip repeating that.

The `--enable-features` merge is load-bearing on its own. I checked the old code on the hidden-titlebar + native-Wayland path: it emitted two separate `--enable-features` switches, and Chromium honours only the last one, so `WindowControlsOverlay` got silently dropped. The accumulator closes that. And the X11 path still carries the overlay flag since the old early `return` became an `if/else`, so no regression there.

I pushed a docs-only commit to the branch (b511a8b). It aligns the `configuration.md` env table with the descope (only Niri auto-defaults to native Wayland), fixes the stale S12 diagnostics line that pointed at a log message the descoped launcher never emits, and adds two portal notes: KDE's app-id requirement since xdg-desktop-portal 1.21, and that a denied portal dialog persists in the permission store across launches.

On CI: the four runs were queued at `action_required` this whole time because it's a first-time fork contribution, and I've now approved them. Shellcheck and Codespell are already green; CI and BATS are running. Merging once the gate finishes.

---

PR #691, opened 2026-06-04
ci(test-artifacts): test arm64 packages + fix AppImage launch-smoke pkill

sabiut:
Two hardening fixes to the artifact-test harness, found while reviewing the post-#670 launch-smoke work.

## 1. Validate arm64 packages, not just amd64

`test-artifacts` only ran against amd64. The `build-arm64` deb/rpm/appimage artifacts shipped with **zero** structural, `--doctor`, or launch-smoke coverage — an arm64-only regression could reach a release unnoticed.

- Parametrize the reusable `test-artifacts.yml` by `arch` (default `amd64`).
- Run the arm64 matrix on a native `ubuntu-22.04-arm` runner (matching `build-arm64`), so the launch smoke test executes the real arm64 binary rather than failing on a foreign architecture.
- `ci.yml` now invokes `test-artifacts` twice (amd64 + arm64); the `release` gate waits on both.

## 2. Fix the AppImage launch-smoke pkill match

A running AppImage execs Electron from its FUSE mount (`/tmp/.mount_claudeXXXX`), so escaped zygote/electron children live under that mount — not under the `.AppImage` path. The pkill sweep was passed the artifact path, so it matched only the already-reaped top-level launcher and never the strays it exists to catch, leaking electron children on the runner.

Pass `mount_claude` instead (per CLAUDE.md's pkill guidance). deb/rpm are unaffected — they already match the real `/usr/lib/claude-desktop` exec path.

## Testing

- `actionlint` + `shellcheck -x` clean.
- Verified the artifact-name contract: builds upload `package-{amd64,arm64}-{deb,rpm,appimage}`; the test job downloads `package-${arch}-${format}` — exact match.
- The arm64-runner path is proven on this CI run (no local arm64 available).

Note: a tool-presence guard (the third item from the same review) was already on `main`, so it's not included here.

Refs #670

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,9 +54,11 @@ jobs:
       release_tag: ${{ startsWith(github.ref, 'refs/tags/v') && github.ref_name || '' }}
 
   test-artifacts:
-    name: Test Build Artifacts
+    name: Test Build Artifacts (amd64)
     needs: [build-amd64]
     uses: ./.github/workflows/test-artifacts.yml
+    with:
+      arch: amd64
 
   build-arm64:
     name: Build Packages (arm64 - ${{ matrix.artifact_suffix }})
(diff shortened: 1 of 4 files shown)


aaddrick (approved):
Hey @sabiut — solid follow-on to #671, approving. The pkill fix is mechanistically right: `$appimage_file` could only ever match the wrapper the process-group kill already reaped, while the strays it exists to catch run from `/tmp/.mount_claude*`. The arm64 plumbing checks out exactly: builds upload `package-{arch}-{format}`, the test downloads the same, no stale `matrix.artifact` refs left, and the `release` gate picked up `test-artifacts-arm64`, so a broken arm64 package now blocks releases like amd64 does.

One thing I pushed to your branch: 47e4011 guards the `mount_claude` sweep behind `[[ -n ${CI:-} ]]`. As a `pkill -KILL -f` pattern, `mount_claude` also matches any live local Claude Desktop AppImage session. These scripts support standalone runs, and Ctrl-C mid-test would take out a developer's running app via the EXIT trap. The helper already no-ops an empty pattern at both sweep sites (`test-artifact-common.sh:186`/`:315`), so local runs cleanly fall back to the process-group kill alone. Nothing else changed.

One coverage caveat for the record: arm64 deb and AppImage genuinely reached ready state under Xvfb on the native runner, but the arm64 rpm job hit the #671 sandbox-denied SKIP inside the Fedora container, same as amd64. So rpm launch stays unproven on both arches. What this PR adds is native deb + AppImage launch proof, and that's the real win.

---

PR #692, opened 2026-06-04
fix(doctor): stop false-green PASS on empty password store + bad disk read

sabiut:
`--doctor` is the project's release-readiness signal — a green `[PASS]` is meant to mean "this was checked and is fine." Two checks violated that: they emitted a green `[PASS]` on data they had actually **failed to read**, so a real problem (or a regression in the diagnostic itself) would read as healthy. This fixes both and pins them with tests.

## Fix 1 — password store: blank PASS on detection failure

**Symptom:** `_doctor_check_password_store` ran `store=$(_detect_password_store)` and immediately did `_pass "Password store: $store"`. If the helper returned an empty string, the output was a green `[PASS] Password store: ` — a blank value presented as success.

**Why it matters:** `_detect_password_store` lives in `launcher-common.sh`, which sources `doctor.sh`. A sourcing-order regression (or the helper failing at runtime) would surface here as a reassuring PASS, not a FAIL — exactly the failure `--doctor` exists to catch. Note `basic` is a *legitimate* fallback (filesystem-permission-protected tokens) and is intentionally **not** treated as an error; only an *empty* result indicates detection itself failed.

**Fix:** warn on an empty backend and return early, before the PASS:

```sh
store=$(_detect_password_store)
if [[ -z $store ]]; then
	_warn 'Password store: unable to detect backend'
	return
fi
_pass "Password store: $store"
```

## Fix 2 — disk space: PASS on a non-numeric `df` reading

**Symptom:** the disk-space check guarded only against an *empty* `df` result (`[[ -n $config_disk_avail ]]`), not a *non-numeric* one. A malformed `avail` field (unusual `df` output, locale, a value like `N/A`) is non-empty, so it skipped the empty-guard, then failed both `((avail < 100))` and `((avail < 500))` arithmetic comparisons and fell through to the `else` branch — reporting `[PASS] Disk space: <garbage>MB free`.

**Why it matters:** it reports free space it never actually parsed, and the whole block had **zero test coverage**.

**Fix:** extracted the inline block into a `_doctor_check_disk_space()` helper with a numeric guard that mirrors the existing `_doctor_check_filename_limit` pattern (`doctor.sh:464`):

```sh
avail=$(df -BM --output=avail "$config_dir" 2>/dev/null \
	| tail -1 | tr -d ' M') || true
[[ $avail =~ ^[0-9]+$ ]] || return 0
```

A non-numeric or unreadable value now skips silently (no line) rather than emitting a false PASS — "better no line than a green PASS on data we couldn't read." Extracting it into a named helper (alongside the other `_doctor_check_*` functions) also makes it unit-testable via a `df` shim.

## Tests (`tests/doctor.bats`, +6 cases — 36/36 pass)

- **password store:** empty backend → `[WARN]` and **not** `[PASS]`.
- **disk space:** `<100MB` → `[FAIL]`; `<500MB` → `[WARN]`; ample → `[PASS]`; non-numeric `df` → no PASS/FAIL/WARN at all; `df` unavailable → silent skip, no output.

## Verification

- `shellcheck -x scripts/doctor.sh` — clean.
- `bash -n` — clean.
- `bats tests/doctor.bats` — 36/36.
- Manually confirmed `run_doctor` still emits the disk-space line after the extraction (the call site is otherwise not covered by a standing test — `run_doctor` integration coverage is tracked separately).

## Scope / notes

- Behaviour-preserving for the healthy paths: a valid backend and a normal numeric `df` reading produce the same output as before.
- These came out of a review of the testing/release-quality surface; no linked issue.

### DIFF
--- a/scripts/doctor.sh
+++ b/scripts/doctor.sh
@@ -562,10 +562,16 @@ _doctor_check_recent_crashes() {
 # sources this file) to surface what keyring Electron will use for
 # safeStorage / cookie encryption. 'basic' is valid but means tokens
 # rely on filesystem permissions alone, so we note it for visibility.
-# Never fails — basic is an intentional fallback, not an error.
+# 'basic' is an intentional fallback, not an error; an empty result
+# means detection itself failed (e.g. a sourcing-order regression) and
+# warns rather than emitting a green PASS with a blank value.
 _doctor_check_password_store() {
 	local store
 	store=$(_detect_password_store)
+	if [[ -z $store ]]; then
+		_warn 'Password store: unable to detect backend'
+		return
+	fi
 	_pass "Password store: $store"
 	if [[ $store == 'basic' ]]; then
 		_info \
@@ -575,6 +581,28 @@ _doctor_check_password_store() {
 	if [[ -n ${CLAUDE_PASSWORD_STORE:-} ]]; then
 		_info \
 			"  → overridden by CLAUDE_PASSWORD_STORE=${CLAUDE_PASSWORD_STORE}"
+	fi
+}
+
+# Report free space on the partition holding the Claude config dir.
+# Arguments: $1 = config directory to check.
+#
+# Skips silently when df is unavailable or yields a non-numeric value:
+# better no line than a green PASS reporting space we could not read.
+_doctor_check_disk_space() {
+	local config_dir="$1"
+	local avail
+	avail=$(df -BM --output=avail "$config_dir" 2>/dev/null \
+		| tail -1 | tr -d ' M') || true
+	[[ $avail =~ ^[0-9]+$ ]] || return 0
+	if ((avail < 100)); then
+		_fail "Disk space: ${avail}MB free on config partition"
+		_info 'Fix: Free up disk space'
+	elif ((avail < 500)); then
+		_warn "Disk space: ${avail}MB free" \
+			"on config partition (low)"
+	else
+		_pass "Disk space: ${avail}MB free"
 	fi
 }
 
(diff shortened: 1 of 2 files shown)


aaddrick (approved):
Hey @sabiut! Both fixes are real. I reproduced the disk-space bug exactly as described: `avail="N/A"` clears the old `[[ -n ]]` guard, both arithmetic comparisons error out, and execution falls through to the PASS branch. Confirmed false green. Pulling the block into `_doctor_check_disk_space` next to the other helpers matches the `_doctor_check_filename_limit` pattern and finally makes it testable. I also checked the usual casualty of a move like this: `_fail` still increments `run_doctor`'s `local _doctor_failures` via dynamic scoping, since the helper is called directly and not in a subshell.

I pushed `d1fe021` to the branch with some non-blocking hardening from review, so nothing here needs a round-trip. The fall-through-to-PASS shape survived one more input class: `avail="0099"` passes `^[0-9]+$`, then both `(( ))` comparisons die on octal ("value too great for base") and you land in the PASS branch again. Same bug class this PR kills, through a different door. `df -BM` never emits leading zeros so it's theoretical, but `avail=$((10#$avail))` after the regex closes it for good. The same residue lived in `_doctor_check_filename_limit`, so I normalized both.

The commit also adds an `_info` line when the df read is skipped (before, the summary printed "All checks passed" over a check that never ran, which is its own quiet false green), dedupes the password-store docstring, and pins three new BATS cases: the 100/500 tier boundaries plus the `0099` input. Suite went 36 to 39 passing, shellcheck clean. Thanks for the fix!

---

PR #693, opened 2026-06-04
test(launcher): cover cleanup_orphaned_cowork_daemon

sabiut:
`cleanup_orphaned_cowork_daemon` (`scripts/launcher-common.sh`) reaps a `cowork-vm-service` daemon left behind by a crashed UI — it `SIGTERM`s, waits, then escalates to `SIGKILL`. Despite killing processes, it had **zero test coverage**. This adds four cases to `tests/launcher-common.bats` pinning each branch:

| Scenario | Expectation |
|---|---|
| No daemon running | returns early — no kill, no log |
| Live UI present | daemon left alone (UI is "expected") |
| Orphan dies on `SIGTERM` | single `SIGTERM`, **no** `SIGKILL`, correct log line |
| Orphan survives `SIGTERM` | escalates to `SIGKILL`, logs the `SIGKILL` variant |

## Approach

- **Live-UI case** uses a real background process as the stand-in UI, so the function's `/proc/$pid/{cmdline,status}` reads resolve naturally — no faking `/proc`. `pgrep`/`kill`/`sleep` are function-stubbed (matching the existing `pgrep() { return 1; }` precedent in this file); the `kill` stub records invocations so the tests assert exactly which signals were sent.
- **Negative assertions** use `run grep …; [[ $status -ne 0 ]]` rather than bare `! grep` — a `!`-negated command that isn't the last in a bats test does not fail the test (SC2314), which would make the "no SIGKILL" check hollow.
- Calls go through `run` so the function's `((_wait++))` (returns 1 when `_wait` starts at 0) doesn't trip bats' errexit. Production has no `set -e`, so that's a harness concern, not a code defect — `launcher-common.sh` is unchanged.

## Verification

- **79/79** `launcher-common.bats` pass; no source changes.
- Mutation-tested to confirm the tests actually bite: neutering the live-UI guard makes the "live UI present" test fail; neutering the `SIGKILL` escalation makes the "orphan survives SIGTERM" test fail.

@RayCharlizard — this is cowork-daemon territory. The test lives in `launcher-common.bats` (where the function is defined) rather than a `cowork-*.bats`, but a sanity check on the expected reap semantics would be welcome.

Note: `_find_virtiofsd`, flagged alongside this in the same review, turned out to already be covered in `tests/cowork-bwrap-config.bats` — no work needed there.

### DIFF
--- a/tests/launcher-common.bats
+++ b/tests/launcher-common.bats
@@ -620,6 +620,115 @@ s.close()
 }
 
 # =============================================================================
+# cleanup_orphaned_cowork_daemon
+#
+# Reaps a cowork-vm-service daemon left behind by a crashed UI, but only
+# when no live Claude UI is running. pgrep/kill/sleep are stubbed; the
+# "live UI" case uses a real background process so the /proc cmdline and
+# status reads resolve naturally without faking /proc.
+# =============================================================================
+
+@test "cleanup_orphaned_cowork_daemon: no daemon running — no action, no log" {
+	# Daemon pgrep finds nothing, so the function returns before any
+	# UI scan or kill.
+	pgrep() { return 1; }
+	kill() { echo "kill $*" >> "$TEST_TMP/kills"; }
+
+	setup_logging
+	run cleanup_orphaned_cowork_daemon
+	[[ $status -eq 0 ]]
+	[[ ! -f "$TEST_TMP/kills" ]]
+	[[ ! -f $log_file ]]
+}
+
+@test "cleanup_orphaned_cowork_daemon: live UI present — daemon left running" {
+	# A real background process stands in for the live Electron UI: its
+	# cmdline ('sleep ...') is neither the cowork daemon nor a --type=
+	# helper, and its state is sleeping (not T/t/Z), so the function
+	# treats it as a live UI and must NOT kill the daemon.
+	sleep 300 &
+	ui_pid=$!
+
+	pgrep() {
+		if [[ $2 == *cowork-vm-service* ]]; then
+			echo 4242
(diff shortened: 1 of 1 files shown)


aaddrick (approved):
Hey @sabiut — good PR. The reaper had zero coverage, and these four cases pin the four top-level outcomes. I mutation-tested the kill paths in a worktree: neutering the live-UI guard, the SIGKILL escalation, and the initial SIGTERM loop each fail the right tests, so the signal assertions check the actual `kill`, not just the log line. The `run grep …; [[ $status -ne 0 ]]` negative-assertion handling (vs. a hollow non-final `! grep`, SC2314) and the documented `((_wait++))` errexit workaround are both careful work.

One coordination note for whoever merges: these tests stub `pgrep`, so the real `app\.asar` fingerprint never runs. PR #700 removes the `app.asar` launcher arg, which static analysis says breaks live-UI detection in this exact function while this suite stays green — and once the fingerprint is re-keyed to something without the substring `app`, the live-UI test's `*app*` stub dispatch falls through and false-fails. So: re-key the fingerprint in #700's fix and update these stubs in the same push. That's merge ordering, not a blocker for this PR.

Non-blocking follow-up if you're up for it: the UI-scan skip branches (`--type=` helpers, T/t/Z state, `$$`/`$PPID`) survive mutation under this suite. A lingering `--type=` helper with the main process gone, and a stopped `app.asar` process, would pin the two cases the function's big comment defends. The same stubs also support a multi-PID partial-death case if you want to go further.

@RayCharlizard — cowork-daemon territory, flagging for a sanity check on the expected reap semantics and the #700 ordering. Approving.

---

PR #694, opened 2026-06-04
fix(deb): auto-install AppArmor userns profile for Cowork bwrap backend

diarized:
> **Depends on [#687](https://github.com/aaddrick/claude-desktop-debian/pull/687). Merge that first.** Until then this diff also shows #687's commits.

## Problem

Cowork's bubblewrap backend needs unprivileged user namespaces. Ubuntu 24.04+ blocks them (`apparmor_restrict_unprivileged_userns=1`). Cowork silently falls back to host-direct (no isolation).

## Fix

The `.deb` `postinst` installs `/etc/apparmor.d/claude-desktop-bwrap`, granting `userns` to `/usr/bin/bwrap`. Separate binary from #687's Electron profile.

## Scope

- `deb.sh`: adds `Recommends: bubblewrap`; postinst installs the profile (abort-proofed, like #687); postrm removes **both** profiles on `remove`/`purge`/`abort-install`.
- `doctor.sh`: bwrap-probe hint points at the profile.
- `troubleshooting.md`, `CHANGELOG.md`: updated.

## Design

- Capability-gated (`apparmor_parser -Q`), not distro-gated. Covers hardened Debian. No-op where unrestricted.
- Static detection in postinst. Root is exempt from the restriction, so a behavioral probe there would falsely pass.
- Behavioral probe stays in `--doctor` (runs as the user).
- Distinct filename + clash guard. Defers to a user's `/etc/apparmor.d/bwrap`. No dpkg conflict.

## Validation (Ubuntu 24.04)

Build → install → both profiles load → `bwrap` works as user → doctor reports `bubblewrap (namespace sandbox)` → remove unloads both → reinstall restores. In-app Cowork session confirmed isolated: empty home, no SSH keys, 5 processes.

## Reviewer Q&A

| # | Likely question | Your answer |
|---|------------------|-------------|
| 1 | Why a separate profile, not folded into #687's `/etc/apparmor.d/claude-desktop`? | Different binary (`bwrap` ≠ Electron). Distinct file means postrm won't delete a user's hand-made `/etc/apparmor.d/bwrap`, and no dpkg conflict if `bubblewrap` ever ships its own. |
| 2 | **Security:** granting `userns` to `/usr/bin/bwrap` is host-wide for *all* callers — regression? | Same trade-off as the existing #351 manual fix and Ubuntu's own `chrome`/`code`/`slack` profiles. Far narrower than disabling the sysctl globally. The profile only *unlocks* userns; bwrap's own empty-root sandbox does the confining. Noted in the troubleshooting security paragraph. |
| 3 | Why not a profile *transition* scoped to only Claude's bwrap? | More precise but significantly more complex/brittle; `flags=(unconfined){ userns }` matches the upstream chrome/code/slack pattern and #687. Reasonable future hardening, not needed now. |
| 4 | Why `Recommends`, not `Depends`? | Cowork is optional; app runs without bwrap (host-direct/KVM). `Depends` would force it onto every install, including Debian where it's moot. |
| 5 | Does this break/affect plain Debian? | No — capability-gated, not distro-gated. Harmless no-op where the restriction is off; fixes hardened Debian where it's on. |
| 6 | Idempotency / upgrade behavior? | postinst overwrites cleanly and is abort-proofed (a read-only `/etc` can't fail the install). postrm cleans on `remove`/`purge`/`abort-install`, skips `upgrade` so the incoming postinst reloads. Tested install→remove→purge→reinstall. |
| 7 | RPM / AppImage? | Out of scope. RPM = SELinux (no `apparmor_parser`, no-op). AppImage has no postinst and runs `--no-sandbox`; its Cowork-bwrap stays host-direct — flagged as separate future work. |
| 8 | Clash guard misses alternate profile names (e.g. `usr.bin.bwrap`)? | True edge; AppArmor resolves two profiles on one binary by attachment specificity (no crash). Non-blocking. |

---
60% AI / 40% Human
Claude: investigated the bwrap-vs-Electron userns gap, implemented the postinst/postrm/control/doctor/docs changes, built and installed the `.deb`, validated isolation end-to-end on Ubuntu 24.04.
Human: directed the discovery-led detection design, chose the scoped profile over a global sysctl, set the validation gates, owns the fork and PR.


diarized:
Rebased onto the updated #687 head (`af05e3d`) — **history rewritten (force-push).** #687's review round moved its AppArmor cleanup from `prerm` to `postrm`, so this branch was rebased to match:

- **Cleanup re-homed to `postrm`.** Both profiles (`claude-desktop` + `claude-desktop-bwrap`) are now unloaded + deleted in one `postrm` (`case remove|purge|abort-install`). The old `prerm` is gone. The cosmetic "drop duplicated prerm comment" commit became a no-op against the merged `postrm` and was dropped (9 → 8 commits).
- **bwrap profile write abort-proofed.** Matched #687's pattern — `mkdir` + heredoc moved into an `if`-condition so a read-only/atypical `/etc` can't fail the install under `set -e`. Both AppArmor blocks in `postinst` are now equally guarded.

No functional change to the bwrap detection/install logic (still capability-gated + clash-guarded). Re-validated: `shellcheck` clean on `deb.sh`/`doctor.sh`; generated `postinst` (both blocks) + `postrm` pass `dash -n` and `shellcheck -s sh`.

Description updated to match (`prerm` → `postrm` throughout).

---

sabiut (changes requested):
Reviewed the #694-only delta (`af05e3d...6c09ca2`). The postinst/postrm mechanics are solid — the postrm consolidation matches what we settled on in #687, the heredoc escaping is correct, validate-then-load with `-Q` plus cleanup on unsupported rules is the right pattern, and keeping the doctor's *behavioral* probe as the source of truth (instead of a static file check) is exactly right given the root-exemption trap you documented in the postinst comment.

Two blockers, both about non-deb users, then some smaller items.

**1. The doctor hint is deb-specific again (blocker)**

The new hint in the bwrap probe-failure path says "The .deb auto-installs an AppArmor profile … reinstall the .deb to restore it." `doctor.sh` ships in every format — an AppImage/Nix/rpm user on Ubuntu 24.04 hits this same probe failure and gets advice that doesn't apply to them. This is the same issue we resolved in #687's review round (where the fix was to drop "reinstall the .deb" and point all hints at `docs/troubleshooting.md`). Same resolution here please: point at the troubleshooting section only, and let that page branch per format.

**2. `troubleshooting.md` drops the manual workaround non-deb users still need (blocker)**

The rewrite says "**Fix:** None needed" and deletes @hfyeh's manual profile snippet. But the auto-install is deliberately deb-only — so AppImage/Nix/rpm/manual installs on Ubuntu 24.04+ now have no documented fix, and "None needed" is wrong for them. The tightened structure is an improvement; just keep the manual heredoc, framed as "for non-deb installs."

**3. Gate inconsistency between the two postinst blocks**

The Electron block gates on knob existence (`[ -e /proc/sys/kernel/apparmor_restrict_unprivileged_userns ]`) — that was #687's review outcome, so stock Debian (knob absent) gets no profile. The bwrap block drops that gate and uses capability detection only. Net effect: on a stock Debian with AppArmor 4 but no knob, the Electron profile is skipped but the bwrap one installs. A kernel that can impose this restriction *has* the knob — it's the mechanism itself — so a knobless system gets a dead profile that takes ownership of `/usr/bin/bwrap`'s attachment for nothing. Please add the same knob-existence gate to the bwrap block, both for behavior and so the two blocks read consistently.

**4. Partial profile on failed write (nit, both blocks)**

If `cat` dies mid-write (ENOSPC), the else branch warns but leaves a truncated file in `/etc/apparmor.d/`, which the parser will choke on at every reload/boot. One-liner: `rm -f` the profile in the else branch. The #687 block has the same gap — happy for both to be fixed here.

**5. Recommends ordering (observation, no change needed)**

On a fresh install where bubblewrap arrives via the new Recommends in the same apt transaction, dpkg only guarantees configure ordering for Depends — postinst can run before bwrap is unpacked, and the profile is silently skipped. The doctor probe catches it and reinstall fixes it, so this is fine; just worth a sentence in the postinst comment.

**6. Symptom literal (docs nit)**

The troubleshooting symptom says `Cowork isolation: host-direct (bwrap probe failed)`; the actual output is `host-direct (bwrap probe failed — see above)`, and the primary failure line is `bubblewrap: sandbox probe failed (rc=N)`. Per the docs styleguide, symptoms should be the literal string.

**7. Codify the maintainer-script lint**

You ran `dash -n` + `shellcheck -s sh` on the generated postinst/postrm by hand — that's exactly the kind of check that should live in the suite so the next edit can't regress it. `test-artifact-deb.sh` already installs the deb in a container (which implicitly proves postinst survives the no-apparmor path); a cheap assertion that the generated scripts lint clean would close the loop. Let's make this a follow-up issue rather than in-PR — I have #691 open touching `test-artifact-deb.sh` and would rather not cross the streams. I'll file it.

---

Heads-up on sequencing: my #692 also touches `doctor.sh` (different hunks — disk-space/password-store, just above the Cowork section). Whichever of #692/#694 merges second will need a trivial rebase.

sabiut:
Quick follow-up — I actually ran the generated postinst/postrm from this branch in a bwrap sandbox (stubbed `apparmor_parser`, `--tmpfs /etc`) on a Debian 12 box, which has no userns knob.

Item 3 isn't hypothetical: with the stub parser accepting `-Q` and the knob absent, the Electron block skipped but `/etc/apparmor.d/claude-desktop-bwrap` got installed anyway. That's exactly what trixie would do (AppArmor 4.x userspace, no restriction patch in the kernel). Same knob gate as the Electron block fixes it.

The rest held up fine: read-only `/etc` warns and exits 0, `-Q` rejection cleans up the profile, the clash guard works, postrm removes both profiles on remove and leaves them on upgrade, and both generated scripts pass `dash -n` + `shellcheck -s sh`.

The sandbox harness is what I'll propose in the item-7 follow-up issue — it caught the item-3 behavior directly, runs in a second, no root needed. Filing after #691 lands to avoid conflicts in `test-artifact-deb.sh`.

aaddrick:
Pushed four commits on top of @diarized's branch; head is now `8319407`. I merged #687's updated head rather than rebasing, so all the original commits are intact and @sabiut's earlier review anchors still resolve. This should close out the outstanding requested-changes round.

What landed: the kernel-knob gate is back on the bwrap profile block, so knobless AppArmor-4 systems (trixie-style) no longer get a dead profile. The gate is strictly narrowing, so Ubuntu behavior is unchanged. The clash guard now scans `/etc/apparmor.d` for any profile attaching to `/usr/bin/bwrap` instead of checking one filename, which catches `apparmor-profiles`' `bwrap-userns-restrict`. The `[ -x /usr/bin/bwrap ]` gate is dropped, since a profile attaching to a missing binary is inert; that removes the Recommends install-order race outright. The bwrap profile also picked up the same Debian-Policy lifecycle shape we settled on in #687: marker header, unmarked-profile preservation, purge-only delete, errexit-safe cleanup. On the docs side, the doctor hint is format-neutral again, @hfyeh's manual setup snippet is restored with credit for non-deb installs, and the `local/` override convention is documented as the supported customization point.

Verification: a stubbed-apparmor_parser harness runs 25/25 across the install, skip, preserve, rewrite, foreign-attachment, self-clean, and remove/purge paths. BATS is 274/274 and shellcheck is clean. One caveat: the harness stubs `apparmor_parser`, so end-to-end behavior on a real Ubuntu 24.04 host hasn't been re-run since this push. @sabiut's review and his stub-parser repro were the map for most of this; the knob gate and failed-write cleanup are his calls, just implemented.

Merge order stands: #687 first, then this one. The diff here is already the clean bwrap-only delta on top of it, so it merges without a rebase. Thanks @diarized for the groundwork. The postinst mechanics and the separate-profile decision were right and carried through unchanged.

---

PR #695, opened 2026-06-04
fix(packaging): normalize install perms so Cowork daemon can launch

caidejager:
The packaged app.asar.unpacked/ directory shipped as mode 0700 owned by the build uid. The launcher runs as the desktop user (a different uid), which cannot traverse that directory, so the app's daemon auto-launch guard — if (fs.existsSync(.../app.asar.unpacked/cowork-vm-service.js)) — silently returns false and the Cowork VM daemon never forks. The symptom is an endless `connect ENOENT` on the VM-service socket with no cowork_vm_daemon.log and no [cowork-autolaunch] line: nothing is even attempted.

Root cause is packaging, not app code. A restrictive build umask creates 0700 directories, and dpkg-deb records the build uid verbatim because the build does not run under fakeroot.

- deb.sh: normalize the installed tree to canonical modes (dirs and executables 755, other files 644) before building, and build with `dpkg-deb --root-owner-group` for root:root ownership.
- appimage.sh: apply the same normalization to the AppDir before mksquashfs (it copies with `cp -a`, preserving the bad modes).
- RPM was already safe: %defattr(-, root, root, 0755) forces dir perms.

Validated by building a mock .deb: app.asar.unpacked/ now ships as drwxr-xr-x root/root and the bundled electron binary stays executable.


aaddrick (approved):
Approving. The diagnosis is the strongest part of this PR: `fs.existsSync()` returning false on a directory the running user can't traverse, with no else branch and no log, is exactly the kind of silent packaging failure that eats a weekend. The learnings entry with the run-as-desktop-user `stat` check will save the next person real time. The deb/AppImage fix is correctly scoped at the packaging boundary, and `u=rwX,go=rX` keeps the binaries executable while dropping data files to 644.

Review verification turned up one gap: the "RPM was already safe" claim. `%defattr(-, root, root, 0755)` forces directory modes in the payload, but the `-` first field preserves file modes verbatim from the buildroot, which `%install` populates with plain `cp -r`. I reproduced it on a Fedora host: a umask 077 build shipped `app.asar` at 600 and the electron binary at 700. CI builds run at umask 022 and were fine, so this only bit local restrictive-umask builds. Same root cause you found, louder symptom.

I pushed d5d7f2a on top of your work rather than hold you up, since you scoped RPM out on a perfectly plausible `%defattr` reading. It adds the equivalent one-line normalization to rpm.sh's `%install` (placed before the chrome-sandbox 4755 chmod so the suid bit survives, verified in the built payload), corrects the CHANGELOG and learnings wording to match, and adds artifact-test assertions pinning `app.asar.unpacked` perms and the setuid sandbox in both deb and rpm.

One more credit: `--root-owner-group` does more than hygiene. Debs built on uid-1001 CI runners recorded that uid as owner of the installed tree, so a local uid-1001 user could overwrite `app.asar`. Your change closes that exposure too. CI runs next; merging once it's green. Thanks @caidejager.

---

PR #700, opened 2026-06-09
fix: stop passing app.asar as an Electron arg (root cause of #696 attach prompt)

### LINKED ISSUE #696: [bug]: File-attach prompt still appears on taskbar reopen after v2.0.18 update

```shell
Claude Desktop Diagnostics

[PASS] Installed version: 1.10628.2-2.0.18
[PASS] Display server: X11 (DISPLAY=:0)
[PASS] SingletonLock: held by running process (PID 3008)
[PASS] MCP config: valid JSON (/home/troijaa/.config/Claude/claude_desktop_config.json)
       MCP servers configured: 0
[WARN] Node.js: not found (required for MCP servers)
       Fix: Install Node.js v20+ from https://nodejs.org
[PASS] Disk space: 801317MB free

Cowork Mode
----------------
[PASS] bubblewrap: found
[PASS] KVM: accessible
[PASS] vsock: module loaded
       QEMU: not found
       socat: not found
       virtiofsd: not found
       VM image: not downloaded yet
[PASS] Log file: 104KB (/home/troijaa/.cache/claude-desktop-debian/launcher.log)

All checks passed.
```

### What happened

The "app.asar" file-attach prompt still appears when reopening 
the app from the taskbar after v2.0.18, which claimed to fix this bug (#668).

###Steps to reproduce

1. Install claude-desktop 1.10628.2-2.0.18
2. Close and reopen from taskbar
3. "app.asar" prompt appears again

Expected: No prompt
Actual: Prompt still appears

### Expected behavior

No "app.asar" prompt should appear. The fix in #669 should prevent 
.asar paths from being passed to the file-drop handler.

### Logs / errors

```shell

```

### Anything else

_No response_

emandel82:
## Summary

Root-cause fix for the recurring **"Attach app.asar to this session?"** prompt (#696, previously #668 / #383). All four launchers (`deb`, `rpm`, `appimage`, `nix`) append `app.asar` as a command-line argument to Electron. Because `app.asar` already sits at Electron's **default `resources/` location next to the binary**, Electron auto-loads it as the app *without consulting the command line* — so the appended path is redundant, and Electron hands it to the app as a **file to open**. The app then forwards it to its file-drop handler and prompts the user.

This removes the redundant argument, killing the prompt at the source.

## Why this is the root cause (and why the JS patches kept regressing)

The renderer-side `.asar` filters in #640 / #650 / #669 patched the *handlers that receive* the path. But the path keeps being **injected by the launcher**, and the app has more than one consumer of it:

- **Startup scan** (`app.whenReady`): builds the file list with a guard `process.argv.slice(1).filter(n => !n.startsWith("-") && resolve(n) !== app.getAppPath())`. In the auto-load layout that guard **does not match** (`getAppPath()` ≠ the argv path), so `app.asar` falls through to the generic file branch.
- **Second-instance handler** (`lKr`, fired on taskbar reopen): has **no `getAppPath()` guard at all** — it only checks `existsSync(i)`, then calls `cCA(A)`. So `app.asar` is handled on *every* reopen. This is exactly #696 ("still appears on taskbar reopen").

The smoking-gun log from #668 is the launcher's exact path flowing into the file-drop handler:

```
Handling file drop: /usr/lib/claude-desktop/node_modules/electron/dist/resources/app.asar
LocalAgentModeSessions.setDraftSessionFolders: 0 folders
```

Removing the path from argv means it can never reach **any** consumer — folder-drop, file-drop, skill, or DXT/MCPB — and the fix is **immune to upstream re-minification** (it touches no JS anchors).

## The change

In each launcher, the app path is no longer appended to `electron_args`. `app_path` is retained and logged for diagnostics:

```bash
# Intentionally NOT appended: app.asar sits in Electron's default
# resources/ dir next to the binary, so Electron auto-loads it. ...
log_message "App (auto-loaded by Electron): $app_path"
```

The nix build already co-locates `app.asar` into the binary's own `resources/` dir (lines ~105–143, because Chromium derives `resourcesPath` from `/proc/self/exe`), so auto-load applies there too and the same fix is correct.

## Testing

- Verified on a live `deb` install (upstream 1.9255.2) that launching the bundled Electron **with no app-path argument still loads Claude** (auto-load from `resources/app.asar`), and that pointing it at an unrelated path is ignored in favor of `resources/app.asar` — confirming the argument was redundant.
- `bash -n` on the generated launcher: valid.
- Confirmed the startup-scan guard and the guard-less `lKr` second-instance path in the 1.9255.2 minified `index.js`.
- **Not** run locally: full `deb`/`rpm`/`appimage`/`nix` package builds or CI. Relies on CI for build validation.

Fixes #696.

---
90% AI / 10% Human
Claude: root-cause investigation (argv tracing, asar inspection, live launch probe), the four-launcher patch, and this write-up.
Human: reported the recurring prompt, authorized the fix and PR, ran the privileged install-side verification.


R00tB33rMan (approved):

aaddrick (approved):
@emandel82, your root-cause diagnosis is right. The explicit `app.asar` argv was redundant for the bundled-Electron path: all four formats place the asar where Electron auto-loads it, and the second-instance collector is what kept re-attaching it on every reopen. Killing the injection in the launcher instead of chasing minified anchors is the durable call. Approving.

Review found two regressions hiding behind that argv. Both are fixed in commits I pushed to your branch (de95718, a4b8511). The big one: the launcher's orphan reaper and the doctor's "orphaned daemon" check both fingerprinted the live UI with `pgrep -f 'app\.asar'`, which only ever matched because of the argv this PR removes. Post-PR, every reopen-while-running would have killed the live Cowork daemon. I re-keyed both checks to `--class=$WM_CLASS` after checking /proc cmdline on the running deb's Electron: the flag is present on the main process, absent on zygote/helper children, and stable across all four launchers since `build_electron_args` emits it. The #693 reaper-test stubs are updated in the same push.

The smaller one: the deb/rpm global-Electron fallback boots `default_app`, the one mode where the explicit path was load-bearing. Without it, a deep link would've been eaten as a bogus app path. That branch now appends `"$app_path"` conditionally. I also rebased onto main and moved the CHANGELOG entry back under `[Unreleased]`.

296/296 BATS locally. Manual check on a deb install: the daemon survives reopen, doctor stays quiet with the UI alive, and warns once it's gone. That's one machine and one format, so the rpm/appimage/nix paths are verified by static analysis only. CI runs next; merging on green.

---

PR #701, opened 2026-06-09
fix(cowork-fallback): validateMountPath rejects $HOME symlinks on immutable distros (Silverblue/Bazzite)

### LINKED ISSUE #702: [bug]: additionalBinds silently dropped on immutable distros (Silverblue/Bazzite) — --dir missing before --bind
## Summary

On Fedora Silverblue and Bazzite (and any distro where `os.homedir()` returns a
path outside `/home`), entries in `coworkBwrapMounts.additionalBinds` and
`additionalROBinds` are **silently ignored** — the mounts never appear in
`/proc/1/mounts` inside the sandbox. `--doctor` reports the config correctly,
so the failure is invisible without manual inspection.

## Environment

- Distro: Fedora Silverblue / Bazzite (immutable, atomic)
- Installation: AppImage via GearLever
- Cowork isolation: bubblewrap (namespace sandbox)
- `os.homedir()` = `/var/home/cloud` (not `/home/cloud`)

## Steps to reproduce

1. On Silverblue/Bazzite, create `~/.config/Claude/claude_desktop_linux_config.json`:

```json
{
  "preferences": {
    "coworkBwrapMounts": {
      "additionalBinds": ["/var/home/cloud/dev"]
    }
  }
}
```

2. Restart the daemon (`pkill -f cowork-vm-service`).
3. Start a Cowork session and run:

```bash
grep '/var/home/cloud/dev' /proc/1/mounts
```

**Expected:** the bind mount appears.  
**Got:** no output — the mount was silently dropped.

## Root cause

The bwrap sandbox starts with `--tmpfs /`, making the root an empty tmpfs.
`mergeBwrapArgs()` in `scripts/cowork-vm-service.js` emits:

```
--bind /var/home/cloud/dev /var/home/cloud/dev
```

…without a preceding `--dir`. When bwrap processes `--bind`, the destination
(`/var/home/cloud/dev`) does not exist in the empty sandbox — the entire
`/var/home/` tree is absent. bwrap silently drops the mount.

The `--dir /var/home/cloud` that *would* create the parent is pushed **after**
`mergeBwrapArgs()` returns (line 1344 of `cowork-vm-service.js`) — too late,
the bind was already evaluated in arg order.

This does not affect standard distros where `$HOME` is under `/home/`, because
those paths happen to resolve correctly under the tmpfs root in the default
setup.

## Fix

Emit `--dir <dst>` immediately before each `--bind` / `--ro-bind` in the
`additionalBinds` / `additionalROBinds` loops inside `mergeBwrapArgs()`.
`bwrap --dir` creates the target path and all intermediate parents, so the
destination exists when the bind is processed.

PR #701 implements this fix.

---
Written with [Claude Code](https://claude.ai/code)

nklowns:
> **Re-targeted onto the v3.1.0 layout** — the bwrap daemon now lives at `scripts/cowork-fallback/` as the opt-in `COWORK_VM_BACKEND=bwrap` backend. This fix only affects that fallback path; the default KVM/microVM path is untouched. Single squashed commit replaces the pre-v3.0.0 history (old paths no longer exist).

## Problem

On Fedora Silverblue, Bazzite, and other immutable distros, `/home` is a symlink to `/var/home`. This causes `additionalBinds` mounts to be silently dropped at daemon startup when running the bwrap fallback backend.

`validateMountPath` calls `fs.realpathSync()` on the candidate path (resolving it to `/var/home/<user>/…`), then compares against `os.homedir()` which returns the un-resolved symlink form `/home/<user>`. Since `/var/home/cloud/dev` does not start with `/home/cloud/`, every path under home is rejected — regardless of whether the user specifies the symlink or real form.

Debug output on an affected host (Bazzite DX, `bubblewrap 0.11.0`):

```
BwrapConfig: rejected mount "/var/home/cloud/dev": Read-write mounts must be under $HOME
BwrapConfig: rejected mount "/home/cloud/dev": Read-write mounts must be under $HOME
```

Reproduction against the real config:

```js
// BEFORE
loadBwrapMountsConfig('~/.config/Claude/claude_desktop_linux_config.json', warn)
// → additionalBinds: []   ← both forms rejected, nothing reaches mergeBwrapArgs

// AFTER
// → additionalBinds: ["/var/home/cloud/dev", "/home/cloud/dev"]
```

## Fix

Resolve `$HOME` before comparing; accept paths under either form. Paths outside `$HOME` (e.g. `/opt/tools`) are still rejected.

```js
const home = os.homedir();
let realHome = home;
try { realHome = fs.realpathSync(home); } catch (_) {}
const check = resolved !== normalized ? resolved : normalized;
const underHome = (h) => check === h || check.startsWith(h + '/');
if (!underHome(home) && !underHome(realHome)) {
    return { valid: false, reason: 'Read-write mounts must be under $HOME' };
}
```

## Secondary hardening

`mergeBwrapArgs` emits `--dir path.dirname(dst)` before each `--bind` / `--ro-bind` — explicit parent creation in the empty tmpfs root, targeting the parent (never the destination itself) so single-file binds (`.gitconfig`, `/etc/hosts` overlays) keep working. Includes @aaddrick's correction from the original review round.

## Live verification (sandbox layout, Bazzite DX)

```bash
# Inside a cowork Bash session
$ echo $HOME
/home/cloud
$ ls /var/home
ls: cannot access '/var/home': No such file or directory
```

The sandbox's `$HOME` is the literal `/home/<user>` form — no `/home → /var/home` symlink exists inside it. The docs note therefore tells immutable-distro users to configure the `/home/...` form (both forms pass the validator; only `/home/...` is reachable under `~/`).

## Tests

`scripts/cowork-fallback/tests/cowork-bwrap-config.bats`: 61 → 65

- `validateMountPath` accepts both home forms on immutable distros, still rejects `/opt/tools`
- `mergeBwrapArgs` pre-creates parent dir for immutable-distro home paths
- file bind targets `--dir` at parent, not the file dst
- RO-overlay onto default mount shape

All fallback suites green: **158/158** (backend-detection 13, bwrap-config 65, bwrap-patch 7, path-translation 72, protocol 1).

## Docs

`docs/configuration.md` — documents `coworkBwrapMounts` in the bwrap-fallback section with the immutable-distro note.

Fixes #702

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -116,6 +116,16 @@ Requirements when flagged:
 
 Run `claude-desktop-unofficial --doctor` with the flag set to see the bwrap-path diagnostics. Isolation is namespace-level, not a VM — weaker than the KVM default, which is the trade for running where KVM can't. Any `COWORK_VM_BACKEND` value other than `bwrap` is a 2.x knob the official client ignores.
 
+Extra host paths can be exposed to the sandbox via `coworkBwrapMounts` (`additionalBinds` / `additionalROBinds`) in `~/.config/Claude/claude_desktop_linux_config.json`.
+
+> **Note for immutable distros (Fedora Silverblue, Bazzite):** on these
+> systems `/home` is a symlink to `/var/home` on the *host*, but the sandbox
+> has no such symlink — `$HOME` inside the sandbox is the literal
+> `/home/<user>` form. Use the same form in your config
+> (for example `"/home/cloud/dev"`, not `"/var/home/cloud/dev"`) so the
+> mount is accessible under `~/` inside the sandbox. Both forms are accepted
+> by the validator; only the `/home/...` form will appear under `$HOME`.
+
 ## Removed in v3.0.0
 
 The v3.0.0 rebase deleted the patches that read these variables. The doctor's legacy-environment check warns when any of them is still set:
--- a/scripts/cowork-fallback/cowork-vm-service.js
+++ b/scripts/cowork-fallback/cowork-vm-service.js
@@ -748,9 +748,17 @@ function validateMountPath(mountPath, opts) {
     }
 
     if (opts.readWrite) {
+        // Compare against both the configured $HOME and its fully-resolved
+        // form. On immutable distros (Fedora Silverblue/Bazzite) /home is a
+        // symlink to /var/home: os.homedir() returns /home/<user> while
+        // realpathSync() on the candidate path returns /var/home/<user>/...,
+        // so a single-form comparison rejects every path under home.
         const home = os.homedir();
+        let realHome = home;
+        try { realHome = fs.realpathSync(home); } catch (_) {}
         const check = resolved !== normalized ? resolved : normalized;
-        if (check !== home && !check.startsWith(home + '/')) {
+        const underHome = (h) => check === h || check.startsWith(h + '/');
+        if (!underHome(home) && !underHome(realHome)) {
             return { valid: false, reason: 'Read-write mounts must be under $HOME' };
         }
     }
@@ -886,19 +894,20 @@ function mergeBwrapArgs(defaultArgs, config) {
         }
     }
 
+    // Pre-create each destination's parent inside the empty tmpfs root.
+    // --dir on the parent (not the destination itself) creates all missing
+    // intermediate directories without colliding with single-file binds.
     for (const m of config.additionalROBinds) {
-        if (typeof m === 'string') {
-            result.push('--ro-bind', m, m);
-        } else {
-            result.push('--ro-bind', m.src, m.dst);
-        }
+        const src = typeof m === 'string' ? m : m.src;
+        const dst = typeof m === 'string' ? m : m.dst;
+        result.push('--dir', path.dirname(dst));
+        result.push('--ro-bind', src, dst);
     }
     for (const m of config.additionalBinds) {
-        if (typeof m === 'string') {
-            result.push('--bind', m, m);
-        } else {
-            result.push('--bind', m.src, m.dst);
-        }
+        const src = typeof m === 'string' ? m : m.src;
+        const dst = typeof m === 'string' ? m : m.dst;
+        result.push('--dir', path.dirname(dst));
+        result.push('--bind', src, dst);
     }
 
     return result;
--- a/scripts/cowork-fallback/tests/cowork-bwrap-config.bats
+++ b/scripts/cowork-fallback/tests/cowork-bwrap-config.bats
@@ -135,6 +135,32 @@ assertDeepEqual(result, { valid: true }, 'rw under home');
 	[[ "$status" -eq 0 ]]
 }
 
+@test "validateMountPath: accepts RW paths under real HOME on immutable distros (Silverblue/Bazzite)" {
+	# On Fedora Silverblue/Bazzite, /home is a symlink to /var/home.
+	# os.homedir() returns /home/<user> but fs.realpathSync resolves it to
+	# /var/home/<user>. Without resolving $HOME before comparing, both the
+	# symlink form (/home/user/dir) and the real form (/var/home/user/dir)
+	# were rejected with "Read-write mounts must be under $HOME".
+	run node -e "${NODE_PREAMBLE}
+const home = os.homedir();
+let realHome = home;
+try { realHome = require('fs').realpathSync(home); } catch (_) {}
+
+// Symlink form: must pass regardless of whether realpath differs
+const r1 = validateMountPath(home + '/dev', { readWrite: true });
+assert(r1.valid, 'symlink-form home path must be accepted: ' + home + '/dev');
+
+// Real path form: must also pass when realHome differs from home
+const r2 = validateMountPath(realHome + '/dev', { readWrite: true });
+assert(r2.valid, 'realpath-form home path must be accepted: ' + realHome + '/dev');
+
+// Non-home path must still be rejected
+const r3 = validateMountPath('/opt/tools', { readWrite: true });
+assert(!r3.valid, '/opt/tools must still be rejected');
+"
+	[[ "$status" -eq 0 ]]
+}
+
 @test "validateMountPath: accepts RO paths anywhere (not forbidden)" {
 	run node -e "${NODE_PREAMBLE}
 const r1 = validateMountPath('/opt/my-tools');
@@ -460,8 +486,8 @@ const result = mergeBwrapArgs(defaults, {
     disabledDefaultBinds: []
 });
 const expected = ['--tmpfs', '/', '--ro-bind', '/usr', '/usr',
-    '--ro-bind', '/opt/tools', '/opt/tools',
-    '--ro-bind', '/nix/store', '/nix/store'];
+    '--dir', '/opt', '--ro-bind', '/opt/tools', '/opt/tools',
+    '--dir', '/nix', '--ro-bind', '/nix/store', '/nix/store'];
 assertDeepEqual(result, expected, 'ro appended');
 "
 	[[ "$status" -eq 0 ]]
@@ -477,7 +503,7 @@ const result = mergeBwrapArgs(defaults, {
     disabledDefaultBinds: []
 });
 const expected = ['--tmpfs', '/', '--ro-bind', '/usr', '/usr',
-    '--bind', home + '/data', home + '/data'];
+    '--dir', home, '--bind', home + '/data', home + '/data'];
 assertDeepEqual(result, expected, 'rw appended');
 "
 	[[ "$status" -eq 0 ]]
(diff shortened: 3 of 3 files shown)


aaddrick:
Hey @nklowns! Thanks for the PR and the companion issue. The loop refactor is a clean readability win, and the parent-creation hardening is worth shipping. I pushed 010572a to your branch with one correction: the emission is now `--dir path.dirname(dst)` instead of `--dir <dst>`.

The unconditional form was fatal for two configs that work today. Binding a single file (a `.gitconfig`, a custom `/etc/hosts`) pre-creates the dest as a directory, and bwrap dies with `Can't create file at ...: Is a directory`. Overlaying a file onto a default RO mount dies with `Can't mkdir ...: Not a directory`. The dirname form creates all intermediate parents and never collides with a file destination. I verified all three shapes live against bubblewrap 0.11.0, updated the six existing test expectations, and added two tests pinning the file-bind and RO-overlay shapes. Suite went 62 -> 64 green.

One thing I couldn't reproduce: bwrap silently dropping a bind. On every version I tested it auto-creates missing destination parents, and when it can't, it dies loudly rather than silently. The mechanism that does fit #702's symptom is the `/home -> /var/home` symlink mismatch on immutable distros: a literal `/home/cloud/...` config path lands at an unpopulated `/home` inside the sandbox while `$HOME` is `/var/home/cloud`. Session runs fine, and the `/var/home` grep finds nothing. Your docs note already pointed at that, so I rewrote it to lead with the mechanism.

Before merge, could you confirm three things on the affected Bazzite host: `bwrap --version`, the literal `additionalBinds` path from your config (was it `/home/...` or `/var/home/...`), and whether the mount now shows in `/proc/1/mounts` inside the sandbox with this branch.

---

nklowns:
Thanks for the correction — `path.dirname(dst)` is the right call. Pre-creating the destination itself as a directory would silently break `.gitconfig` / single-file bind configs, which is worse than the original problem.

Answers to your three questions, from this Bazzite DX host — but with a bigger update below.

**bwrap version:** `bubblewrap 0.11.0` — same as yours.

**The literal path in the config:**

```json
{
  "preferences": {
    "coworkBwrapMounts": {
      "additionalBinds": ["/var/home/cloud/dev", "/home/cloud/dev"]
    }
  }
}
```

Both forms — symlink and real path. Neither worked before this PR.

**Mount in `/proc/1/mounts`:** Not yet confirmed from a live cowork session, but I found the actual rejection point (see below).

---

## Root cause confirmed — `validateMountPath`, not bwrap ordering

While investigating with `COWORK_VM_DEBUG=1`, I found the paths are rejected before they ever reach `mergeBwrapArgs`. The real failure is in `validateMountPath`:

```js
// resolves symlinks on the candidate path
resolved = fs.realpathSync(normalized);

// picks resolved path when it differs from normalized
const check = resolved !== normalized ? resolved : normalized;

// compares against os.homedir(), which is NOT resolved
if (check !== home && !check.startsWith(home + '/')) {
    return { valid: false, reason: 'Read-write mounts must be under $HOME' };
}
```

On Bazzite/Silverblue, `/home` is a symlink to `/var/home`. So:

- `os.homedir()` → `/home/cloud` (symlink form, not resolved)
- `fs.realpathSync('/home/cloud/dev')` → `/var/home/cloud/dev`
- `check` = `/var/home/cloud/dev`
- `/var/home/cloud/dev`.startsWith(`/home/cloud/`) → **false → rejected**

The same happens with `/var/home/cloud/dev` directly — `os.homedir()` was never resolved, so neither form ever passes. Both paths are dropped at daemon startup; `mergeBwrapArgs` never sees them.

The daemon log confirms it:

```
BwrapConfig: rejected mount "/var/home/cloud/dev": Read-write mounts must be under $HOME
BwrapConfig: rejected mount "/home/cloud/dev": Read-write mounts must be under $HOME
```

Code-level reproduction against the real config (unpatched AppImage script):

```js
loadBwrapMountsConfig('~/.config/Claude/claude_desktop_linux_config.json', warn)
// → additionalBinds: []   ← both paths rejected, nothing reaches mergeBwrapArgs
```

With the fix applied:

```js
loadBwrapMountsConfig('~/.config/Claude/claude_desktop_linux_config.json', warn)
// → additionalBinds: ["/var/home/cloud/dev", "/home/cloud/dev"]  ← both accepted
```

## The fix (commit `2169091`)

Resolve `$HOME` before comparing, accept paths under either form:

```js
if (opts.readWrite) {
    const home = os.homedir();
    let realHome = home;
    try { realHome = fs.realpathSync(home); } catch (_) {}
    const check = resolved !== normalized ? resolved : normalized;
    const underHome = (h) => check === h || check.startsWith(h + '/');
    if (!underHome(home) && !underHome(realHome)) {
        return { valid: false, reason: 'Read-write mounts must be under $HOME' };
    }
}
```

Paths outside `$HOME` (e.g. `/opt/tools`) are still rejected.

## Live sandbox verification still needed

To confirm the mounts now appear inside the sandbox, restart Claude with this branch and run in a cowork Bash session:

```bash
grep -E '/home/cloud/dev|/var/home/cloud/dev' /proc/1/mounts
ls /var/home/cloud/dev 2>&1 | head -3
ls /home/cloud/dev 2>&1 | head -3
```

Expected: both lines appear in mounts, both directories accessible.

---
Host: `bubblewrap 0.11.0` · `/home → var/home` symlink · `os.homedir()` = `/home/cloud`

Written with [Claude Code](https://claude.ai/code)

aaddrick:
@nklowns The realpathSync($HOME) fix is the right call. I pulled the branch and ran the full bats suite: green, 65/65. It doesn't regress normal hosts or break any working config.

One thing that's more than a wording nit. The configuration.md note tells people to use the real-path form (/var/home/...), but that's backwards. A bind is only reachable via $HOME if its form matches the sandbox's $HOME, and you reported os.homedir() = /home/cloud, the symlink form. So the symlink form is the one that lines up. The note should say "use whichever form matches the sandbox $HOME."

When you restart on this branch, grab `echo $HOME` from inside a cowork Bash session along with the `/proc/1/mounts` grep. That tells us which form to document. The code's good to land on its own. This last bit is just about the docs.

---

nklowns:
You're right — the note was backwards. Fixed in `1479129`.

Ran the sandbox layout check on this Bazzite DX host (AppImage install, unpatched — no fix needed to observe the layout):

```bash
# Inside a cowork Bash session
$ echo $HOME
/home/cloud

$ grep -E 'home|var' /proc/1/mounts
(no output)

$ ls /var/home 2>&1
ls: cannot access '/var/home': No such file or directory

$ ls /home
cloud
```

So the sandbox's `$HOME` is `/home/cloud` (symlink form), `/var` is not present, and home content is not bound by default — only the empty dir exists via `--dir`. A bind configured as `/var/home/cloud/dev` would be mounted there but unreachable as `~/dev`. The `/home/cloud/dev` form is the one to use.

Updated note:

> On immutable distros `/home` is a symlink to `/var/home` on the *host*, but the sandbox has no such symlink — `$HOME` inside the sandbox is the literal `/home/<user>` form. Use the same form in your config (`"/home/cloud/dev"`, not `"/var/home/cloud/dev"`) so the mount is accessible under `~/` inside the sandbox. Both forms are accepted by the validator; only the `/home/...` form will appear under `$HOME`.

Still pending: confirming that `/home/cloud/dev` actually appears in `/proc/1/mounts` after the fix is applied and the daemon restarts. That requires patching the AppImage or a release build.

---
Written with [Claude Code](https://claude.ai/code)

aaddrick:
Post-v3.0.0 triage: the substance here is still good and still wanted — but the target moved. At the rebase the bwrap daemon was parked and returned in v3.1.0 as the opt-in `COWORK_VM_BACKEND=bwrap` fallback ([cowork dossier: **park**](https://github.com/aaddrick/claude-desktop-debian/blob/main/docs/reports/CDL-ANT-0009_patch-suite-history/dossiers/cowork.md)), so the code this PR patches now lives at [`scripts/cowork-fallback/cowork-vm-service.js`](https://github.com/aaddrick/claude-desktop-debian/blob/main/scripts/cowork-fallback/cowork-vm-service.js) — `validateMountPath` with the same unresolved-`os.homedir()` comparison is at line 708, and `cowork-bwrap-config.bats` moved to `scripts/cowork-fallback/tests/`. Your fix (resolve `$HOME`, accept either form) and the `--dir`-before-`--bind` hardening are also exactly what the re-scoped #702 asks for.

Two notes for the re-target: (1) on Silverblue/Bazzite the *default* Cowork path is now the official KVM microVM, so this fix applies when users opt into the bwrap fallback (e.g. KVM-blocked hosts, #772) — worth a line in the PR body; (2) `/home → var/home` is precisely the environment where that fallback matters, so the fix is worth landing despite the narrower audience. Please rebase onto current main with the new paths and reopen the review request — if you think the KVM default makes this not worth carrying, say so and we'll fold the analysis into #702 instead.

---

---

nklowns:
Re-targeted onto current `main` (v3.1.0 layout) per the post-v3.0.0 triage:

- Fix re-applied at `scripts/cowork-fallback/cowork-vm-service.js` — the parked daemon carried the pre-fix `validateMountPath`, so both the unresolved-`$HOME` comparison and the missing `--dir` parent hardening were still present.
- Tests ported to `scripts/cowork-fallback/tests/cowork-bwrap-config.bats` (61 → 65), including the file-bind and RO-overlay shapes from the earlier review round.
- All five fallback suites green locally: 158/158.
- Docs note moved into the `configuration.md` bwrap-fallback section, documenting `coworkBwrapMounts` there.
- Old history replaced by a single commit (`064a5d0`) — the pre-v3.0.0 paths no longer exist, so a rebase would have been all conflicts.

Agreed on the scope note: on Silverblue/Bazzite the default KVM path sidesteps this, so this matters for hosts where KVM is blocked (`COWORK_VM_BACKEND=bwrap`, e.g. Crostini in #772). I can retest under the flag on 3.1.0 from this Bazzite host if useful for confirming the carried-over daemon behaviour end-to-end.

---
Written with [Claude Code](https://claude.ai/code)

CherrieTheShifter:

PR #712, opened 2026-06-10
fix(doctor): report installed version from the package manager that owns the install

### LINKED ISSUE #711: --doctor reports the version from the wrong package manager when both dpkg and rpm are present
`scripts/doctor.sh` decides which package manager to ask based on whether `dpkg-query` exists, not on which one actually owns the install. On a Fedora host with `dpkg` installed for deb work, doctor reports whatever the dpkg database says and never consults `rpm`.

Hit this today on Nobara fc43 with claude-desktop-1.11847.5-2.0.19 installed from the DNF repo. My dpkg database still had a stale `claude-desktop 1.5354.0` record from an old deb experiment. Doctor printed `[PASS] Installed version: 1.5354.0` for an rpm install that's actually on 1.11847.5.

## Repro

has repro:

```bash
claude-desktop --doctor | grep 'Installed version'
# [PASS] Installed version: 1.5354.0

rpm -q claude-desktop
# claude-desktop-1.11847.5-2.0.19.x86_64
```

## Root cause

`scripts/doctor.sh:638` gates the "Installed package version" block on `command -v dpkg-query` and reports the dpkg answer unconditionally. There's no `rpm -q` branch at all, so rpm installs on dual-DB hosts get scored by the wrong database. This is the same false-report family as #692: a confident PASS sourced from data that doesn't describe the actual install.

## Suggested fix

Decide ownership of the real install instead of trusting whichever DB answers first. Two shapes that should work:

1. Ask who owns a known installed file: `rpm -qf` / `dpkg -S` on `/usr/lib/claude-desktop/node_modules/electron/dist/electron`, then query that manager for the version.
2. Cheaper variant: try `rpm -q claude-desktop` before `dpkg-query` when rpm owns the path.

AppImage and Nix installs would continue to hit the existing not-found warn either way.

## Note for affected users

Don't `dpkg -r` the stale record on an rpm system to clean this up. dpkg removal deletes files that are now owned by the rpm.

---

aaddrick:
## Summary

`--doctor`'s "Installed package version" block gated on `command -v dpkg-query` only and trusted the dpkg database unconditionally. On a dual-DB host (Fedora with dpkg installed for deb work), a stale dpkg record shadowed the live rpm install — doctor printed `[PASS] Installed version: 1.5354.0` for an rpm install actually on `1.11847.5-2.0.19`. Same false-report family as #692.

## Fix

Decide ownership instead of trusting whichever DB answers first:

- Probe `rpm -qf` on the installed Electron binary (`/usr/lib/claude-desktop/node_modules/electron/dist/electron`, or the launcher-provided path). Querying the **path** is the discriminator — a stale dpkg record can't own a file rpm installed. If rpm owns it, report `%{VERSION}-%{RELEASE}` from the owning package.
- Otherwise fall back to `dpkg-query -W` exactly as before.
- AppImage/Nix installs keep the existing not-found warn; hosts with no package tools stay silent.

The block is extracted into `_doctor_check_pkg_version` so it's unit-testable; the output format (`Installed version: X`) is unchanged, so nothing downstream moves.

## Testing

- `bats tests/doctor.bats` — 44/44 pass (5 new cases: the #711 repro with rpm-owned path + stale dpkg record, dpkg-only host, rpm-present-but-deb-owned fallback, neither-owns warn, tool-less silence).
- Mutation check: reordering the helper to dpkg-first makes the #711 repro test fail (`Installed version: 1.11847.5-2.0.19` assertion), so a revert can't slip past CI.
- `shellcheck -x scripts/doctor.sh` — clean.
- Live sanity run on the repro host (Nobara fc43, rpm install + stale dpkg `1.5354.0` record): `_doctor_check_pkg_version ''` now prints `[PASS] Installed version: 1.11847.5-2.0.19.fc42`.

Fixes #711

---
90% AI / 10% Human
Claude: root-cause confirmation, ownership-probe implementation, bats cases, mutation check, live-host verification
Human: found the bug while dogfooding doctor on a dual-DB host, directed the fix design

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ The format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) — 
 
 <!-- Updated automatically by check-claude-version; will be current at release time. -->
 
+### Fixed
+
+- `claude-desktop --doctor` reports the installed version from the package manager that actually owns the install (probed via `rpm -qf` on the bundled Electron binary) instead of trusting `dpkg-query` alone — rpm installs on hosts that also carry a stale dpkg record (e.g. Fedora boxes with dpkg installed as a build tool) no longer show a months-old version with a PASS. ([#712](https://github.com/aaddrick/claude-desktop-debian/pull/712), fixes [#711](https://github.com/aaddrick/claude-desktop-debian/issues/711))
+
 ## [v2.0.19] — 2026-06-10
 
 Tracks upstream Claude Desktop 1.11847.5.
--- a/scripts/doctor.sh
+++ b/scripts/doctor.sh
@@ -618,6 +618,57 @@ _doctor_check_disk_space() {
 	fi
 }
 
+# Report the installed claude-desktop version from the package manager
+# that actually owns the install (#711). On dual-DB hosts (e.g. a
+# Fedora box with dpkg installed for deb work) a stale dpkg record
+# must not shadow the live rpm install, so rpm ownership of the real
+# Electron binary is probed first: `rpm -qf <path>` succeeds only when
+# rpm installed the file, which a stale dpkg record can never claim.
+# dpkg is consulted only when rpm does not own the path.
+#
+# AppImage and Nix installs (no package owns the path) keep the
+# existing not-found warn; hosts with no package tools stay silent.
+#
+# Usage: _doctor_check_pkg_version <electron_path>
+_doctor_check_pkg_version() {
+	local electron_path="${1:-}"
+	local probe_path="$electron_path"
+	local pkg_version=''
+
+	if [[ -z $probe_path ]]; then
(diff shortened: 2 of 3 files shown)


sabiut (changes requested):
Ran this locally — 44/44 passing, ShellCheck clean, and I like that the RPM not-owned-message-on-stdout edge case has its own test.

I did spot one small gap though: the dpkg fallback still trusts removed-but-not-purged (rc) packages. dpkg-query will still return a version even when `${db:Status-Status}` is `config-files`, so an old removed package could incorrectly show as installed.

A simple fix would be to query both status and version, and only accept results where the status is installed. Worth adding a small bats test for the config-files case too.

aaddrick:
Thanks Sum,

I'll review here momentarily and push another PR

PR #713, opened 2026-06-10
fix(doctor): gate dpkg version branch on installed status (#711 follow-up)

### LINKED ISSUE #711: --doctor reports the version from the wrong package manager when both dpkg and rpm are present

aaddrick:
> **Updated in place 2026-07-11** — force-pushed over the original commit
> (`10251b1`), same fix re-anchored onto post-v3.0.0 `main`. The rework
> split the single function the original diff patched into two
> dpkg-consulting sites, so the status gate now lands in both. Original
> review context below still applies.

## Problem

`claude-desktop-unofficial --doctor` reports a stale `[PASS]` / bogus
collision message for a `.deb` that was removed but not purged. After
`apt remove claude-desktop-unofficial` (without `--purge`), dpkg keeps a
`config-files` (rc) record and `dpkg-query` still answers `${Version}`,
`${Maintainer}`, etc. for it — even though the software is no longer
installed. `--doctor` trusted that output. This is the #711 false-PASS.

## Root cause

`doctor.sh` trusted `dpkg-query` metadata without gating on install
**status**. The reference-correct pattern already exists in the file:
`_pkg_installed` (`scripts/doctor.sh:95-102`) greps `${Status}` for
`install ok installed`. Two `claude-desktop*` sites did not follow it.

The v3.0.0 rebase split the one function the original #713 diff patched
into two sites, so the re-anchor lands the same gate in both:

1. **`_doctor_check_pkg_version`** dpkg branch (`scripts/doctor.sh:747-767`)
   — queries `claude-desktop-unofficial`. An rc-state record produced a
   false `[PASS] Installed version: <stale>` and recorded a stale
   `_installed_pkg_version` (consumed by `_check_official_drift`). Direct
   #711 bug.
2. **`_check_name_collision`** (`scripts/doctor.sh:876-892`) — queries
   plain `claude-desktop` via `${Maintainer}`/`${Version}`. This matters
   more now than at #713's original writing: the transitional
   `claude-desktop` 1.16000.0 dummy that migrates legacy installs to
   `claude-desktop-unofficial` is routinely autoremoved to rc state
   post-migration, so rc records on this exact name are common. Without
   the gate, an rc dummy at `>= 1.16000` with Anthropic's repo configured
   falls through to a bogus "Anthropic's official package is also
   installed" INFO; an rc pre-rename record `< 1.16000` fires a bogus
   pre-rename WARN.

Patch-zero clean (D-002): the change is entirely in the diagnostic script;
no `app.asar` bytes touched.

## Fix

Both sites now gate on `${db:Status-Status}` being `installed` before
trusting any other `dpkg-query` field, mirroring `_pkg_installed`:

- `_doctor_check_pkg_version` queries
  `-f='${db:Status-Status} ${Version}'` and only reports / records the
  version when the status prefix is `installed `; otherwise it falls
  through to the existing AppImage/Nix "not found via dpkg/rpm" warn.
- `_check_name_collision` probes `-f='${db:Status-Status}'` first and
  `return 0` (silent) unless the status is `installed`, before the
  maintainer/version classification runs. Behavior is unchanged for a
  genuinely-installed `claude-desktop`; only rc / not-installed records
  now stay silent (the not-installed case already returned early, so no
  regression).

## Tests

`tests/doctor.bats` (80 → 83):

- Updated the two dpkg-only-host mocks to the new
  `${db:Status-Status} ${Version}` output format.
- Added an rc-state case to `_doctor_check_pkg_version` (`config-files`
  record → WARN, not PASS, and no stale version echoed).
- Extended `_stub_dpkg_query` with an optional 3rd status arg
  (default `installed`), so the four existing collision call sites are
  byte-for-byte unaffected.
- Added two `_check_name_collision` cases: a pre-rename rc record (stays
  silent instead of the pre-rename WARN) and the transitional-dummy rc
  scenario at `>= 1.16000` with the repo configured (stays silent instead
  of the bogus Anthropic INFO).

Real output:

```
$ bats tests/doctor.bats
...
ok 46 _doctor_check_pkg_version: rpm owns the path — rpm version wins over stale dpkg record (#711)
ok 47 _doctor_check_pkg_version: dpkg-only host reports dpkg version
ok 48 _doctor_check_pkg_version: dual-DB host where rpm does not own the path falls back to dpkg
ok 49 _doctor_check_pkg_version: removed-but-not-purged dpkg record (rc state) warns, not PASS (#711 follow-up)
...
ok 72 _check_name_collision: removed-but-not-purged pre-rename record (config-files) stays silent (#711 follow-up)
ok 73 _check_name_collision: removed-but-not-purged transitional dummy (config-files, >=1.16000) with repo configured stays silent (#711 follow-up)
...
83 tests, 0 failures
```

Full suite and lint:

```
$ for f in tests/*.bats; do bats "$f"; done   # 198/198 ok
$ shellcheck -x scripts/doctor.sh             # clean
```

Mutation check (revert the fix → the new test must fail):

```
# revert Change 1's status gate
not ok 49 _doctor_check_pkg_version: removed-but-not-purged dpkg record (rc state) warns, not PASS (#711 follow-up)

# revert Change 2's status gate
not ok 72 _check_name_collision: removed-but-not-purged pre-rename record (config-files) stays silent (#711 follow-up)
not ok 73 _check_name_collision: removed-but-not-purged transitional dummy (config-files, >=1.16000) with repo configured stays silent (#711 follow-up)
```

Restoring either gate makes its tests pass again.

## Notes

- **deb**: fixed (both sites). **rpm**: untouched — its branch probes the
  file via `rpm -qf`, which can't answer for an uninstalled path.
  **AppImage / Nix**: untouched — they hit the existing not-found warn.
- **rc path not staged on physical hardware in this pass.** Static
  analysis and the bats mocks say the `config-files` record now stays
  silent / warns correctly on both sites; I did not stage a real rc `.deb`
  on metal here. @sabiut previously reproduced the deb rc-state path on
  metal for the original #713.
- `${db:Status-Status}` is a dpkg ≥ 1.17 virtual field (universal on
  supported distros) — the same field the original #713 shipped, kept for
  consistency over switching to `${Status}` + grep.
- Change 2 is separable: a reviewer wanting the minimal #713 carry can drop
  it plus its two tests without affecting Change 1.
- @sabiut's open doctor refactors (#744/#745/#782) touch `scripts/doctor.sh`
  and `tests/doctor.bats` but not `_doctor_check_pkg_version` or
  `_check_name_collision` (the one #744 hit is a context call line, not an
  edit) — no logic conflict, only same-file adjacency. Re-check anchors if
  any of them merge first.

Fixes #711.

---
Co-Authored-By: Claude <noreply@anthropic.com>
85% AI / 15% Human
Claude: implementation (Claude Sonnet 5) re-authored the two status gates
and the three new/updated bats cases; planning, gates, and this review
(Claude Opus 4.8) produced the root-cause re-anchor plan, ran the mutation
and full-suite checks, and drafted this body; orchestration (Claude Fable 5)
sequenced the workflow.
Human: triaged #711, wrote the root-cause handover brief that scoped the
two-site re-anchor and the transitional-dummy relevance, and does final
review + posting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ The format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) — 
   child), scopes the impact to opt-in `COWORK_VM_BACKEND=bwrap` launches on
   Ubuntu 24.04+, and points at the tracked scoped fix.
   ([#542](https://github.com/aaddrick/claude-desktop-debian/issues/542))
+- `claude-desktop-unofficial --doctor` no longer treats a removed-but-not-purged deb (dpkg `config-files`/rc state) as installed: the version-reporting branch (`claude-desktop-unofficial`) and the name-collision classifier (`claude-desktop`) both gate on `${db:Status-Status}` being `installed`, mirroring `_pkg_installed`. A leftover record from `apt remove` without `--purge` — made common by the transitional `claude-desktop` dummy being autoremoved to rc — now warns like an AppImage/Nix install (or stays silent) instead of a stale `[PASS]`/collision message. ([#713](https://github.com/aaddrick/claude-desktop-debian/pull/713), fixes [#711](https://github.com/aaddrick/claude-desktop-debian/issues/711))
 
 ## [v3.1.0] — 2026-07-10
 
--- a/scripts/doctor.sh
+++ b/scripts/doctor.sh
@@ -748,10 +748,18 @@ _doctor_check_pkg_version() {
 	# Our deb is claude-desktop-unofficial since the Phase 3 rename;
 	# plain claude-desktop is Anthropic's official package on dpkg
 	# hosts and is not ours to report here.
+	# Gate on install status, not just version: dpkg-query still
+	# prints a version for a removed-but-not-purged package (rc /
+	# config-files state), so trusting the version alone PASSes for
+	# software no longer installed — the #711 false-PASS.
 	if command -v dpkg-query &>/dev/null; then
-		pkg_version=$(dpkg-query -W -f='${Version}' \
-			claude-desktop-unofficial 2>/dev/null) || pkg_version=''
-		if [[ -n $pkg_version ]]; then
+		local dpkg_status
+		dpkg_status=$(dpkg-query -W \
+			-f='${db:Status-Status} ${Version}' \
+			claude-desktop-unofficial 2>/dev/null) \
+			|| dpkg_status=''
+		if [[ $dpkg_status == 'installed '* ]]; then
+			pkg_version="${dpkg_status#installed }"
 			_installed_pkg_version="$pkg_version"
 			_pass "Installed version: $pkg_version"
 			return 0
@@ -866,6 +874,16 @@ _check_name_collision() {
 	fi
 
 	# (b) Is a package named claude-desktop installed, and whose is it?
+	# Gate on install status first: dpkg-query still answers
+	# ${Maintainer}/${Version} for a removed-but-not-purged record
+	# (config-files / rc state), which the transitional claude-desktop
+	# dummy makes common post-rename — classifying that as an install
+	# would warn/info about software no longer present (#711 family).
+	local dpkg_status
+	dpkg_status=$(dpkg-query -W -f='${db:Status-Status}' \
+		claude-desktop 2>/dev/null) || dpkg_status=''
+	[[ $dpkg_status == 'installed' ]] || return 0
+
 	local maintainer version
 	maintainer=$(dpkg-query -W -f='${Maintainer}' claude-desktop \
 		2>/dev/null) || maintainer=''
--- a/tests/doctor.bats
+++ b/tests/doctor.bats
@@ -644,7 +644,8 @@ _hide_pkg_tools() {
 
 @test "_doctor_check_pkg_version: dpkg-only host reports dpkg version" {
 	_hide_pkg_tools rpm
-	dpkg-query() { printf '1.11847.5'; }
+	# -f='${db:Status-Status} ${Version}': status prefix first.
+	dpkg-query() { printf 'installed 1.11847.5'; }
 
 	run _doctor_check_pkg_version ''
 	[[ $status -eq 0 ]]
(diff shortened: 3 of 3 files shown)


sabiut (approved):
Pulled the branch — 45/45 doctor.bats, 309/309 full suite, shellcheck clean.

Got to test this on metal too: my dev box has a real removed-but-not-purged package in rc state, so I pointed the helper at it. On main it false-PASSes with the stale version; on this branch it falls through to the warn like it should. Mock matches real dpkg-query output exactly.

Also re-did the mutation check by hand — dropping the gate fails the new test, so a revert can't sneak past CI.

sabiut (approved):
Re-reviewed at e882209 (the June approval predates the force-push, so treating this as a fresh review). Approving.

Verified locally, not just read:

- **Suite**: 88/88 doctor.bats green at the PR head.
- **Both gates mutation-checked**: weakening site 1's `== 'installed '*` to any-non-empty fails exactly the new rc-state test (the leaked `config-files 1.5354.0` string trips the no-PASS assert); deleting site 2's status gate fails exactly the two new stays-silent tests. Neither gate is decorative.
- **Live dpkg behavior confirmed on a real Debian host**: an actual rc-state package here (`cryptsetup`) answers `config-files 2:2.7.5-2` to `dpkg-query -W` — a version for software that isn't installed, which is the #711 mechanism verbatim. `${db:Status-Status}` reports `installed` only for genuinely installed packages.
- **Gate placement in `_check_name_collision` is behavior-preserving for the absent-package path**: before, absent → empty maintainer+version → `return 0`; now, absent → failed status probe → `return 0`. Only rc-state records change classification, which is the point.
- The `_stub_dpkg_query` extension keys the status probe on the same emptiness convention as the existing probes, so "not installed" fails all three consistently.

One non-blocking note: the CHANGELOG bullet will collide trivially with the pending doctor-campaign PRs (#744/#745/#782/#740/#781) — all just append under [Unreleased]. Mechanical keep-both whichever lands first.

Issue #323, opened 2026-03-21
Does not resize when using hyprland


lukedev45:
I am using Omarchy and when flicking through different windows the claude desktop app does not resize accordingly. The app window will resize accordingly but the actual interface will not. Refer to the video that I have created:

https://github.com/user-attachments/assets/5170d1c7-244f-45e9-a00d-e2f176b1a317

This is probably more of a hyprland problem but I might take a look at it and see how the claude desktop app resizes the window. In the video you can see that the problem is solved for as long as I have the app to the correct side of the screen on a new desktop.

aaddrick:
Hey @lukedev45 — your Hyprland resize report (and the jiggle fix you landed in #333) put you on the shortlist for this one. Totally optional favor ask.

I've got PR #538 up that adds a hybrid titlebar mode to get claude.ai's in-app topbar (hamburger / sidebar / search / nav) actually rendering and clickable on Linux. I've verified it on KDE Plasma X11 + Wayland, but Hyprland is still unchecked on the test plan and you're the obvious person to ask. The hybrid work is independent of the resize-jiggle path, so it shouldn't conflict with your #333 logic, but I haven't confirmed that on Hyprland myself.

If you've got 5 minutes: check out `experiment/linux-topbar-wco`, run `./build.sh --build appimage --clean no`, and let me know whether the in-app topbar renders and the hamburger/sidebar buttons actually click. That's the whole ask. No worries if you can't get to it.

And if you're open to me looping you in on future Hyprland test asks like this, let me know — no is a fine answer either way.

lukedev45:
Hi @aaddrick,

Given the last time you were so helpful on fixing the last bug I think it would be rude of me not to get involved with this. I will take a look at it after work and will upload a video to #538. I'd be happy to continue to help in the future too because I do use the Claude desktop app everyday and it's nice to give back to the open source community. 🙂

Issue #239, opened 2026-02-17
WebContentsView layout broken on resize and login/logout transitions


Voork1144:
## Root Cause: Chromium Layout Invalidation Bug on Linux Snap/Tile

`getContentBounds()` and `getContentSize()` return stale values during corner snap, half-screen snap, and tile operations on Linux. Traced through Chromium source to a missing `ScheduleRelayout()` call.

**Confirmed as Electron bug [#42586](https://github.com/electron/electron/issues/42586) (closed "Not Planned").**

---

### Root Cause (from Chromium source)

Three-part failure chain:

**1. KWin corner-snap does NOT set `_NET_WM_STATE_MAXIMIZED` atoms**

Corner/half snap uses `setQuickTileMode()` — a "quick tile", not maximize. No `_NET_WM_STATE_MAXIMIZED_VERT` or `_NET_WM_STATE_MAXIMIZED_HORZ` atoms are set. Only `ConfigureNotify` fires.

**2. Chromium's `OnXWindowStateChanged()` never fires for tiling**

From [`ui/ozone/platform/x11/x11_window.cc`](https://github.com/chromium/chromium/blob/main/ui/ozone/platform/x11/x11_window.cc):

\`\`\`cpp
void X11Window::OnXWindowStateChanged() {
  auto new_state = PlatformWindowState::kNormal;
  if (IsMinimized()) new_state = kMinimized;
  else if (IsFullscreen()) new_state = kFullScreen;
  else if (IsMaximized()) new_state = kMaximized;  // Checks _NET_WM_STATE atoms

  if (new_state != state_) {  // Only fires if state CHANGED
    platform_window_delegate_->OnWindowStateChanged(old_state, state_);
  }
}
\`\`\`

For corner snap: `IsMaximized()` returns false (no atoms) → `new_state` stays `kNormal` → `OnWindowStateChanged()` **never fires**.

**3. Without `OnWindowStateChanged()`, `ScheduleRelayout()` never runs**

From [`desktop_window_tree_host_platform.cc`](https://github.com/chromium/chromium/blob/main/ui/views/widget/desktop_aura/desktop_window_tree_host_platform.cc):

\`\`\`cpp
void DesktopWindowTreeHostPlatform::OnWindowStateChanged(...) {
  ScheduleRelayout();  // Calls InvalidateLayout() on the entire view tree
}
\`\`\`

[...]

Issue #84, opened 2025-06-29
Content not sized correctly for window unless resized


aaddrick:
Content of the login page and the main window don't scale correctly for the window extents unless the user manually changes the shape of the window. I expect this is because the window functions in the claude-native library aren't being called and need to be overwritten in the native stub.

DominicWatson:
Yes, for me this is a problem when using regular window manager maximize - the window does not appear to get a resize event.

If I drag to resize, the content is repainted and resized correctly. further more, right clicking the app in the title bar area shows native window manager context menu but I cannot click any of the options - this feels related.

<img width="1275" height="391" alt="Image" src="https://github.com/user-attachments/assets/ab10e688-0638-4cff-b912-50058744a4db" />

message:
Same for me on Cinnamon, but I still use `jammy`, maybe it works in newer versions.
```
[~] ➤  cat /etc/issue
Linux Mint 21.2 Victoria \n \l
```

<img width="1921" height="1199" alt="Image" src="https://github.com/user-attachments/assets/2c433e32-e25d-4960-b100-e300efb9db20" />

CupOfOwls:
Yeah same. I think this is related to #113 and #85

sandrakennel:
Found a workaround for the frozen/unresponsive resize handles: hold the Windows/Super key while hovering over the Claude Desktop window, right-click, and choose "Resize" from the window manager's context menu. On my system (GNOME) this snapped the window up into the top-left corner, barely visible — but from there the native resize/move handles worked again, so I could drag it back down and resize normally. After that, the app's own resize handles started working too.

PR #716, opened 2026-06-10
fix(frame-fix): arm refocus jiggle only on tiling WMs

### LINKED ISSUE #715: [bug]: workspace-switch jiggle fires on every alt-tab on stacking WMs, visibly nudging the window
```
rpm: claude-desktop-1.11847.5-2.0.19.fc42.x86_64

Claude Desktop Diagnostics

       Desktop: KDE
(remaining checks all PASS; full output available on request)
```

### What happened
On KDE Plasma, alt-tabbing away from and back to the Claude window makes the window visibly resize/shift for a fraction of a second, ~150 ms after each refocus. With a maximized window on a fractional-scale display the window briefly pokes past the work area edge (below the panel). The content view re-bounds at the same time.

Root cause: the frame-fix wrapper arms the workspace-switch jiggle on `blur→focus` and `hide→show` pairs unconditionally (scripts/frame-fix-wrapper.js, `armPair('blur', 'focus')` — added in #336 for Hyprland/i3/sway same-size workspace switches, #323). Alt-tab on any WM is a blur→focus pair, so `jiggleIfStale()` runs on every refocus: 100 ms debounce, `setSize(w+1, h)`, 50 ms later `setSize(w, h)`.

Main-process event log (window listeners + 25 ms bounds polling; BaseWindow id 1, maximized, KDE Plasma 6.6 Wayland, XWayland backend, 135% scaling), one real alt-tab cycle — captured on the stock 2.0.19 rpm install before I locally disabled the jiggle as a workaround (re-verification below uses an unpatched AppImage built from main, since this box's rpm is now modified):

```
+121088 blur   b=0,27,1906,1120
+122761 focus  b=0,27,1906,1120
+122861 move   b=0,26,1908,1122            <- +100 ms: jiggle setSize(w+1, h)
+122862 resize b=0,27,1906,1120  child view re-bounded to 1908x1122
+122912 move   b=0,26,1907,1122            <- +150 ms: restore setSize(w, h)
+122949 POLL   b=0,27,1906,1120  steady again
```

The same signature reproduces on a programmatic blur→focus cycle driven via KWin scripting with no Alt key involved, ruling out menu-bar/Alt handling. Swallowing only the jiggle's setSize calls in a live instance removes the visible nudge completely.

Reproduced again on an unpatched AppImage built from current main (same instrumentation), with two extra side effects: the jiggle's `setSize` on a maximized window can drop the window out of the maximized state entirely, and on XWayland with fractional scaling the +1px/−1px pair doesn't restore bounds exactly — the window drifts ~1 px in position and ~2 px in size per blur→focus cycle:

```
+1259 restore  b=0,27,1906,1120  max=true
+1360 move     b=0,26,1908,1122  max=true    <- jiggle
+1505 move     b=644,264,610,610 max=false   <- maximize state lost
+1606 move     b=643,263,613,612             <- next cycle: drift
+2780 move     b=641,261,615,614             <- and again
```

### Steps to reproduce
1. KDE Plasma (Wayland session, default XWayland backend). Open Claude, maximize.
2. Alt-tab to another window, alt-tab back.
3. Watch the window edges ~150 ms after refocus.

### Expected behavior
No programmatic resize on plain refocus on stacking WMs. Expected the jiggle to be reserved for the tiling WMs it was written for; got it on every alt-tab everywhere.

### Anything else
Proposed fix (PR follows): arm the blur→focus pair only when a tiling compositor is detected via its session env vars (`HYPRLAND_INSTANCE_SIGNATURE`, `SWAYSOCK`, `I3SOCK`, `NIRI_SOCKET`), with a `CLAUDE_TILING_WM=1`/`0` override for tiling WMs that don't expose a recognizable env var (xmonad, dwm, river, …). Detection is an allowlist on purpose: the armed-pair jiggle was only ever validated for Hyprland/i3/sway (#323/#336), and unlisted tiling WMs keep the main `resize`-event path. The hide→show pair (tray restore — not hit by alt-tab), the `resize`-event path (#323), ready-to-show jiggle (#84) and KWin moved/state handling (#239) are not affected.

kotok9:
Fixes #715

Alt-tab on a stacking WM is a blur→focus pair, so the workspace-switch jiggle (#336) ran ~100 ms after every refocus — `setSize(w+1, h)`, then `setSize(w, h)` 50 ms later, plus a content-view re-bound. On KDE Plasma this visibly nudges the window; with a maximized window on a fractional-scale display it briefly pokes below the panel, can drop the window out of the maximized state, and drifts bounds ~1–2 px per cycle. Event logs and full analysis in #715.

The jiggle exists for tiling WMs (#323) whose same-size workspace switches produce no resize events. This PR arms the `blur→focus` pair only when a tiling compositor is detected via its session env vars (`HYPRLAND_INSTANCE_SIGNATURE`, `SWAYSOCK`, `I3SOCK`, `NIRI_SOCKET`), covering the WMs the jiggle was written and validated for. Tiling WMs without a recognizable env var (xmonad, dwm, river, …) can opt back in with `CLAUDE_TILING_WM=1` (documented in docs/configuration.md); `CLAUDE_TILING_WM=0` forces it off; unrecognized values warn and fall back to auto-detection. Unchanged: the `hide→show` pair (tray restore — not hit by alt-tab, keeps the fixChildBounds safety net), the `resize`-event path (#323), the ready-to-show jiggle (#84), KWin moved/state-change handling (#239).

## Testing

All on KDE Plasma 6.6 (Fedora 43, Wayland session, XWayland backend), with the main process instrumented via the Node inspector (window event listeners + 25 ms bounds polling).

- Before (unpatched AppImage built from main): blur→focus cycles produce move/resize pulses at +100/+150 ms after every `focus`; the maximized-window jiggle also dropped the window out of the maximized state, and bounds drifted ~1–2 px per cycle under fractional scaling. The same real-alt-tab signature was first captured on a stock 2.0.19 rpm install (logs in #715).
- After (AppImage built from this branch): ~15 real alt-tab cycles on a maximized window plus programmatic minimize/restore cycles — zero geometry events after `focus`; the only resize in the log is the maximize itself. No visible nudge.
- Control: launching the same AppImage with `HYPRLAND_INSTANCE_SIGNATURE=x` or `CLAUDE_TILING_WM=1` arms the pair again — `[Frame Fix] Tiling WM: true`, jiggle pulses back at +100/+150 ms after focus. `CLAUDE_TILING_WM=0` with the Hyprland var set keeps it off; an unrecognized value logs the warning and falls back to detection.
- `./build.sh --build appimage --clean no` succeeds; `node --check scripts/frame-fix-wrapper.js`; BATS suite 308/308.

---
~95% AI / ~5% Human
Claude: root-cause investigation (live main-process instrumentation), fix, testing, PR text
Human: symptom report, on-device verification of fix candidates, adversarial review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) — 
 
 ### Fixed
 
+- The workspace-switch jiggle's blur→focus armed pair is now armed only when a tiling compositor is detected via its session env vars (`HYPRLAND_INSTANCE_SIGNATURE`, `SWAYSOCK`, `I3SOCK`, `NIRI_SOCKET`); `CLAUDE_TILING_WM=1`/`0` overrides the detection for compositors without a recognizable env var. On stacking WMs the pair fired on every alt-tab and the 1px jiggle visibly nudged the window and re-bounded the content view ~100 ms after each refocus — most noticeable on KDE Plasma with a maximized window. The hide→show pair (tray restore), the `resize`-event path and the KWin `moved`/state-change fixes are unchanged. ([#716](https://github.com/aaddrick/claude-desktop-debian/pull/716), fixes [#715](https://github.com/aaddrick/claude-desktop-debian/issues/715))
 - `claude-desktop --doctor` reports the installed version from the package manager that actually owns the install (probed via `rpm -qf` on the bundled Electron binary) instead of trusting `dpkg-query` alone — rpm installs on hosts that also carry a stale dpkg record (e.g. Fedora boxes with dpkg installed as a build tool) no longer show a months-old version with a PASS. ([#712](https://github.com/aaddrick/claude-desktop-debian/pull/712), fixes [#711](https://github.com/aaddrick/claude-desktop-debian/issues/711))
 
 ## [v2.0.19] — 2026-06-10
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -16,6 +16,7 @@ Model Context Protocol settings are stored in:
 | `CLAUDE_USE_WAYLAND` | unset (auto) | Force the display backend on Wayland: `1` = native Wayland, `0` = XWayland. Unset auto-detects per compositor (only Niri defaults to native Wayland). See [Wayland Support](#wayland-support) below. |
 | `CLAUDE_MENU_BAR` | unset (`auto`) | Controls menu bar behavior: `auto` (hidden, Alt toggles), `visible` / `1` (always shown), `hidden` / `0` (always hidden, Alt disabled). See [Menu Bar](#menu-bar) below. |
 | `CLAUDE_TITLEBAR_STYLE` | unset (`hybrid`) | Controls window decoration style: `hybrid` (system frame + in-app topbar), `native` (system frame, no in-app topbar), `hidden` (frameless WCO — broken on X11, kept for diagnostics). See [Titlebar Style](#titlebar-style) below. |
+| `CLAUDE_TILING_WM` | unset (auto) | Override tiling-WM detection for the workspace-switch refocus jiggle: `1` = treat the compositor as tiling (arm the blur→focus jiggle), `0` = treat as stacking (never jiggle on refocus). Unset auto-detects via `HYPRLAND_INSTANCE_SIGNATURE` / `SWAYSOCK` / `I3SOCK` / `NIRI_SOCKET`; unrecognized values warn and fall back to auto-detection. Set `1` on tiling WMs without a recognizable env var (xmonad, dwm, river, …) if window content goes stale on workspace switches. |
 | `COWORK_VM_BACKEND` | unset (auto-detect) | Force a specific Cowork isolation backend: `kvm` (full VM), `bwrap` (bubblewrap namespace sandbox), or `host` (no isolation). See [Cowork Backend](#cowork-backend) below. |
 
 ### Wayland Support
--- a/scripts/frame-fix-wrapper.js
+++ b/scripts/frame-fix-wrapper.js
@@ -90,6 +90,26 @@ console.log(`[Frame Fix] Close-to-tray: ${CLOSE_TO_TRAY ? 'on' : 'off'}`);
 const KEEP_AWAKE = process.env.CLAUDE_KEEP_AWAKE !== '0';
 console.log(`[Frame Fix] Keep awake: ${KEEP_AWAKE ? 'on (default)' : 'suppressed (CLAUDE_KEEP_AWAKE=0)'}`);
 
+// Tiling-WM detection for the workspace-switch jiggle (the
+// blur→focus armPair below). Hyprland, sway, i3 and niri export
+// their session sockets/signatures into the environment; stacking
+// WMs (KWin, Mutter) set none of these. Tiling WMs without a
+// recognizable env var (xmonad, dwm, river, ...) can opt back in
+// with CLAUDE_TILING_WM=1; CLAUDE_TILING_WM=0 forces the refocus
+// jiggle off.
+const rawTilingWm = process.env.CLAUDE_TILING_WM;
+if (rawTilingWm !== undefined && rawTilingWm !== ''
+  && rawTilingWm !== '1' && rawTilingWm !== '0') {
+  console.warn(`[Frame Fix] Unknown CLAUDE_TILING_WM value '${rawTilingWm}', falling back to auto-detection. Valid: 1, 0`);
+}
+const TILING_WM = rawTilingWm === '1' ? true
+  : rawTilingWm === '0' ? false
+  : !!(process.env.HYPRLAND_INSTANCE_SIGNATURE
+    || process.env.SWAYSOCK || process.env.I3SOCK
+    || process.env.NIRI_SOCKET);
+console.log(`[Frame Fix] Tiling WM: ${TILING_WM}`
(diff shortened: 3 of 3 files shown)


aaddrick:
Good work, @kotok9. The diagnosis and the evidence hold up. I ran the env-detection tri-state across ten permutations and it behaves exactly as documented, `node --check` passes, and the #715 symptom on KDE is real. Before I merge I want to lay out the whole history in one place, because this is a clear net win but it isn't the root-cause fix, and you might want first crack at that part.

Here's the chain. The jiggle exists for #323: on Hyprland, a same-size workspace switch produces no resize event, so the content view goes stale. It got fixed in stages. #331 added the resize-event path, then #336 added the debounced blur→focus and hide→show armed pairs. #336 was picked over the alternative in #333 *specifically because* #333's blur/focus jiggle fired on every Alt+Tab on stacking WMs. But #336 armed blur→focus unconditionally anyway, betting that the `fixChildBounds()`-found-no-mismatch guard made it a harmless no-op on stacking WMs. Its own test plan claimed no visible flicker on Alt+Tab on KDE. #715 is the proof that bet doesn't hold under fractional scaling with a maximized window: the 1px `setSize` poke becomes visible. So this PR finishes what #336 intended, just with env detection instead of the leaky guard.

The catch is what keeps this from being the final answer. blur→focus is a leaky proxy. It can't tell apart (a) the same-size workspace switch the jiggle is actually for, (b) an Alt+Tab on a stacking WM, which is #715 and is what this PR fixes, and (c) a mouse re-enter under focus-follows-mouse. That last one is already a known category here. #416 / #589 cover it, and #416 names "i3 with focus_follows_mouse" specifically. On i3/sway/Hyprland with sloppy focus, which is a common tiling setup and exactly this PR's allowlist, blur→focus fires on every pointer re-entry. So the gated jiggle still runs on mouse-enter there. It's pre-existing since #336, not introduced here, and usually invisible on a tiled window. But it's the same #715 symptom on a floating window or under fractional scaling. The fix relocates the residual from stacking WMs to tiling-plus-sloppy-focus rather than removing it.

I'm still happy to land it, because as a share of affected users it's clearly positive. The stacking-WM Alt+Tab case covers KDE, GNOME, Cinnamon, Xfce, the large majority, and it's universally visible. This kills it. The residual needs a tiling WM, focus-follows-mouse, and a floating or fractional-scale window all at once, which is a much smaller slice and usually imperceptible. Net, far more people stop seeing the nudge than keep seeing it.

The longer-term shape is to drop the proxy itself. The clean version triggers the jiggle on the actual condition, only when bounds genuinely drifted, or skips it when the window is maximized or floating, which would also close the maximized hide→show poke that's still armed on every WM. blur→focus standing in for "workspace switch" is the part that leaks. I'm good to merge this as-is for the net win. Before I do though, @kotok9, do you want to take a swing at the more holistic version, dropping the blur→focus proxy or adding a bounds-actually-drifted-or-not-maximized guard? Or would you rather land this now and track the proxy redesign as its own issue? Either's fine, your call. Happy to open the tracking issue and point at the spots if you'd rather hand it off.

---

kotok9:
Thanks @aaddrick — your breakdown matches what I found, and the leaky-proxy framing is the right one: blur→focus can't separate a real workspace switch from an Alt+Tab or a sloppy-focus pointer re-entry, so it's a stand-in, not a signal.

My preference is to land this as-is now. The visible jump on Alt+Tab hits the stacking-WM majority and shows on every refocus, so removing it is a real UX win even as a workaround over a workaround — I'd rather not hold that behind the larger redesign.

On the holistic version, I'd rather hand it off for now. I don't yet understand the tiling-WM side well enough — specifically why and when the window actually gets resized on a same-size workspace switch — to do the proxy redesign justice, so I'd be guessing. If you open the tracking issue and point at the spots, that's the right home for it.

Two things worth folding into that issue:

- The blind 1px poke fires even when `fixChildBounds()` finds no mismatch. Dropping it, or gating it on bounds-actually-drifted, looks like the real fix — pending a check of whether the `resize`-event path from #331 already covers the stale-cache case on current Electron.
- `hide→show` is still armed on every WM, so the poke still runs on tray-restore. I checked this on KDE with a maximized window: on restore it fires a brief ~1–2px wobble (the window stays maximized on this path here, so I'm not reproducing a maximize-drop via the tray — just the poke running with nothing stale to fix). A skip-when-maximized / only-when-drifted guard would remove it.

Happy to test candidate patches on KDE + fractional scaling whenever someone picks it up.

aaddrick:
Hey @kotok9,

I appreciate it. I'll action here in a bit. I'll put you on my shortlist for testing too.

Thanks again!

aaddrick:
Closing as obsoleted by the v3.0.0 rebase — through no fault of this PR, which was a careful, well-instrumented fix. The target no longer exists: `frame-fix-wrapper.js` was deleted wholesale (matrix verdict **delete**; [frame-fix-wrapper dossier](https://github.com/aaddrick/claude-desktop-debian/blob/main/docs/reports/CDL-ANT-0009_patch-suite-history/dossiers/frame-fix-wrapper.md)), because the official Linux build uses the system frame and owns its own window behavior — the jiggle and the whole apparatus it lived in are gone, and #715 was closed on that basis. Your live main-process instrumentation confirmed the delete verdict was the right call, so this work wasn't wasted.

If window-nudging on alt-tab ever reappears on a `claude-desktop-unofficial` 3.x build, rebase and reopen (or comment on #715) — with the wrapper gone it would be a new upstream bug, and your event-log methodology is exactly how to pin it.

---

---

Issue #122, opened 2025-10-21
is there a way to display the tray icon


rcspam:
Hi,

Bravo for the work.
Is there a way to display the tray icon  (KDE plasma)
Thanks

Issue #66, opened 2025-04-07
Processes hanging when app is closed


MischaPanch:
I used the AppImage based install. When I close Claude, there are still about 10 electron processes hanging that I need to kill manually.

aaddrick:
Apologies about the delay. Delete ~/claude-desktop-launcher.log, recreate the scenario, and take a peak if anything pops out. Feel free to drop the contents here.

MischaPanch:
I used the newest release. Recreate by starting the app image, then right-click and close on the tray.

![Image](https://github.com/user-attachments/assets/8a878e95-0429-453b-bef5-a10f6fdf15fc)

And here the log

```
getWindowArguments []
App is not installed, not enabling auto-updates
(node:748422) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 destroyed listeners added to [WebContents]. MaxListeners is 10. Use emitter.setMaxListeners() to increase limit
(Use `electron --trace-warnings ...` to show where the warning was created)
```

PR #717, opened 2026-06-11
fix(tray): quit on main-window close when tray disabled on Linux

LiukScot:
## Problem

On Linux, closing the main window does **not** quit Claude Desktop — it hides to a background process that only `Ctrl+Q` can actually kill. The window disappears, no tray icon is shown (tray disabled), yet the Electron process keeps running.

## Root cause

Upstream's main-window `close` handler only quits-on-close (instead of `preventDefault()` + `hide()`) when `process.platform==="win32"` **and** the tray is disabled:

```js
n.on("close",C=>{
  if(Wl())return;
  if(io&&!hi("menuBarEnabled")){           // io === process.platform==="win32"
    S.info("Quitting app on main window close since tray is disabled"),cd();return
  }
  C.preventDefault(); /* ...hide()... */
})
```

On Linux `io` is always `false`, so the handler always falls through to hide — the quit branch is unreachable. Only `Ctrl+Q` (which sets the quitting flag, making `Wl()` truthy) exits.

## Fix

New `patch_close_quits_when_tray_disabled()` in `scripts/patches/tray.sh` (called from `app-asar.sh` after `patch_menu_bar_default`) widens the platform guard from win32-only to **every non-macOS platform**:

```
io  ->  process.platform!=="darwin"
```

So a tray-less Linux session quits on window close; macOS keeps its dock-resident convention; Windows behaviour is unchanged (`!=="darwin"` is still true there).

Per the minified-JS guidelines, the patch:
- **anchors on the stable `"Quitting app on main window close since tray is disabled"` log string**, not the minified identifiers (`io`/`hi`/`S`/`cd` rename between releases);
- **asserts the guard site is unique** before rewriting (counts `grep -o` matches, since minified JS is one physical line and `grep -c` would always report 1) and bails on 0 or >1;
- **verifies the substitution landed** afterwards, emitting `WARNING:` if the PCRE extraction and ERE sed ever diverge.

Note: when `menuBarEnabled` is *unset*, upstream's defaults object merges `menuBarEnabled:!0` (true), so the guard stays false and close still hides — quit only fires when the user has **explicitly** disabled the tray. Intended behaviour.

## Testing

Validated hands-on against both shipped bundles by extracting `app.asar` and running the function on a build-dir copy:

- **1.11847.5** (current build target): finds guard `Eo`, rewrites exactly one site, `node --check` passes.
- **1.9659.2**: finds guard `io`, applies correctly.
- **Idempotency**: second run prints `already patched`, byte-identical output.
- **Ambiguity guard**: an injected duplicate log-string site makes it print `Expected 1 close-handler guard, found 2 — skipping` and apply nothing.

---
~85% AI / ~15% Human
Claude: root-caused the win32-gated close handler by extracting and inspecting the asar, wrote the patch function + hardening, validated against both bundle versions.
Human: reported the bug (window stays in background, only Ctrl+Q quits) and the tray-disabled context, authorized the PR.


aaddrick:
@LiukScot thanks for this. The bug is real and your writeup nailed it. Upstream's close handler is win32-gated, so on Linux there's nothing that ends the process and it strands. Anchoring the patch on the stable log string was the right technique too. I verified all of that locally and the fixtures behave exactly as you described.

The catch is the layer. The project already owns Linux close behavior in `scripts/frame-fix-wrapper.js`. CLOSE_TO_TRAY (default on Linux) does `preventDefault()`+`hide()` on close (#448), and `CLAUDE_QUIT_ON_CLOSE=1` makes it quit by calling `app.quit()` first and riding upstream's own quit-in-progress short-circuit so the window closes cleanly (#623). So a user who wants close-to-quit can already set `CLAUDE_QUIT_ON_CLOSE=1` today. #717 patches a second close handler in the bundled `index.js` that runs alongside the wrapper's. That's redundant, and it only covers the narrow case where the tray is explicitly turned off (`menuBarEnabled` is actually the tray toggle, and it defaults on).

The bigger thing it doesn't fix is the zombie. The wrapper's CLOSE_TO_TRAY is gated only on platform plus env (`frame-fix-wrapper.js:80`), never on whether a tray actually exists. So on Linux it hides-to-tray on close even when there's no tray host to come back from. That's the real problem on vanilla GNOME without an AppIndicator extension, where the tray is "on" but invisible, and it hits the tray-disabled case too. Same complaint as #66 and #122, and it affects more people than the explicit-disable path.

The cleaner fix lives one layer up. Gate the wrapper's CLOSE_TO_TRAY on the tray being genuinely available: `menuBarEnabled !== false` AND a StatusNotifier host actually registered.

```js
// trayAvailable = menuBarEnabled !== false  &&  a StatusNotifier host is registered
const CLOSE_TO_TRAY = process.platform === 'linux'
  && process.env.CLAUDE_QUIT_ON_CLOSE !== '1'
  && trayAvailable;
```

The host check is a one-line session-bus probe in the launcher, same pattern as the keyring probes at `launcher-common.sh:155-173`: `org.kde.StatusNotifierWatcher` Get `IsStatusNotifierHostRegistered`, exported as an env the wrapper already reads. When no tray is available, fall through to the existing #623 quit path. That fixes both the explicit-disable and the GNOME-no-host cases in one place, keeps close-to-tray working for everyone with a real tray (KDE, Ubuntu/GNOME with AppIndicator), and avoids a second close handler. One caveat worth a note: an SNI host can register slightly after launch, so a launch-time probe wants a sanity check or a bias toward keeping the window.

I'm going to close this one out, since the fix really wants to live in the wrapper rather than run alongside it — but that's a redirect, not a no. Would you be willing to pick it up as a new PR against `frame-fix-wrapper.js` (the CLOSE_TO_TRAY gate at line 80 plus the close branches around 356-389) plus the launcher probe? I'm happy to point at the exact spots or pair on it if that's easier. Your diagnosis is what surfaced all of this, and the credit carries straight over.

---

LiukScot:
Picked it up as you suggested — opened #732 against `frame-fix-wrapper.js` + the launcher probe, not a second close handler.

The gate now reads `CLOSE_TO_TRAY = linux && CLAUDE_QUIT_ON_CLOSE !== '1' && CLAUDE_TRAY_AVAILABLE !== '0'`. The launcher computes `CLAUDE_TRAY_AVAILABLE` in `_detect_tray_available()` (same shape as `_detect_password_store`): `0` when `menuBarEnabled` is off **or** no StatusNotifier host is registered on the session bus, `1` otherwise. So the vanilla-GNOME-no-AppIndicator case (#66/#122) falls through to the existing #623 quit path, and KDE / GNOME-with-AppIndicator keep close-to-tray.

On the launch-time race you flagged: a successful probe (or an ambiguous reply) biases toward keeping the window; only a bus name with no owner reads as "no host", which is the stable vanilla-GNOME signal. Kept it desktop-agnostic (the `org.kde.*` name is the freedesktop spec) and added bats coverage for the four states.

Thanks for the precise redirect — the line/anchor pointers made it quick.

---

Issue #718, opened 2026-06-12
[bug]: nix build failure on d2ce046


fdnt7:
### Version (`claude-desktop --doctor` output)

```shell
N/A
```

### What happened

On d2ce046, the nix flake failed to build in NixOS x86-64.

### Steps to reproduce

```console
$ nix build github:aaddrick/claude-desktop-debian/d2ce0466315033938ea6c2066d424239feffabf0 
```

### Expected behavior

The build should succeed and output a result `symlink` in the working directory.

### Logs / errors

```shell
Running phase: patchPhase
@nix { "action": "setPhase", "phase": "patchPhase" }
Running phase: updateAutotoolsGnuConfigScriptsPhase
@nix { "action": "setPhase", "phase": "updateAutotoolsGnuConfigScriptsPhase" }
Running phase: configurePhase
@nix { "action": "setPhase", "phase": "configurePhase" }
no configure script, doing nothing
Running phase: buildPhase
@nix { "action": "setPhase", "phase": "buildPhase" }
[1;36m--- Architecture Detection ---[0m
Detecting system architecture...
Detected machine architecture: x86_64
Configured for amd64 (x86_64) build.
Target Architecture: amd64
[1;36m--- End Architecture Detection ---[0m
[1;36m--- Distribution Detection ---[0m
Detecting Linux distribution family...
Warning: Could not detect distribution family
  AppImage build will still work, but native packages (deb/rpm) may not
Distribution: 
Distribution family: unknown
[1;36m--- End Distribution Detection ---[0m
Running as user: nixbld (Home: /build)
System Information:
Distribution: 
Distribution family: unknown
Target Architecture: amd64
[1;36m--- Argument Parsing ---[0m
Selected build format: nix
Cleanup intermediate files: no
[1;36m--- End Argument Parsing ---[0m
[1;36m--- Download the latest Claude executable ---[0m
Using local Claude installer: /build/Claude-Setup.exe
Local installer copied to build directory
Extracting resources from Claude-Setup-x64.exe into separate directory...

7-Zip [64] 17.05 : Copyright (c) 1999-2021 Igor Pavlov : 2017-08-28
p7zip Version 17.05 (locale=C,Utf16=off,HugeFiles=on,64 bits,12 CPUs x64)

Scanning the drive for archives:
  0M Scan /build/build/                       1 file, 203032224 bytes (194 MiB)

Extracting archive: /build/build/Claude-Setup-x64.exe
  0% Open         --
[...]

PR #722, opened 2026-06-16
fix(config): patch --add-dir filter to handle SDK bundled twice in 1.12603.1

marveon:
## Summary

- In Claude Desktop 1.12603.1 the Claude Code SDK is bundled in two places (chat panel and Code/Agent panel), so `for(let F of e)Y.push("--add-dir",F)` appears twice in `index.js`
- The previous patch asserted exactly 1 match and fatally exited on 2, breaking the build
- Fix: accept 1–2 sites (still fatal if >2), and switch `code.replace()` to `code.split().join()` so all occurrences get the `.asar` filter applied

## Test plan

- [ ] `./build.sh --build deb --clean no` completes without error on 1.12603.1
- [ ] Build log shows `Filtered --add-dir dispatch (for-of variant, 2 site(s))`

---
95% AI / 5% Human
Claude: diagnosed the double-match root cause, wrote the fix, tested the build
Human: triggered the build, identified the failure, requested the commit and PR

### DIFF
--- a/scripts/patches/config.sh
+++ b/scripts/patches/config.sh
@@ -206,12 +206,14 @@ let patchCount = 0;
         process.exit(1);
     }
 
-    // Count assertion: exactly 1 match expected
+    // Count sites — 2 is expected when the SDK is bundled twice
+    // (e.g. chat panel + Code/Agent panel each embed a copy).
     const escaped = match[0].replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
     const allMatches = code.match(new RegExp(escaped, 'g'));
-    if (allMatches && allMatches.length > 1) {
+    const siteCount = allMatches ? allMatches.length : 1;
+    if (siteCount > 2) {
         console.error('FATAL: --add-dir pattern matches ' +
-            allMatches.length + ' times (expected 1).');
+            siteCount + ' times (expected 1–2).');
         process.exit(1);
     }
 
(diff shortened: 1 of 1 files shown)


aaddrick:
Your diagnosis here is spot on, @marveon. 1.12603.1 bundles the Claude Code SDK per-panel now, so the `--add-dir` dispatch loop shows up twice, and the old #650 "exactly 1 match or FATAL" assertion sees 2 and kills the build. That's exactly the root cause. It matches @typedrat's independent analysis in #718/#723, so you both landed on the same thing from different directions.

I want to be fair about that old assertion too: the "exactly 1 or bail" check isn't a bug. It's a deliberate safety idiom we use so a new second site surfaces loudly instead of getting silently no-op'd. It did its job here. The reason you found this at all is that the check refused to guess.

Here's why I'm landing #723 instead, and it's a structural call rather than a correctness one. #722 uses `code.split(match[0])`, which splits on the literal text of the first loop, so it only filters occurrences that are byte-identical to it. For this release that's fine. You can actually prove the two loops are byte-identical from the #718 FATAL itself, since the old count path only reports "2" when the second loop matches the first byte-for-byte. So #722 genuinely works today. But if a future re-mangle gives the two loops different variable names, split quietly skips the second one and leaves it unfiltered. That's a silent regression of the #649 .asar guard, and it's the exact failure mode `docs/learnings/patching-minified-js.md` warns about: anchor on structure, not on a copy of one site. I checked it on a fixture with differing names, and #722 filters one loop and leaves the second raw. #723 takes the same diagnosis but rebuilds each loop from its own captured vars, so it handles differing names and a mixed for-of/forEach release, scales past two sites without a hard cap, and ships a regression test plus a CHANGELOG entry.

So I'm landing #723 for the durability, with your diagnosis credited. The credit for pinning the cause is yours. Sorry for the collision with the parallel fix. Closing this in favor of #723.

---

PR #732, opened 2026-06-20
fix(tray): quit on main-window close when no StatusNotifier host (Linux)

### LINKED ISSUE #66: Processes hanging when app is closed

LiukScot:
## Problem

On Linux, closing the main window hides the app to the system tray (close-to-tray, #448). But the wrapper gates that only on `process.platform === 'linux'` and `CLAUDE_QUIT_ON_CLOSE` — never on whether a tray host actually exists.

On vanilla GNOME without an AppIndicator extension, `menuBarEnabled` is on but **no StatusNotifier host renders the icon**. The window hides into a tray that isn't there, stranding an invisible background process that only `Ctrl+Q` can kill. Same complaint as #66 and #122.

This is the redirect from #717: the fix belongs in the wrapper's close-to-tray gate plus a launcher probe, not in a second close handler in the bundled `index.js`.

## Approach

- **`scripts/launcher-common.sh`** — new `_detect_tray_available()` (mirrors `_detect_password_store`): returns `0` when the tray is off (`menuBarEnabled: false`) **or** no StatusNotifier host is registered on the session bus, `1` otherwise. `setup_electron_env` exports `CLAUDE_TRAY_AVAILABLE`. A successful D-Bus reply biases toward keeping the window (a host can register shortly after launch); only a name with no owner — vanilla GNOME — reads as no host.
- **`scripts/frame-fix-wrapper.js`** — one term added to the existing gate: `&& process.env.CLAUDE_TRAY_AVAILABLE !== '0'`. A tray-less session falls through to the existing `CLAUDE_QUIT_ON_CLOSE` → `app.quit()` path (#623). No second close handler.

Desktop-agnostic by design: `org.kde.StatusNotifierWatcher` is the freedesktop StatusNotifierItem spec, not KDE-specific — on GNOME it's provided by an AppIndicator extension.

## Tests

New `bats` cases in `tests/launcher-common.bats`: tray off → `0`; tray on + host → `1`; tray on + no host → `0`; no config (default on) + host/no-host; and `setup_electron_env` exporting the var. Full suite (97 tests) green; `shellcheck -x` clean.

## Verification

`_detect_tray_available()` was run from the built AppImage artifact against a real GNOME (Wayland) session bus across all four states (AppIndicator extension on/off × `menuBarEnabled` on/off) — each returned the expected value, confirmed live by toggling the extension and watching `IsStatusNotifierHostRegistered`.

Fixes #66, #122

---
75% AI / 25% Human
Claude: launcher probe + wrapper gate, bats tests, local build & verification
Human: original diagnosis (#717), design direction, testing on a real GNOME box, review


LiukScot:
Heads-up on the red **Build Packages (amd64 - deb)** check — it's the pre-existing `main` breakage, not this PR.

- The failure is in the minified-JS patches: `FATAL: Failed to patch cowork platform gate for Linux`, plus `could not resolve tray menu function (inline + fallback both failed)` and `menuBarEnabled pattern not found or already patched`. Upstream re-minified and those anchors no longer match.
- This diff touches only `scripts/launcher-common.sh`, `scripts/frame-fix-wrapper.js`, and `tests/launcher-common.bats` — nothing under `scripts/patches/`.
- The `CI` workflow on `main` has been failing for the last 3 runs (2026-06-17 / 18 / 19) for the same reason, so the red is inherited.

Everything this PR actually exercises is green: BATS, Shellcheck, Codespell, Test Flags Parsing.

---

sabiut (approved):
Approving from my side, as I'm only reviewing tests/launcher-common.bats, which is the only file within my area.

One non-blocking observation on the test file: the test case "tray on + no host (vanilla GNOME) -> 0" doesn't actually verify the "no host" scenario. _detect_tray_available only checks the exit status of dbus-send and ignores the boolean value returned. With the current fail stub, the test is really validating that a failed D-Bus call returns 0.

The actual edge case we're interested in, where the watcher is running but IsStatusNotifierHostRegistered returns false, isn't currently covered by a test. It may be worth either renaming the test to better reflect what it's validating, or adding a separate test that returns exit code 0 with a boolean false response and asserting that it still returns 0. Either approach is fine as a follow-up.

aaddrick:
Closing as obsoleted by the v3.0.0 rebase. All three pieces this PR builds on were removed: `frame-fix-wrapper.js` (deleted — [dossier](https://github.com/aaddrick/claude-desktop-debian/blob/main/docs/reports/CDL-ANT-0009_patch-suite-history/dossiers/frame-fix-wrapper.md)), `CLAUDE_QUIT_ON_CLOSE` (removed — the official build moved close-to-tray behavior into **Settings ▸ General ▸ System Tray**, off = quit on close), and the `_detect_password_store` pattern the launcher probe mirrored (also removed; [v3.0.0 CHANGELOG](https://github.com/aaddrick/claude-desktop-debian/blob/main/CHANGELOG.md)). The good news is the UX gap this solved now has a supported answer: vanilla-GNOME users without an AppIndicator extension can turn the tray toggle off and get quit-on-close natively.

The residual idea — auto-detecting a missing StatusNotifier host instead of making the user flip a toggle — is upstream product behavior now, not something we can gate under patch-zero ([D-002](https://github.com/aaddrick/claude-desktop-debian/blob/main/docs/decisions.md#d-002--rebase-onto-the-official-linux-deb-patch-zero)). If you think there's a launcher-side angle that still applies (e.g. a `--doctor` warning when the tray is enabled but no SNI host exists — that part of your D-Bus probe work would translate directly), rebase and reopen, or open an issue and I'll gladly look again.

---

---

PR #734, opened 2026-06-22
Fix Linux startup hang: stub Windows-only native methods (#729)

### LINKED ISSUE #729: [bug]: hangs indefinitely, app window never shows up

```shell
Claude Desktop Diagnostics

       Desktop: KDE
            (disables global hotkeys)
[PASS] Electron: v41.7.2 (/nix/store/0hiz6nfzgi3qixgpqcmsbfixnrhyi4j1-claude-desktop-1.14271.0/lib/claude-desktop/electron/electron-wrapper)
[FAIL] Chrome sandbox: perms=777, owner=root
       Fix: sudo chown root:root /nix/store/0hiz6nfzgi3qixgpqcmsbfixnrhyi4j1-claude-desktop-1.14271.0/lib/claude-desktop/electron/chrome-sandbox
            sudo chmod 4755 /nix/store/0hiz6nfzgi3qixgpqcmsbfixnrhyi4j1-claude-desktop-1.14271.0/lib/claude-desktop/electron/chrome-sandbox
[PASS] Password store: kwallet6
[WARN] MCP config: exists but cannot validate (no python3 or node available)
[WARN] Node.js: not found (required for MCP servers)
       Fix: Install Node.js v20+ from https://nodejs.org
[WARN] Desktop entry not found (expected for AppImage installs)
[PASS] Disk space: 1704478MB free

Cowork Mode
----------------
[WARN] bubblewrap: not found
       Fix: Install bubblewrap using your package manager
[PASS] KVM: accessible
[PASS] vsock: module loaded
[PASS] QEMU: found
       socat: not found
       virtiofsd: not found
       VM image: not downloaded yet
       Cowork isolation: KVM (full VM isolation)
[PASS] Log file: 7KB (/home/fdl/.cache/claude-desktop-debian/launcher.log)

1 check(s) failed.
See above for fixes.
```

### What happened

On da341d7, the app window never shows up in NixOS x86-64. Instead, the command hangs indefinitely in the CLI -- not exiting or outputting any logs.

### Steps to reproduce

```console
$ nix run github:aaddrick/claude-desktop-debian/da341d733ce120463ea9f565751122ee844b3a00
```

### Expected behavior

The app window should eventually show up.

### Logs / errors

```shell
--- Claude Desktop Launcher Start (NixOS) ---
Timestamp: Fri Jun 19 03:18:53 PM +07 2026
Arguments: 
env={
  XDG_SESSION_TYPE=wayland
  WAYLAND_DISPLAY=wayland-0
  DISPLAY=:0
  XDG_CURRENT_DESKTOP=KDE
  GTK_IM_MODULE=
  XMODIFIERS=
  QT_IM_MODULE=
  CLAUDE_USE_WAYLAND=
  CLAUDE_TITLEBAR_STYLE=
  CLAUDE_PASSWORD_STORE=
  CLAUDE_GTK_IM_MODULE=
  CLAUDE_DISABLE_GPU=
}
Password store: kwallet6
Using X11 backend via XWayland (for global hotkey support)
App (auto-loaded by Electron): /nix/store/pq6fihpjl6k8j2x74wy025dl9407gb09-claude-desktop-1.13576.0/lib/claude-desktop/electron/resources/app.asar
Executing: /nix/store/pq6fihpjl6k8j2x74wy025dl9407gb09-claude-desktop-1.13576.0/lib/claude-desktop/electron/electron-wrapper --disable-features=CustomTitlebar --class=@@WM_CLASS@@ --password-store=kwallet6 --no-sandbox --ozone-platform=x11 
[Frame Fix] Wrapper loaded
[Frame Fix] Menu bar mode: auto
[Frame Fix] Titlebar style: hybrid
[Frame Fix] Close-to-tray: on
[Frame Fix] Keep awake: on (default)
[Autostart] XDG Autostart shim installed
[Frame Fix] Upgrade watcher armed: /nix/store/pq6fihpjl6k8j2x74wy025dl9407gb09-claude-desktop-1.13576.0/lib/claude-desktop/electron/resources/app.asar
[Frame Fix] Patches built successfully
(node:122117) [DEP0180] DeprecationWarning: fs.Stats constructor is deprecated.
(Use `electron --trace-deprecation ...` to show where the warning was created)
```

### Anything else

This may be related to the CI failures in the aforementioned commit or #678, but probably not related to #726 (although it shares similar symptoms) as deleting `~/.cache/claude-desktop-debian/launcher.log` does not fix the issue. That being said, the attached launcher log file is a fresh one (the old one was immediately removed prior).

[launcher.log](https://github.com/user-attachments/files/29129626/launcher.log)

chrisw1005:
## Problem

On Linux, Claude Desktop ≥ 1.13576.0 launches but the main window never appears — the process stays alive (idling in the libuv event loop) with no window, no renderer process, and nothing written to `~/.config/Claude/logs/main.log`. This is #729.

## Root cause

The upstream bundle (`.vite/build/index.pre.js` and `index.js`) calls Windows-only `@ant/claude-native` methods **unconditionally at module top-level** for the managed-config / enterprise-policy lookup, e.g.:

```js
const n = ((o = g2()) == null ? void 0 : o.readRegistryValues(r)) ?? [];
```

The guard only handles the native module being `null`, not the method being **absent**. Our Linux stub (`scripts/claude-native-stub.js`) doesn't implement `readRegistryValues`, so this throws `TypeError: readRegistryValues is not a function` during top-level execution of `index.pre.js` — before `require('./index.js')` and before the main window is ever created.

Because the throw is in `index.pre.js`'s own top-level (outside the `try/catch` it wraps the `index.js` require in), no `launch-failure.err` is written, which is why it presents as a silent hang rather than a crash.

Confirmed by injecting probes into a repacked `app.asar` run in a throwaway `--user-data-dir`:

```text
[P] entry top
[P] pre.js top
[P] uncaughtException: TypeError: o.readRegistryValues is not a function
        at ... index.pre.js:61
```

After guarding only that one call, startup advanced to the next identical throw — `s.readRegistryValues is not a function` in `index.js` — confirming several call sites of the same missing-method class.

## Fix

Add safe no-op stubs for the Windows-only native methods the bundle calls to `scripts/claude-native-stub.js`:

- `readRegistryValues: () => []`
- `writeRegistryValue: () => {}`
- `writeRegistryDword: () => {}`
- `getWindowsElevationType: () => "default"`
- `getCurrentPackageFamilyName: () => null`

On Linux there is no Windows registry / MSIX package / UAC elevation, so returning empty/neutral values is semantically correct, and the existing `?? []` / `?? "default"` consumers handle them. Fixing the stub (rather than each minified call site) covers every call site at the source and is robust against re-minification.

## Testing

Built the `.deb` from this branch and installed on Ubuntu (GNOME, Wayland + XWayland), Claude Desktop 1.14271.0:

- Main window appears and renders the real UI (`window_created_ms` and `main_view_dom_ready_ms` both logged).
- `main.log` shows **no** `"is not a function"` errors (previously `readRegistryValues` and `getWindowsElevationType`).
- Logged-in chat UI is fully functional.

Affected upstream versions observed: 1.13576.0 and 1.14271.0.

Fixes #729

---
85% AI / 15% Human
Claude: root-cause diagnosis via probe-injection experiment, the stub fix, build + install verification
Human: direction, environment/reproduction, verification approval

### DIFF
--- a/scripts/claude-native-stub.js
+++ b/scripts/claude-native-stub.js
@@ -42,6 +42,25 @@ class AuthRequest {
 
 module.exports = {
   getWindowsVersion: () => "10.0.0",
+
+  // Windows-only native methods that do not exist on Linux. Newer
+  // upstream (Claude Desktop >= 1.13576.0) calls readRegistryValues()
+  // UNCONDITIONALLY at startup (managed-config / enterprise-policy
+  // lookup) from index.pre.js and index.js top-level. The bundle only
+  // guards the module being null (`(o=g2())==null?void 0:o.readRegistryValues(r)`),
+  // not the method being absent, so a missing method throws
+  // "readRegistryValues is not a function" during top-level execution —
+  // before the logger and the main window are created. The app process
+  // stays alive but no window ever appears (idles in the event loop).
+  // Stub these as safe no-ops returning "nothing here" so the lookups
+  // succeed. Fixes the Linux "hangs indefinitely, app window never
+  // shows up" regression (upstream issue #729).
+  readRegistryValues: () => [],
+  writeRegistryValue: () => {},
+  writeRegistryDword: () => {},
+  getWindowsElevationType: () => "default",
+  getCurrentPackageFamilyName: () => null,
+
   setWindowEffect: () => {},
   removeWindowEffect: () => {},
 


xilec:
Confirming the same issue and that this fix resolves it.

**Environment:** NixOS (flake build via the repo's `claude-desktop-fhs` output), niri / Wayland, NVIDIA.

On `1.13576.0` the process started, the frame-fix ran, but the main process hung **before window creation** — no renderer, no logs in `~/.config/Claude/logs` at all. We had pinned back to `1.11847.5` as a workaround.

Built `1.14271.0` from this PR branch and ran it against an isolated profile: the window renders normally, logs are produced, the renderer reaches `claude.ai/login`, and no `claude-native ... is not a function` errors appear. So the root cause matches exactly and the stub fix works.

Thanks! 🙏

aaddrick:
Thanks @chrisw1005 — your root-cause work here was the correct one, and the probe writeup (catching that startup advances to a *second* throw on `getWindowsElevationType` after the first is guarded) is exactly why the complete five-method stub is the right call rather than patching a single site.

I've consolidated this into #737 against current `main` (now upstream 1.15200.0): it takes your full stub and adds bats coverage. You're credited via `Co-Authored-By`. I also audited every `@ant/claude-native` call site in the shipped 1.15200.0 asar to confirm the stub is complete for the current version — all three startup-reachable methods are covered.

I'll close this once #737 lands (CI running now). Appreciate the thorough diagnosis.

*— Written by Claude via [Claude Code](https://claude.ai/code)*

---

aaddrick:
Landed via #737, which merged your complete five-method stub plus bats coverage and is verified against the current upstream (1.15200.0). Closing as superseded — thanks again for the root-cause diagnosis, it was the basis for the fix. You're credited via `Co-Authored-By` and I'll add you to the README Acknowledgments.

*— Written by Claude via [Claude Code](https://claude.ai/code)*

---

PR #736, opened 2026-06-23
fix(patches): re-derive cowork Linux patches for the yukonSilver VM refactor (1.13576+)

pjordanandrsn:
## Summary

`main` CI has been red since the **1.13576.0** Claude Desktop bump (2026-06-17): the **Build Packages** jobs fail at the **"Verify cowork patches in shipped asar"** step with 9/11 cowork markers missing. This re-derives the cowork Linux patch suite against the refactored bundle (current `1.14271.0`).

## Root cause

Between Claude Desktop **1.12603.1** (last green) and **1.13576.0 → 1.14271.0**, upstream partially re-architected the cowork/VM subsystem ("yukonSilver"):

- The platform gate moved out of an inline `VAR!=="darwin"&&VAR!=="win32"` check (near `unsupported_platform`) and into `startVM`'s feature-flag check: `const{yukonSilver:r}=D_(); if((r==null?void 0:r.status)!=="supported"){...return}`.
- The vmClient module load moved behind the isMsix detector inside `YBt()` (`return Rl()?...{vm:...}...:null`).
- `sharedCwdPath`/`mountConda` were removed from the spawn payload; user-selected folders are now first-class (`userSelectedFolders` → `additionalMounts`).

**Patch 1** anchored on the now-gone `darwin`/`win32` check, so it `process.exit(1)`'d — which killed the single `node` patch block and dropped every subsequent cowork patch (the two `.asar` guards survived in separate blocks).

## Changes (`scripts/patches/cowork.sh` + `scripts/cowork-patch-markers.tsv`)

| Patch | Change |
|---|---|
| 1 platform gate | Re-anchored on `startVM`'s `yukonSilver.status` check (located via the unique `[startVM] VM not supported` log); Linux bypasses it. New marker `vm-supported-linux-gate`. |
| 2 vmClient load | Widened the isMsix gate in `YBt()` so the TS vmClient loads on Linux — collapses old 2a+2b into one site; gate fn captured dynamically (not hardcoded). New marker `vmclient-linux-gate`. |
| 6 step 2 auto-launch | Re-anchored on the new `await <helper>(<delay>)` retry call (was the inline `new Promise(r=>setTimeout(r,N))`). |
| 9 smol-bin copy | Idempotency guard fixed — it false-matched upstream's **own** new win32 `[VM:start] Copying smol-bin` log and would skip injecting the Linux block on a clean bundle; now keys on the injected `to bundle (Linux)` sentinel. Also gained a marker (`smol-bin-linux-copy`) — it was load-bearing but previously unverified. |
| 12 sharedCwdPath | Retired — `mountConda`/`sharedCwdPath` are gone; folder sharing now flows through first-class `userSelectedFolders` → the daemon's mountMap, so the threading is moot. Marker removed. |
| 6b / 8 | Anchors absent on this architecture; documented as safe no-ops (the Linux VM-download driver is unreachable — its `downloadVM` gate stays `yukonSilver`-gated, which Patch 1 deliberately leaves closed). |
| 3, 4, 4b, 6 step 1, 10 | Anchors survived; existing logic still matches — verified. |

Markers re-derived (−3 stale, +3 new); no cross-matches, samples self-consistent, comfortably above the bats `>= 10` floor.

## Verification

- **Local patch run against the real 1.14271.0 bundle**: 8 cowork patches apply, **11/11 markers present**, patched `index.js` passes `node --check`, idempotent on re-run.
- **Shipped artifact**: extracted the built `.deb`'s `app.asar` and confirmed 11/11 markers in the actual pipeline output, and that the **client↔daemon contract holds end-to-end** — the patched client socket path `(XDG_RUNTIME_DIR||"/tmp")+"/cowork-vm-service.sock"` matches the daemon's `SOCKET_PATH`, the autolaunch forks the `app.asar.unpacked/cowork-vm-service.js` that ships in the package, and both use the same 4-byte length-prefixed JSON `{method,params}`/`{success,result,error}` framing.
- **CI**: the same change is green on a downstream repackage (all 6 build variants + artifact validation + BATS/shellcheck/codespell). I traced this regression there first, then verified the fix applies cleanly to this repo's `main` (the patch files were byte-identical at the base).

## Caveat

CI builds the asar and verifies markers/contract but does not boot the app, so this hasn't been GUI-runtime-tested (launch + open a cowork session) on a Linux host — the static contract analysis indicates `startVM` is the sufficient main-process gate, but a click-through on Linux is the remaining confirmation.

---
~85% AI / ~15% Human
Claude: diagnosed the regression against the extracted 1.14271.0 bundle, re-derived the patches, verified locally + against the shipped artifact, and wrote the change.
Human: triaged, directed the approach, and authorized upstreaming.


aaddrick:
Hey! 

Thanks for the heads up. Will review this evening and ship.

pjordanandrsn:
Thanks! Let me know if anything needs adjusting.

aaddrick:
Verified this against the current `main` upstream (**1.15200.0**), not just the 1.14271.0 it was authored against — extracted the shipped `app.asar` and ran the re-derived patch suite end to end:

| Stage | Result |
|---|---|
| `patch_asar_path_filter` | ✅ found `F9A(A)` |
| `patch_asar_argv_file_drop_guard` | ✅ collector found (`param=n, fsVar=qA`) |
| `patch_cowork_linux` | ✅ 8 patches; vmClient gate fn `bl` captured dynamically |
| `--add-dir` filter (#649/#718) | ✅ filtered 2 dispatch loops (the #718 double-loop holds on 1.15200) |
| `node --check` patched bundle | ✅ parses |
| **All 11 markers** | ✅ **present** (`verify-patches` exit 0) |

```
OK vm-supported-linux-gate      OK getdownloadstatus-suppression  OK smol-bin-linux-copy
OK vmclient-linux-gate          OK econnrefused-on-linux          OK asar-adddir-filter
OK unix-socket-path             OK cowork-daemon-pid              OK asar-file-drop-guard
OK empty-linux-bundle-manifest  OK cowork-linux-daemon-shutdown
All 11 patch markers present.
```

The two patching WARNINGs (`reinstall file list array`, `mkdtemp("wvm-")`) are the documented safe no-ops — the Linux VM-download path they'd touch is gated off anyway, and neither affects a marker. No anchor drift between 1.14271 and 1.15200, so this needs no further work.

This is the gate that's currently keeping `main` (and #737) red. Plan: merge this first → `main` cowork gate goes green → then I'll update #737 (the #729 startup fix) from `main` so it inherits the fixed patches and goes fully green.

*— Written by Claude via [Claude Code](https://claude.ai/code)*

---

PR #740, opened 2026-06-25
fix(doctor): warn on non-symlink SingletonLock instead of false PASS

sabiut:
The `--doctor` SingletonLock check only handled the symlink case. A stale `hostname-PID` symlink correctly produced a warning, and a missing lock file correctly passed, but any other file type fell through to the default branch and was reported as:

`[PASS] SingletonLock: no lock file (OK)`

That isn't always true. A regular file can be left behind after an unclean update and still prevent Electron from acquiring its single-instance lock. In that situation, `--doctor` would incorrectly report everything as healthy, creating a false positive similar to the empty-password-store and octal-disk-read issues we fixed previously.

### Changes

* Refactored the SingletonLock logic in `scripts/doctor.sh` into a dedicated `_doctor_check_singleton_lock` helper, following the existing `_doctor_check_*` pattern and making it easier to test.
* Added an explicit `elif [[ -e $lock_file ]]` check to detect unexpected non-symlink files and emit a warning with an `rm` cleanup hint.
* Expanded `tests/doctor.bats` to cover all four scenarios:

  * Missing lock file → PASS
  * Live-PID symlink → PASS
  * Dead-PID symlink → WARN
  * Regular file → WARN

The regular-file case acts as a regression test to prevent this false-positive behaviour from returning. The dead-PID test now uses a known-dead PID generated during the test rather than relying on a hard-coded PID value.

### Validation

* `bats tests/doctor.bats` — 48/48 passing (up from 44)
* `bats tests/*.bats` — 319/319 passing
* `shellcheck -x scripts/doctor.sh` — clean

Also verified that the new regular-file test fails against the old implementation and passes with this change, confirming that it correctly captures the regression.


aaddrick:
Post-v3.0.0 triage: still wanted — the SingletonLock check is still inline in current [`scripts/doctor.sh`](https://github.com/aaddrick/claude-desktop-debian/blob/main/scripts/doctor.sh) (around line 1395, no `_doctor_check_singleton_lock` helper exists yet), so both the false-PASS fix and the helper extraction remain open targets. This looks like the one PR in your doctor-coverage campaign that didn't get the Jul 11 re-push that #744/#745 got — when you rebase it onto post-rebase main like those two, it should slot right in (and #782's userns extraction suggests the anchors won't collide).

If anything about the rebase makes you think this no longer applies, say so — otherwise looking forward to the re-push.

---

---

aaddrick:
Reviewed after the rebase ask, and the rebase is the easy part. I ran `git merge-tree --write-tree origin/main <head>` and it produces a clean tree, no conflicts, and it stays clean in every landing order against #744/#745/#782. I materialized the merged tree and ran the suite on it: 84/84 doctor.bats green, shellcheck clean. So a plain `git rebase origin/main` and force-push is genuinely all this needs. No code changes to land it.

The fix is real, and I read the sources to confirm it. Chromium's `process_singleton_posix.cc` gets an empty result from `ReadLink` on a regular file, `ParseProcessSingletonLock` fails, and the `symlink()` call then hits `EEXIST`. The only non-symlink recovery path (`ReplaceOldSingletonLock`) is compiled on macOS only, Electron's vendored patch doesn't touch it, and the app quits when `requestSingleInstanceLock()` returns false. So a regular file at `SingletonLock` hard-blocks the next cold launch on Linux. The false PASS this kills is a real one. Nice property too: graceful shutdown unlinks the lock, so the WARN fires exactly in the surviving bad state.

Two things to fix up at rebase time. The helper's header comment attributes "blocks startup until removed" to the stale dead-PID symlink case, but Chromium self-heals that one (orphaned symlink, unlink, continue). It's the regular-file case that actually blocks, so that's a one-line comment fix. The other is a severity call worth making explicit. The regular-file case provably blocks startup, which is the class doctor rates `_fail` elsewhere (missing Electron binary, sandbox perms), while the self-healing symlink case gets the same WARN. So severity is inverted relative to impact. I lean toward keeping WARN for consistency with the existing stale-symlink precedent — a one-line comment saying why will stop the next person from relitigating it.

One follow-up, not for this PR: the live-PID branch has the mirror-image false PASS. A stale lock whose PID got reused by any live same-user process reports "held by running process." Chromium guards this with an exe-basename check (`IsChromeProcess`), and the helper could do the same via `readlink /proc/$pid/exe`. Opened #784 so it doesn't get lost. The directory-vs-`rm`-hint and `|| true` nits are cosmetic and pre-existing, so I'd leave them.

---

sabiut:
Rebased and re-pushed as f4c7e61 — now a single commit on current main (the old branch base predated the v3.0.0 restructure, so this is a clean replant, not a replay of history). Both comment fixes are folded in:

- **Header comment corrected** — the stale dead-PID symlink is now described as self-healed (Chromium unlinks the orphan and continues), and the regular-file case carries the actual blocking mechanics (empty ReadLink → failed lock parse → `symlink()` EEXIST → app quits on next cold launch).
- **Severity call recorded** — the elif keeps WARN with an inline comment saying it's for consistency with the stale-symlink precedent even though the regular-file case provably blocks, so the next reader doesn't relitigate it.

Suite at the new head: 89/89 green (85 on main + the 4 singleton-lock states), shellcheck clean. Good catch on #784 for the PID-reuse mirror image — the `readlink /proc/$pid/exe` basename check is a natural follow-up once this lands.

Like the others, self-merge waits on #792 (AU-1 tripwire re-point) making CI green again.

---

PR #744, opened 2026-06-25
refactor(doctor): extract the display-server check into a testable helper

sabiut:
Extracts run_doctor's inline display-server check into `_doctor_check_display_server` with 5 bats tests (Wayland/X11/none triage, native-vs-XWayland mode notes). Pure behavior-preserving refactor — doctor output verified byte-identical.

**2026-07-11:** re-scoped and force-pushed onto post-v3.0.0 main. The menu-bar, titlebar, and keep-awake extractions from the original PR are gone — the rebase deleted those checks and their env vars are now covered by the already-tested `_check_legacy_env` — leaving display-server as the one surviving target.


aaddrick:
This is merge-ready. I diffed the extracted helper against main's inline block at scripts/doctor.sh:1243-1260 (e7673c9) and it's a byte-identical move. All the `${VAR:-}` guards are intact and the section ordering of the whole --doctor report is unchanged. I ran the full doctor.bats suite at the PR head: 85/85 green. I also mutation-checked the five new tests. Swapping the Wayland/X11 precedence, inverting the CLAUDE_USE_WAYLAND tri-state, and breaking the else-branch each fail at least one test, so they're pinning real behavior and not just passing on green.

One thing before merge. The neither-display FAIL test doesn't assert that `_doctor_failures` got incremented, and that increment is the only thing from this block feeding doctor's exit status (`return "$_doctor_failures"`). This is the same run-subshell-discards-the-mutation gap we fixed in #774. The redirect-to-file pattern for asserting it already exists at tests/doctor.bats:1128-1140, so it's a small addition.

Two optional cases, take them or leave them. A `CLAUDE_USE_WAYLAND='0'` test would pin the `== '1'` comparison against a set-but-not-1 value, which is the tri-state's force-XWayland case. Right now only unset vs '1' are covered. And `XDG_CURRENT_DESKTOP` isn't in setup()'s unset list. Harmless today since tests 2 and 4 don't assert the Desktop line, but it goes host-dependent the moment someone does.

Sequencing and CHANGELOG notes for the three same-anchor extractions are over on #782 to keep them in one place.

---

sabiut:
All three landed in f80644c, mutation-verified:

- **Failure-counter assert** — the neither-display FAIL test now calls the helper directly with output redirected to a file (same pattern as the `_cowork_incomplete` tests at tests/doctor.bats:1128-1140) and asserts `_doctor_failures` goes 0→1. Mutation check: replacing the `_fail` call with a bare `echo "[FAIL] ..."` (prints, doesn't increment) fails exactly this test and nothing else.
- **Tri-state `'0'` test** — pins `== '1'` against the set-but-not-1 force-XWayland value, asserting on the `Mode:` prefix (the else-branch's tip line contains the literal "native Wayland", so a bare substring exclusion self-collides). Mutation check: loosening the comparison to `-n "${CLAUDE_USE_WAYLAND:-}"` fails exactly this test.
- **`XDG_CURRENT_DESKTOP`** added to setup()'s unset list.

Suite at the new head: 86/86 green. Shellcheck delta is zero (same SC2034 count before/after; the new assignments follow the file's existing consumed-by-sourced-helper pattern).

Heads-up on the red CI: the build legs fail on every branch right now because upstream 1.19367.0 dropped the `apt_channel_pending` marker and the AU-1 tripwire exits 1 by design — main's own bump commit fails the same way. Investigating the new bundle's updater gate separately; not caused by (or fixable from) this PR.

---

PR #745, opened 2026-06-25
refactor(doctor): extract Electron + chrome-sandbox checks into helpers

sabiut:
Extracts run_doctor's inline Electron-binary and chrome-sandbox checks into `_doctor_check_electron_binary` / `_doctor_check_chrome_sandbox`, plus a `_DOCTOR_DEB_SANDBOX` test hook (default = real install path), and 10 bats tests. Pure behavior-preserving refactor — doctor output verified byte-identical.

**2026-07-11:** force-pushed as a re-application onto post-v3.0.0 main (the original June branch predated the rebase); current paths and hint strings preserved, stat stub rewritten without `eval`, tests re-anchored so this no longer conflicts with #744.


aaddrick:
This is merge-ready in substance, with two test asks below. I diffed the electron-binary helper against main's inline block at scripts/doctor.sh:1271-1289 and it's mechanically byte-identical, `${ver:+v${ver#v} }(system)` quirk and all. The sandbox helper differs only by the `_DOCTOR_DEB_SANDBOX` hook, which defaults to the same literal, plus it closes the one variable leak. `for sandbox_path` was the only non-local in that block, and nothing reads it downstream. I checked the `${!var}` indirection sites too. Call site, section headers, and check order are untouched. Full suite at PR head is 90/90 green, and the mutations reproduce: '4755'→'0755' breaks exactly 3 tests, and dropping the deb path from the probe array breaks the deb-path test.

Two test asks before I merge it. First, the stat stub keys on `$2 == '%a'`, so a typo in the real invocation that keeps `$2` still passes all 90. For example `stat -c` → `stat -f`, which on GNU turns %a into free-block-count garbage. The FAIL branch doesn't need the stub at all. A real 0644 file gives real `stat` output (`644`/`$USER`), so it exercises the actual flags and the parse. Converting one FAIL-branch test to real-stat covers that honesty gap while keeping the stub for the un-fakeable 4755+root PASS case. Second, nothing pins probe precedence. Deleting the `break` entirely passes all tests, and in production that double-reports when both the deb path and an electron-adjacent sandbox exist. One test with both files present, asserting the deb path wins and only one line prints, closes it.

One shared blind spot, filed as a follow-up rather than fixed here. The "Chrome sandbox not found (expected for AppImage)" WARN is structurally unreachable on AppImage. AppRun stages the full official tree including chrome-sandbox beside the ELF (appimage.sh:30-32), so the adjacent probe always finds it inside the read-only squashfs. Static analysis says the check then most plausibly FAILs with a `sudo chown` hint against a transient mount, which is unactionable, and moot anyway since the AppImage runs --no-sandbox. The inline original has the identical flaw, so your PR is behavior-preserving right down to the bug. But the new docstring repeats the wrong "expected for AppImage" claim, so I opened #785 to skip or soften the check when a mount path is detected. Naming nit while we're here: rpm installs to the same /usr/lib/claude-desktop-unofficial/ path, so `_DOCTOR_DEB_SANDBOX` and "deb path" are a misnomer. Fine to leave.

Last thing: the PR body's "no longer conflicts with #744" holds only for the bats side — the doctor.sh helper insertions still collide. Sequencing and CHANGELOG notes for the three same-anchor extractions are over on #782 to keep them in one place.

---

sabiut:
Both asks landed in f6f9d75, mutation-verified: `stat -c` → `stat -f` now fails exactly the converted real-stat test, and deleting the `break` fails exactly the new precedence test (both files present → deb path judged, one line printed). doctor.bats 90 → 91.

You're right about the doctor.sh side of the anchor collision — only the bats insertions were de-conflicted; the helper blocks in #744/#745/#782 all insert before `run_doctor`. Whichever lands after the first two needs a trivial re-anchor; I'll handle each rebase same-day as merges happen, any order works.

Agreed #785 is the right home for the unreachable-AppImage WARN — the docstring inherited the claim from the inline original, and I'd rather keep this PR byte-preserving, bug included. Naming nit likewise left per your call.

One heads-up for after this merges: the precedence test pins the *current* deb-first semantics on purpose. My follow-up chrome-sandbox fix (branch ready, stacked on this) intentionally flips that expectation — when an electron path is given it judges only that binary's sandbox — so expect that test's assertion to invert in the follow-up's diff rather than read it as a regression.

sabiut:
The red CI on this PR (and every branch since the 1.19367.0 pin) is the AU-1 tripwire, not this change: upstream renamed the Linux updater's disable reason from `apt_channel_pending` to `managed_by_package_manager`, so the anchor grep fails the build by design. Investigated the new bundle — the updater gate itself is unchanged and still inert on Linux, so D-001 is unaffected; the tripwire just needed re-pointing. Fix with full evidence: #792. CI here should go green after that merges and this branch re-runs.

---

sabiut:
Follow-up promised above is now open as #816 — the electron-wins behavior change, rebased onto current main.

PR #751, opened 2026-06-26
fix(tray): re-derive in-place fast-path for the 1.13576+ rebuild refactor

aaddrick:
The "yukonSilver"-era refactor (1.13576+, verified against 1.15962.0)
restructured the tray rebuild block in two ways that silently broke
patch_tray_inplace_update:

  - The destroy/recreate guard became
    `if(X=[],Y=!1,TRAY&&(TRAY.destroy()...),!enabled)` instead of the
    old `;if(TRAY&&(TRAY.destroy()`, so the injection anchor missed.
  - The context menu is now a prebuilt object
    (`M=BUILDER();TRAY.setContextMenu(M)`), and the *first* setContextMenu
    call site is a `setContextMenu(null)` menu-clear. The resolver's
    `head -1` latched that `null`, so the builder never resolved, the
    #680 WARNING fired, and the fast-path was skipped — re-arming the
    #515 KDE/Plasma duplicate-icon StatusNotifier race. Tray patches
    warn-and-continue, so this shipped through green CI.

Fixes:
  - Menu-function resolver walks every setContextMenu argument, skips the
    `null` clear, and takes the first that resolves to a VAR=BUILDER()
    assignment.
  - Injection switched from a backslash-heavy regex anchor to locating
    TRAY.destroy() and walking back to the `;if(` that opens its
    statement, landing correctly in both the old and new shapes.
  - patch_menu_bar_default now recognizes that upstream defaults
    menuBarEnabled to true via its defaults map (`menuBarEnabled:!0`), so
    the legacy !!-rewrite is a no-op by design rather than a bare miss;
    it warns loudly only if neither shape is present.

Adds tests/tray-patches.bats (7 cases): builder resolution past the null
decoy, injection placement before the destroy block + valid JS,
idempotency, loud-fail on a missing destroy site, and the three
menu-bar-default branches. Full suite 329/329.

Verified end-to-end against the real 1.15962.0 app.asar. Runtime
confirmation of the #515 race avoidance still needs a KDE Plasma host.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ The format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) — 
 
 <!-- Updated automatically by check-claude-version; will be current at release time. -->
 
+### Fixed
+
+- The tray in-place fast-path applies again on Claude Desktop 1.13576+ (verified against 1.15962.0). The "yukonSilver"-era refactor restructured the tray rebuild block — the destroy/recreate is now guarded by `if(X=[],Y=!1,TRAY&&(TRAY.destroy()…),!enabled)` instead of `;if(TRAY&&(TRAY.destroy()`, and the context menu became a prebuilt object (`M=BUILDER();TRAY.setContextMenu(M)`) with a leading `setContextMenu(null)` menu-clear. The patch's `head -1` latched that `null` clear, so the menu builder never resolved, the #680 WARNING fired, and the fast-path was skipped — silently re-arming the #515 KDE/Plasma duplicate-icon StatusNotifier race because tray patches warn-and-continue. The menu-function resolver now walks every `setContextMenu` argument, skips the `null` clear, and resolves the real builder; the injection anchors on the `;if(` that opens the destroy statement so it lands in both the old and new shapes. New `tests/tray-patches.bats` covers builder resolution, injection placement, idempotency, the loud-fail-on-missing-anchor path, and the menu-bar-default cases. ([#746](https://github.com/aaddrick/claude-desktop-debian/issues/746) investigation)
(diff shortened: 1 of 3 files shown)


aaddrick:
🙏 **Requesting a KDE Plasma runtime check before merge.**

The static layers are all green (bats 330/330, `node --check` on the patched bundle, re-derived + verified end-to-end against the real 1.15962.0 `app.asar`), but the in-place fast-path exists specifically to dodge the Plasma StatusNotifier re-registration race from #515 — and only a real Plasma session can confirm that timing. Static checks are blind to it by design (see `docs/learnings/patching-minified-js.md` → "only a runtime launch verifies behavior").

@IliyaBrook — you wrote the original fast-path in #515 and have the exact repro; @kotok9 — you've been doing instrumented KDE Plasma 6.6 testing on #716. Would either of you be up for a quick check?

**No need to build from source** — CI on this PR produces an AppImage. Grab it from this PR's CI run → the **`claude-desktop-amd64-appimage`** artifact.

**Repro steps (KDE Plasma):**
1. Run the PR artifact AppImage.
2. Confirm the tray icon renders and the right-click menu (Show App / Quit) works on 1.15962.0.
3. Change the Plasma theme to flip light↔dark — **Appearance → Colors** (or Plasma Style / Global Theme). All three fire the same `nativeTheme` "updated" event.
4. Watch the tray icon: it should **swap in place** (light/dark template) with **no brief duplicate icon** appearing next to it. A duplicate flash = the fast-path isn't engaging.

For contrast, a build from `main` (where the fast-path silently no-ops on 1.13576+) should show the duplicate-icon flash on theme change — so if you have a moment, that's a useful before/after.

Thanks either way — much appreciated. 🐧

---
Written by Claude via [Claude Code](https://claude.ai/code)

---

Kuchiru:
Had the same issue on Gnome as well (double tray icon), this fixed it.

Issue #588, opened 2026-05-06
[bug]: icon of claude in taskbar in XFCE isn't showing properly


sagisar1:
### Version (`claude-desktop --doctor` output)

```shell
Claude Desktop Diagnostics
================================

[PASS] Installed version: 1.6259.0-2.0.10
[PASS] Display server: X11 (DISPLAY=:0.0)
       Menu bar mode: auto (default, Alt toggles visibility)
       Titlebar style: hybrid (default, native frame + in-app topbar)
[PASS] Electron: v41.5.0 (/usr/lib/claude-desktop/node_modules/electron/dist/electron)
[PASS] Chrome sandbox: permissions OK (/usr/lib/claude-desktop/node_modules/electron/dist/chrome-sandbox)
[PASS] SingletonLock: held by running process (PID 3907)
[PASS] MCP config: valid JSON (/home/kali/.config/Claude/claude_desktop_config.json)
       MCP servers configured: 0
[PASS] Node.js: v22.22.2
       Path: /usr/bin/node
[PASS] Desktop entry: /usr/share/applications/claude-desktop.desktop
[PASS] Disk space: 40082MB free

Cowork Mode
----------------
[PASS] bubblewrap: found
[PASS] bubblewrap: sandbox probe succeeded
       KVM: not available
[PASS] vsock: module loaded
       QEMU: not found
[PASS] socat: found
       virtiofsd: not found
       VM image: not downloaded yet
       Cowork isolation: bubblewrap (namespace sandbox)
[PASS] Cowork daemon: running (parent alive)
[PASS] Log file: 339KB (/home/kali/.cache/claude-desktop-debian/launcher.log)
```

### What happened

<img width="1917" height="42" alt="Image" src="https://github.com/user-attachments/assets/87af0b77-f16b-405c-bb3b-c84c6f3502cf" />
as you can see in the image, the icon i circled in red isn't showing claude icon, and also the black dot between the speaker and network settings in the right side of the taskbar (which rpresents the quick window screen where you can ask claude something) isn't showing properly  

### Steps to reproduce

1. Install Kali linux with xfce or probably any  linux distribution with xfce DE
2. Install latest version of claude from this repository
3. Launch claude-desktop either by GUI or by terminal
4. Watch the incorrect icon rendering

### Expected behavior

Should show claude Icon 

### Logs / errors

```shell

```

### Anything else

if you need anything else, tell me

aaddrick:
[...]

Issue #557, opened 2026-05-02
[bug]: sys tray icon displays a cryptic transparent window full of source code (Linux Mint Cinnamon X11)


brainexcerpts:
### Version (`claude-desktop --doctor` output)

```shell
Claude Desktop Diagnostics
================================

[PASS] Installed version: 1.5354.0-2.0.7
[PASS] Display server: X11 (DISPLAY=:0)
       Menu bar mode: auto (default, Alt toggles visibility)
       Titlebar style: hybrid (default, native frame + in-app topbar)
[PASS] Electron: v41.4.0 (/usr/lib/claude-desktop/node_modules/electron/dist/electron)
[PASS] Chrome sandbox: permissions OK (/usr/lib/claude-desktop/node_modules/electron/dist/chrome-sandbox)
[PASS] SingletonLock: held by running process (PID 4955)
[PASS] MCP config: valid JSON (/home/arkan/.config/Claude/claude_desktop_config.json)
       MCP servers configured: 0
[PASS] Node.js: v20.20.2
       Path: /usr/bin/node
[PASS] Desktop entry: /usr/share/applications/claude-desktop.desktop
[PASS] Disk space: 212861MB free

Cowork Mode
----------------
[PASS] bubblewrap: found
[PASS] bubblewrap: sandbox probe succeeded
[PASS] KVM: accessible
[PASS] vsock: module loaded
       QEMU: not found
[PASS] socat: found
       virtiofsd: not found
       VM image: not downloaded yet
       Cowork isolation: bubblewrap (namespace sandbox)
[PASS] Cowork daemon: running (parent alive)
[PASS] Log file: 46KB (/home/arkan/.cache/claude-desktop-debian/launcher.log)

All checks passed.
```

### What happened

Click on sys tray and a cryptic window appears, background is transparent, it's full of source code:

<img width="611" height="502" alt="Image" src="https://github.com/user-attachments/assets/c99be2a9-80b2-45ca-85c1-3f103fba51a7" />

<img width="655" height="523" alt="Image" src="https://github.com/user-attachments/assets/9adcfd14-7a4e-49e2-bd5f-4e6004963366" />

Source code inside that window (copy / paste from it) most of it likely from node-pty npm package (v1.1.0) by Microsoft 

[code_source.txt](https://github.com/user-attachments/files/27296852/code_source.txt)

### Steps to reproduce

Click on claude desktop sys tray icon 

[...]

PR #752, opened 2026-06-26
fix(tray): route icon format to png on Linux — fixes black-square tray icon (#746)

### LINKED ISSUE #746: [bug]: Menu bar icon is broken after update to v. 1.15200.0

```shell
Claude Desktop Diagnostics

[PASS] Installed version: 1.15200.0-2.0.22
[PASS] Display server: X11 (DISPLAY=:0)
[PASS] MCP config: valid JSON (/home/nassoskranidiotis/.config/Claude/claude_desktop_config.json)
       MCP servers configured: 2
[PASS] Node.js: v24.11.1
       Path: /home/nassoskranidiotis/.nvm/versions/node/v24.11.1/bin/node
[PASS] Disk space: 515429MB free

Cowork Mode
----------------
[PASS] bubblewrap: found
[PASS] KVM: accessible
[PASS] vsock: module loaded
[PASS] QEMU: found
[PASS] socat: found
[PASS] virtiofsd: found at /usr/libexec/virtiofsd (not on PATH)
       VM image: not downloaded yet
[PASS] Log file: 1042KB (/home/nassoskranidiotis/.cache/claude-desktop-debian/launcher.log)

All checks passed.
```

### What happened

OS: Linux Mint 22.3 - Cinnamon 64-bit
Menu bar icon is broken after update to v. 1.15200.0. The menu bar icon is not the Claude icon, but an empty icon.  Right clicking the menu bar icon shows a menu with two options: show app, and quit. Neither of them is working. Clicking the icon won't open the Claude prompt bar.  Uninstalling and re-installing the app won't solve the issue. 

### Steps to reproduce

Uninstalling and re-installing the app won't solve the issue. 

### Expected behavior

The menu bar icon should be responsive.

### Logs / errors

```shell

```

### Anything else

_No response_

aaddrick:
Fixes #746 (and the on-device black-square repro on Nobara KDE).

## Problem

The system tray / menu-bar icon renders as a **black square** (or invisible/empty) on Linux since the upstream 1.13576+ refactor.

Upstream replaced the runtime platform check for the tray icon *format* with a build-time constant baked for the **Windows** build (`uPi="ico"`, part of a `gPi=!0,lPi=36,uPi="ico",IPi=!0,EPi="registry"` constant block), routed through:

```js
switch(uPi){                                          // uPi === "ico" — never reassigned by platform
  case"ico":            e=dark?"Tray-Win32-Dark.ico":"Tray-Win32.ico"; break;   // ← Linux lands here
  case"template-image": e="TrayIconTemplate.png"; break;
  case"png":            e=dark?"TrayIconTemplate-Dark.png":"TrayIconTemplate.png"; break;
}
```

Because the project repackages the Windows asar verbatim, that constant stays `"ico"` on Linux, so the tray loads the Windows multi-resolution `Tray-Win32.ico`. Electron's `nativeImage.createFromPath` on a Windows `.ico` renders as a **black square** on the freedesktop/KDE StatusNotifier.

## Why it slipped through

`patch_tray_icon_selection` anchored on `:VAR="TrayIconTemplate.png"`, which now matches the **`template-image` case** — a **dead branch** on Linux (the live case is `"ico"`). So the patch logged green while the icon stayed broken. Same silent-no-op class as the other 1.13576+ tray regressions (the build's tray patches warn-and-continue).

## Fix

Route Linux to the already-theme-aware `"png"` case by wrapping the switch discriminant:

```js
switch(process.platform==="linux"?"png":uPi){case"ico":…
```

The `"png"` case already does `dark?"TrayIconTemplate-Dark.png":"TrayIconTemplate.png"`, and the build already ships + processes those PNG templates for Linux visibility (`scripts/staging/icons.sh` binarizes their alpha). No icon-name re-derivation needed.

Per `docs/learnings/patching-minified-js.md`:
- **Literal anchor** — `switch(VAR){case"ico":`; the `case` labels are developer strings that survive minification, the discriminant (`uPi`) is extracted dynamically, never hardcoded.
- **Uniqueness asserted** — bails with a loud warning if the switch count isn't exactly 1.
- **Idempotent** — the wrapped discriminant is its own marker.
- **Non-fatal** — a miss warns (icon may be wrong) but doesn't abort the build.

## Tests

New `tests/tray-icon-format.bats` (5 cases): discriminant routing, theme-aware `png` case preserved, idempotency, warn-and-skip on a missing switch, and warn on an ambiguous (>1) switch. Full suite green.

## Verification

Verified end-to-end against the real **1.15962.0** `app.asar` (SHA-checked against the `detect-host.sh` pin): the patch wraps exactly one switch, output passes `node --check`, and Linux now selects `TrayIconTemplate.png`/`-Dark.png`.

**Runtime confirmation on a black-square host is still welcome** — the static layers can't see the rendered StatusNotifier.

## Relationship to #751

Independent of #751 (which fixes the duplicate-icon *race*). This one is the actual #746 black-square fix, and is a prerequisite for meaningfully testing #751 — you can't observe an in-place icon swap when the icon is a black square. Both touch `tray.sh` but different functions; they should merge cleanly.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q54GPhun4M91hja2T9Gw8G)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ The format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) — 
 
 <!-- Updated automatically by check-claude-version; will be current at release time. -->
 
+### Fixed
+
+- The system tray / menu-bar icon no longer shows as a black square (or an invisible/empty icon) on Linux. Upstream 1.13576+ replaced the runtime platform check for the tray icon *format* with a build-time constant baked for Windows (`uPi="ico"`) routed through `switch(uPi){case"ico":…Tray-Win32.ico…;case"png":…TrayIconTemplate…}`. Because the project repackages the Windows asar verbatim, Linux kept loading the Windows multi-resolution `.ico`, which Electron's `nativeImage` renders as a black square on the freedesktop/KDE StatusNotifier. The existing icon patch was silently rewriting the now-dead `template-image`/`png` case while the live `"ico"` case stayed broken. `patch_tray_icon_selection` now wraps the switch discriminant (`switch(process.platform==="linux"?"png":uPi)`) so Linux takes the already-theme-aware `"png"` case, which selects the `TrayIconTemplate[-Dark].png` assets the build ships and processes for visibility. ([#746](https://github.com/aaddrick/claude-desktop-debian/issues/746))
+
 ## [v2.0.22] — 2026-06-25
 
 Tracks upstream Claude Desktop 1.15200.0.
--- a/scripts/patches/tray.sh
+++ b/scripts/patches/tray.sh
@@ -71,18 +71,59 @@ patch_tray_menu_handler() {
 }
 
 patch_tray_icon_selection() {
-	echo 'Patching tray icon selection for Linux visibility...'
+	echo 'Patching tray icon format selection for Linux visibility...'
 	local index_js='app.asar.contents/.vite/build/index.js'
-	local dark_check="${electron_var_re}.nativeTheme.shouldUseDarkColors"
 
-	if grep -qP ':[$\w]+="TrayIconTemplate\.png"' "$index_js"; then
-		sed -i -E \
-			"s/:([[:alnum:]_\$]+)=\"TrayIconTemplate\.png\"/:\1=${dark_check}?\"TrayIconTemplate-Dark.png\":\"TrayIconTemplate.png\"/g" \
-			"$index_js"
-		echo 'Patched tray icon selection for Linux theme support'
-	else
-		echo 'Tray icon selection pattern not found or already patched'
+	# Upstream 1.13576+ bakes the tray icon *format* as a build-time
+	# constant set for the Windows build (`uPi="ico"`) and routes it
+	# through a switch:
+	#   switch(uPi){
+	#     case"ico": e=dark?"Tray-Win32-Dark.ico":"Tray-Win32.ico"; break;
+	#     case"template-image": e="TrayIconTemplate.png"; break;
+	#     case"png": e=dark?"TrayIconTemplate-Dark.png":"TrayIconTemplate.png";
+	#   }
+	# Because we repackage the Windows asar verbatim, that constant stays
+	# "ico" on Linux, so the tray loads the Windows multi-resolution
+	# `.ico`, which Electron's nativeImage renders as a black square on the
+	# freedesktop/KDE StatusNotifier (issue #746). The "png" case already
+	# selects the theme-aware PNG templates the build ships and processes
+	# for Linux visibility (TrayIconTemplate[-Dark].png, made opaque in
+	# scripts/staging/icons.sh) — so the fix is to route Linux to that
+	# case rather than re-deriving the icon names.
+	#
+	# Anchor on the stable `switch(VAR){case"ico":` shape: the case labels
+	# are developer strings that survive minification; VAR (`uPi`) is
+	# extracted dynamically, never hardcoded. Non-fatal — on a miss the
+	# tray may show the wrong icon but the app still runs — so warn loudly
+	# instead of exiting, matching the rest of the tray patch set.
+
+	# Idempotency: our wrapped discriminant is its own marker.
+	if grep -qF 'switch(process.platform==="linux"?"png":' "$index_js"; then
+		echo '  Tray icon format already routed to png on Linux'
+		echo '##############################################################'
+		return
 	fi
+
+	local sw_count sw_var sw_var_re
+	sw_count=$(grep -coP 'switch\([$\w]+\)\{case"ico":' "$index_js")
+	if [[ $sw_count -ne 1 ]]; then
+		echo "WARNING: expected exactly 1 tray icon-format switch" \
+			"(switch(VAR){case\"ico\":), found ${sw_count} — the Linux" \
+			"tray icon may render as a black square (#746); the" \
+			"icon-format switch shape likely changed" >&2
+		echo '##############################################################'
+		return
+	fi
+	sw_var=$(grep -oP 'switch\(\K[$\w]+(?=\)\{case"ico":)' "$index_js" \
+		| head -1)
+	echo "  Found icon-format switch var: $sw_var"
+
+	# Escape `$` for the sed pattern (minified names can be e.g. `u$i`).
+	sw_var_re="${sw_var//\$/\\$}"
+	sed -i -E \
+		"s/switch\(${sw_var_re}\)\{case\"ico\":/switch(process.platform===\"linux\"?\"png\":${sw_var}){case\"ico\":/" \
+		"$index_js"
+	echo '  Routed tray icon format to png on Linux'
 	echo '##############################################################'
 }
 
--- a/tests/tray-icon-format.bats
+++ b/tests/tray-icon-format.bats
@@ -1,0 +1,111 @@
+#!/usr/bin/env bats
+#
+# tray-icon-format.bats
+# Tests for patch_tray_icon_selection in scripts/patches/tray.sh — the
+# fix that routes the tray icon *format* to the PNG case on Linux.
+#
+# Regression guard for #746: upstream 1.13576+ bakes the icon format as a
+# build-time constant (uPi="ico") and routes it through
+#   switch(uPi){case"ico":…Tray-Win32.ico…;case"png":…TrayIconTemplate…}
+# Since we repackage the Windows asar, that constant stays "ico" on Linux,
+# so the tray loads the Windows .ico and renders as a black square on the
+# freedesktop/KDE StatusNotifier. The patch wraps the switch discriminant
+# so Linux takes the already-theme-aware "png" case.
+
+SCRIPT_DIR="$(cd "$(dirname "${BATS_TEST_FILENAME}")" && pwd)"
+TRAY_SH="$SCRIPT_DIR/../scripts/patches/tray.sh"
+
+setup() {
+	TEST_TMP=$(mktemp -d)
+	export TEST_TMP
+
+	index_js_dir="$TEST_TMP/app.asar.contents/.vite/build"
+	mkdir -p "$index_js_dir"
+	index_js="$index_js_dir/index.js"
+
+	project_root="$TEST_TMP"
+	electron_var='aA'
+	electron_var_re='aA'
+	export project_root electron_var electron_var_re
+
+	# shellcheck source=../scripts/patches/tray.sh
+	source "$TRAY_SH"
+}
+
+teardown() {
+	if [[ -n "${TEST_TMP:-}" && -d "$TEST_TMP" ]]; then
+		rm -rf "$TEST_TMP"
+	fi
+}
+
+# Minified fixture mirroring the 1.13576+ icon-format switch: a baked
+# uPi="ico" constant, the three-case switch (Windows .ico / template /
+# theme-aware png), and the Tray creation that consumes the result.
+write_icon_switch_fixture() {
+	cat > "$index_js" <<'JS'
+const aA=require("electron");
+const gPi=!0,lPi=36,uPi="ico",IPi=!0;
+function rebuild(){
+let e;switch(uPi){case"ico":e=aA.nativeTheme.shouldUseDarkColors?"Tray-Win32-Dark.ico":"Tray-Win32.ico";break;case"template-image":e="TrayIconTemplate.png";break;case"png":e=aA.nativeTheme.shouldUseDarkColors?"TrayIconTemplate-Dark.png":"TrayIconTemplate.png";break}
+const t=X.join(toi(),e);
+vE=new aA.Tray(aA.nativeImage.createFromPath(t));
+}
+JS
+}
+
+@test "icon-format: routes the switch discriminant to png on Linux" {
+	write_icon_switch_fixture
+	cd "$TEST_TMP"
+	run patch_tray_icon_selection
+	[[ "$status" -eq 0 ]]
+	echo "$output" | grep -q 'Found icon-format switch var: uPi'
+	grep -qF 'switch(process.platform==="linux"?"png":uPi){case"ico":' \
+		"$index_js"
+	node --check "$index_js"
+}
+
+@test "icon-format: the routed-to png case still selects the theme-aware PNG" {
+	write_icon_switch_fixture
+	cd "$TEST_TMP"
+	patch_tray_icon_selection
+	# The case Linux now lands on must keep the dark/light PNG template
+	# selection — i.e. the fix is inert if the png case lost it.
+	grep -qF 'case"png":e=aA.nativeTheme.shouldUseDarkColors?"TrayIconTemplate-Dark.png":"TrayIconTemplate.png"' \
+		"$index_js"
+}
+
+@test "icon-format: idempotent — second run is a no-op" {
+	write_icon_switch_fixture
+	cd "$TEST_TMP"
+	patch_tray_icon_selection
+	run patch_tray_icon_selection
+	[[ "$status" -eq 0 ]]
+	echo "$output" | grep -q 'already routed to png on Linux'
+	[[ "$(grep -coF 'process.platform==="linux"?"png":' "$index_js")" -eq 1 ]]
+	node --check "$index_js"
+}
+
+@test "icon-format: warns and skips when the switch shape is absent" {
+	cat > "$index_js" <<'JS'
+const aA=require("electron");
+function rebuild(){const e="TrayIconTemplate.png";vE=new aA.Tray(aA.nativeImage.createFromPath(e));}
+JS
+	cd "$TEST_TMP"
+	run patch_tray_icon_selection
+	[[ "$status" -eq 0 ]]
+	echo "$output" | grep -qi 'WARNING'
+	# Must not have injected a bogus discriminant.
+	! grep -qF 'process.platform==="linux"?"png":' "$index_js"
+}
+
+@test "icon-format: warns when more than one icon-format switch exists" {
+	write_icon_switch_fixture
+	cat >> "$index_js" <<'JS'
+function other(){switch(zZ){case"ico":e=1;break}}
+JS
+	cd "$TEST_TMP"
+	run patch_tray_icon_selection
+	[[ "$status" -eq 0 ]]
+	echo "$output" | grep -qiE 'expected exactly 1|found 2'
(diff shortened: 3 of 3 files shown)


sabiut (changes requested):
I pulled the branch down and walked through it locally. I traced it against the other tray-related functions and manually exercised the sed and idempotency paths. From a logic perspective, I'm happy with it. The $ escaping holds up (I tried uPi, u$i, and $e), a second run is a clean no-op, and it doesn't interfere with either patch_tray_inplace_update or the menu handler. If anything, the in-place fast path benefits from this change since it now re-selects the correct dark/light PNG on every rebuild.

There is one thing I'd like to see addressed before this merges, plus a few smaller observations that are optional.

## Blocker: the $-in-name path isn't covered by tests.

All of the current fixtures use `uPi`, which means this line:

```bash
sw_var_re="${sw_var//\$/\\$}"
```

never actually gets exercised.

I verified that if you remove the escaping entirely, all five existing tests still pass. However, with a switch variable such as `u$i`, the code breaks exactly as expected: sed interprets `$` as an end-of-line anchor, nothing matches, the icon remains broken, and the function still returns success. That's the entire reason the escaping exists, yet it's the one branch the test suite doesn't currently validate.

Could you add a fixture with a `$` in the switch variable name (for example `u$i`) and assert both the wrapped discriminant and a successful `node --check`?

## Smaller observations (optional)

- Test 2 ("the routed-to png case still selects the theme-aware PNG") is effectively checking the fixture against itself — the sed never touches the `png` case body, so that grep can't fail from a regression in the actual fix. I'd either give it a real wrap assertion or drop it.
- The anchor `switch\([$\w]+\)\{case"ico":` is fairly strict: no `\s*` tolerance, and it assumes `ico` is the first case. esbuild gives us no spaces today and it fails safe (warns and skips rather than corrupting anything), so this is robustness-only — but a spaced-shape fixture would lock in the intent, and CLAUDE.md does nudge toward handling both minified and beautified spacing.
- Minor: the `[[ "$status" -eq 0 ]]` checks quietly become no-ops on bash 3.2 (the macOS default) when they aren't the last command — `[ ... ]` or a trailing `|| false` keeps them honest.

The patch itself is solid; I just want the tests to actually exercise the thing they're guarding. I haven't run it on a real black-square host yet — I'll give it a proper look once #751 is in too.

davisilvarafacho:
Independently hit this and arrived at the same root cause before finding this PR — confirming the fix works.

**Environment:** Linux Mint 22.3 (Cinnamon 6.6.7, X11), panel theme `Mint-Y-Dark-Red`, Claude Desktop `1.15962.1`, host `xapp-sn-watcher`.

**Pre-fix evidence (DBus):** the SNI registered fine and was `Active`, but the item shipped *no* icon — `org.kde.StatusNotifierItem.GetAll` returned `IconPixmap = [(0, 0, [])]` and no `IconName` key at all. So on Cinnamon's `xapp-sn-watcher` it manifests as a *missing* icon rather than a black square, but it's the same cause: `switch(IPi)` with `IPi` baked to `"ico"`, never reassigned on Linux → `Tray-Win32.ico` → empty `nativeImage`. The existing `patch_tray_icon_selection` had silently no-op'd (anchored on the dead `template-image` branch), matching your "logged green while broken" note.

**Verified:** applied the png-routing approach, rebuilt the `.deb`, reinstalled — tray icon now renders correctly (theme-aware light/dark). Built bundle confirms `switch(process.platform==="linux"?"png":IPi){…}`.

Likely also closes #604, #557, #588.

---

brianjcohen:
**Runtime confirmation + the `$`-name test @sabiut asked for**

Independent repro and root-cause match on **KDE Plasma 6.7.1 (Wayland), Claude Desktop `1.15962.1-2.0.22.fc42`, Electron 41.5.0** (openSUSE Tumbleweed, host watcher `org.kde.StatusNotifierWatcher`, app on `--ozone-platform=x11`):

- **DBus:** the SNI registers and is `Active`, but ships no icon — `org.kde.StatusNotifierItem` `GetAll` → `IconPixmap = [(0, 0, [])]` and no usable `IconName`. Same signature @davisilvarafacho saw on Cinnamon / `xapp-sn-watcher`.
- **strace (`-e openat`):** startup opens `…/resources/Tray-Win32.ico` (fd returned), probes `Tray-Win32@{1.25,1.33,2,3,…}x.ico` (all `ENOENT`), and opens **zero** `TrayIconTemplate*.png`. So the `"ico"` case is taken, the Windows `.ico` loads but yields an empty `nativeImage` → the empty pixmap above. Exactly the chain this PR fixes.
- **Cross-build datapoint:** on `1.15962.1` the discriminant is minified to **`IPi`** (`…,uPi=36,IPi="ico",EPi=!0,CPi="registry"`), not `uPi` as on the `1.15962.0` you verified against — the minifier renamed it. The dynamic `switch(VAR){case"ico":` extraction picks it up cleanly, which is a nice confirmation that extracting `VAR` (never hardcoding it) is the right call.

**On the review blocker** (@sabiut's point that the `$`-in-name path is never exercised): here's a drop-in `tests/tray-icon-format.bats` update. It parameterises the fixture's discriminant and adds a `u$i` case. I confirmed it's a real guard — removing the `sw_var_re="${sw_var//\$/\\$}"` escaping from `patch_tray_icon_selection` fails **only** the new test while the other six stay green:

```
not ok 3 icon-format: handles a $-containing minified discriminant (u$i)
ok 1,2,4,5,6,7  (unchanged)
```

It also folds in the smaller notes:
- the self-checking test now makes a real transform assertion (bare `switch(uPi){` is gone + wrapped form present + png case intact);
- intermediate assertions use `[[ … ]] || return 1` so they stay honest on bash 3.2 — keeping `[[ ]]` per the bash style guide rather than switching to POSIX `[ ]`;
- a fail-safe canary documents the warn-and-skip contract for a beautified/spaced switch shape.

Full suite: **329/329**. No change to the patch itself — tests only.

<details><summary>diff — <code>tests/tray-icon-format.bats</code></summary>

```diff
diff --git a/tests/tray-icon-format.bats b/tests/tray-icon-format.bats
index a46f3ea..190f4b3 100644
--- a/tests/tray-icon-format.bats
+++ b/tests/tray-icon-format.bats
@@ -11,6 +11,14 @@
 # so the tray loads the Windows .ico and renders as a black square on the
 # freedesktop/KDE StatusNotifier. The patch wraps the switch discriminant
 # so Linux takes the already-theme-aware "png" case.
+#
+# Assertion style: intermediate checks are written `[[ … ]] || return 1`.
+# Bats only fails a test on the *last* command's status unless its ERR
+# trap fires, and that trap does not fire for an intermediate compound
+# command on bash 3.2 (the macOS default) — so a bare mid-test `[[ … ]]`
+# silently becomes a no-op there. `|| return 1` fails the test on every
+# bash while keeping `[[ ]]` per the project bash style guide (POSIX `[ ]`
+# is a documented anti-pattern).
 
 SCRIPT_DIR="$(cd "$(dirname "${BATS_TEST_FILENAME}")" && pwd)"
 TRAY_SH="$SCRIPT_DIR/../scripts/patches/tray.sh"
@@ -39,14 +47,21 @@ teardown() {
 }
 
 # Minified fixture mirroring the 1.13576+ icon-format switch: a baked
-# uPi="ico" constant, the three-case switch (Windows .ico / template /
+# <var>="ico" constant, the three-case switch (Windows .ico / template /
 # theme-aware png), and the Tray creation that consumes the result.
+#
+# The discriminant variable name is parameterised ($1, default uPi) so the
+# same shape can be emitted with a `$`-containing minified identifier
+# (e.g. u$i) to exercise the sed `$`-escaping branch. The heredoc is
+# intentionally unquoted so ${sw} interpolates; the JS body contains no
+# other `$`, backtick, or backslash, so nothing else expands.
 write_icon_switch_fixture() {
-	cat > "$index_js" <<'JS'
+	local sw="${1:-uPi}"
+	cat > "$index_js" <<JS
 const aA=require("electron");
-const gPi=!0,lPi=36,uPi="ico",IPi=!0;
+const gPi=!0,lPi=36,${sw}="ico",IPi=!0;
 function rebuild(){
-let e;switch(uPi){case"ico":e=aA.nativeTheme.shouldUseDarkColors?"Tray-Win32-Dark.ico":"Tray-Win32.ico";break;case"template-image":e="TrayIconTemplate.png";break;case"png":e=aA.nativeTheme.shouldUseDarkColors?"TrayIconTemplate-Dark.png":"TrayIconTemplate.png";break}
+let e;switch(${sw}){case"ico":e=aA.nativeTheme.shouldUseDarkColors?"Tray-Win32-Dark.ico":"Tray-Win32.ico";break;case"template-image":e="TrayIconTemplate.png";break;case"png":e=aA.nativeTheme.shouldUseDarkColors?"TrayIconTemplate-Dark.png":"TrayIconTemplate.png";break}
 const t=X.join(toi(),e);
 vE=new aA.Tray(aA.nativeImage.createFromPath(t));
 }
@@ -57,21 +72,49 @@ JS
 	write_icon_switch_fixture
 	cd "$TEST_TMP"
 	run patch_tray_icon_selection
-	[[ "$status" -eq 0 ]]
-	echo "$output" | grep -q 'Found icon-format switch var: uPi'
+	[[ "$status" -eq 0 ]] || return 1
+	echo "$output" | grep -q 'Found icon-format switch var: uPi' || return 1
 	grep -qF 'switch(process.platform==="linux"?"png":uPi){case"ico":' \
-		"$index_js"
+		"$index_js" || return 1
 	node --check "$index_js"
 }
 
-@test "icon-format: the routed-to png case still selects the theme-aware PNG" {
+@test "icon-format: rewrites the bare discriminant in place (real transform)" {
 	write_icon_switch_fixture
 	cd "$TEST_TMP"
 	patch_tray_icon_selection
-	# The case Linux now lands on must keep the dark/light PNG template
-	# selection — i.e. the fix is inert if the png case lost it.
+	# A genuine transform assertion, not a fixture-vs-itself check: the
+	# original bare `switch(uPi){` must be gone (proving the sed matched
+	# and rewrote it), the wrapped discriminant must be present, and the
+	# "png" case Linux now lands on must keep its dark/light PNG selection
+	# — the fix is inert if any of these regress.
+	! grep -qF 'switch(uPi){case"ico":' "$index_js" || return 1
+	grep -qF 'switch(process.platform==="linux"?"png":uPi){case"ico":' \
+		"$index_js" || return 1
 	grep -qF 'case"png":e=aA.nativeTheme.shouldUseDarkColors?"TrayIconTemplate-Dark.png":"TrayIconTemplate.png"' \
-		"$index_js"
+		"$index_js" || return 1
+	node --check "$index_js"
+}
+
+@test "icon-format: handles a \$-containing minified discriminant (u\$i)" {
+	# The minifier can emit identifiers containing `$` (e.g. u$i). The
+	# patch escapes it before building the sed pattern
+	# (sw_var_re="${sw_var//\$/\\$}"). Without that escaping, sed reads `$`
+	# as an end-of-line anchor, the pattern never matches, and the
+	# function still returns success — the exact silent no-op this guards.
+	write_icon_switch_fixture 'u$i'
+	cd "$TEST_TMP"
+	# Sanity: the fixture really emitted the $-name and is valid JS.
+	grep -qF 'switch(u$i){case"ico":' "$index_js" || return 1
+	node --check "$index_js" || return 1
+	run patch_tray_icon_selection
+	[[ "$status" -eq 0 ]] || return 1
+	echo "$output" | grep -qF 'Found icon-format switch var: u$i' || return 1
+	grep -qF 'switch(process.platform==="linux"?"png":u$i){case"ico":' \
+		"$index_js" || return 1
+	# Original bare form gone, and the result still parses.
+	! grep -qF 'switch(u$i){case"ico":' "$index_js" || return 1
+	node --check "$index_js"
 }
 
 @test "icon-format: idempotent — second run is a no-op" {
@@ -79,9 +122,10 @@ JS
 	cd "$TEST_TMP"
 	patch_tray_icon_selection
 	run patch_tray_icon_selection
-	[[ "$status" -eq 0 ]]
-	echo "$output" | grep -q 'already routed to png on Linux'
-	[[ "$(grep -coF 'process.platform==="linux"?"png":' "$index_js")" -eq 1 ]]
+	[[ "$status" -eq 0 ]] || return 1
+	echo "$output" | grep -q 'already routed to png on Linux' || return 1
+	[[ "$(grep -coF 'process.platform==="linux"?"png":' "$index_js")" -eq 1 ]] \
+		|| return 1
 	node --check "$index_js"
 }
 
@@ -92,10 +136,10 @@ function rebuild(){const e="TrayIconTemplate.png";vE=new aA.Tray(aA.nativeImage.
 JS
 	cd "$TEST_TMP"
 	run patch_tray_icon_selection
-	[[ "$status" -eq 0 ]]
-	echo "$output" | grep -qi 'WARNING'
+	[[ "$status" -eq 0 ]] || return 1
+	echo "$output" | grep -qi 'WARNING' || return 1
 	# Must not have injected a bogus discriminant.
-	! grep -qF 'process.platform==="linux"?"png":' "$index_js"
+	! grep -qF 'process.platform==="linux"?"png":' "$index_js" || return 1
 }
 
 @test "icon-format: warns when more than one icon-format switch exists" {
@@ -105,7 +149,40 @@ function other(){switch(zZ){case"ico":e=1;break}}
 JS
 	cd "$TEST_TMP"
 	run patch_tray_icon_selection
-	[[ "$status" -eq 0 ]]
-	echo "$output" | grep -qiE 'expected exactly 1|found 2'
-	! grep -qF 'process.platform==="linux"?"png":' "$index_js"
+	[[ "$status" -eq 0 ]] || return 1
+	echo "$output" | grep -qiE 'expected exactly 1|found 2' || return 1
+	! grep -qF 'process.platform==="linux"?"png":' "$index_js" || return 1
+}
+
+@test "icon-format: beautified/spaced switch shape is skipped safely (fail-safe)" {
+	# The anchor (switch(VAR){case"ico":) is deliberately strict because
+	# esbuild emits no spaces today. This locks in the fail-safe contract
+	# for a spaced/beautified shape: warn and leave the file byte-for-byte
+	# untouched rather than corrupt it. If whitespace-tolerant matching is
+	# ever added, this is the canary that must be updated alongside it.
+	cat > "$index_js" <<'JS'
+const aA = require("electron");
+const gPi = !0, lPi = 36, uPi = "ico", IPi = !0;
+function rebuild() {
+  let e;
+  switch (uPi) {
+    case "ico":
+      e = aA.nativeTheme.shouldUseDarkColors ? "Tray-Win32-Dark.ico" : "Tray-Win32.ico";
+      break;
+    case "png":
+      e = aA.nativeTheme.shouldUseDarkColors ? "TrayIconTemplate-Dark.png" : "TrayIconTemplate.png";
+      break;
+  }
+}
+JS
+	cd "$TEST_TMP"
+	local before
+	before=$(cat "$index_js")
+	run patch_tray_icon_selection
+	[[ "$status" -eq 0 ]] || return 1
+	echo "$output" | grep -qi 'WARNING' || return 1
+	# File left byte-for-byte untouched (no corruption), still valid JS.
+	[[ "$(cat "$index_js")" == "$before" ]] || return 1
+	! grep -qF 'process.platform==="linux"?"png":' "$index_js" || return 1
+	node --check "$index_js"
 }
```
</details>

*Written by Claude Opus 4.8 via [Claude Code](https://claude.ai/code)*

sabiut (commented):
Re-reviewed after the v3.0.0 rebase, and this is now resolved upstream rather than by us: the official 1.18286.0 bundle bakes the icon-format discriminant to `"png"` (`UVr="png"` — the `"ico"` branch is dead code), the png case selects new Linux-specific assets with a GNOME branch (`TrayIconLinux.png` / `TrayIconLinux-Dark.png`, both shipping in the deb's `resources/`), and the rebase deleted `scripts/patches/tray.sh` + the tray suite this PR patches. Same convergence we saw with the SNI re-registration race.

So: the root cause both @davisilvarafacho and @brianjcohen confirmed was real, this fix was correct for the 2.x line, and upstream shipped the equivalent. Recommend closing this PR, and closing #746 (likely #604/#557/#588 too) as fixed-upstream in 1.18286.0 once someone confirms on a v3.0.x install.

@brianjcohen — the `[[ … ]] || return 1` assertion pattern (and the mutation-tested `$`-name guard) resolved my earlier blocker exactly as asked; the launcher bats suites have the same hollow-assertion problem, so I'll carry that convention into a separate tests PR where it still has a home.

brianjcohen:
This is **fixed upstream** — recommending this be closed (I don't have close rights on the repo, over to @aaddrick / maintainers).

Agreed with @sabiut's re-review: the official `1.18286.0` Linux `.deb` bakes the icon-format discriminant to `"png"` (the `"ico"` branch is now dead code) and the png branch natively selects the new Linux assets `TrayIconLinux.png` / `TrayIconLinux-Dark.png` (Dark on GNOME/dark theme), and the v3.0.0 rebase deleted `scripts/patches/tray.sh` and its suite — so there's nothing left here to patch. Nice outcome: upstream converged on exactly this fix.

For anyone landing here from a blank/black-square tray: update to a v3.0.x build (official `1.18286.0`+) and it's handled natively — no patch needed.

Thanks @sabiut for the thorough review and @davisilvarafacho for the independent Cinnamon repro. Glad the `[[ … ]] || return 1` assertion convention and the mutation-tested `$`-name guard will find a home in the launcher-tests PR. Happy to post a runtime confirmation once I'm on a v3.0.x install, which should let #746 (and likely #604/#557/#588) close as fixed-upstream too.

*Written by Claude Opus 4.8 via [Claude Code](https://claude.ai/code)*

aaddrick:
Closing as obsoleted by the v3.0.0 rebase. This patches `tray.sh`, which was deleted wholesale — the patch-necessity matrix verdict for both tray rows is **delete**, and the official Linux build natively routes the Linux tray to the purpose-made `TrayIconLinux.png` / `TrayIconLinux-Dark.png` selection ([tray dossier, "Fate under the official-deb rebase"](https://github.com/aaddrick/claude-desktop-debian/blob/main/docs/reports/CDL-ANT-0009_patch-suite-history/dossiers/tray.md)) — which is exactly the routing this PR forced. #746 was closed on the same basis. The diagnosis here (build-time `"ico"` constant baked for the Windows build) stands as a correct record of why the Windows repackage was broken.

If a wrong-icon-format symptom somehow resurfaces on a 3.x build, rebase and reopen — though at that point it's upstream-report material under patch-zero ([D-002](https://github.com/aaddrick/claude-desktop-debian/blob/main/docs/decisions.md#d-002--rebase-onto-the-official-linux-deb-patch-zero)).

---

---

PR #759, opened 2026-07-01
scripts/patches: scaffold buddy-ble-stub.sh

deepybee:
Placeholder patch script sourced by build.sh so the WIP PR stays buildable while the BuddyBleTransport handler stub + change-listener fix land in follow-up commits (see PR body).

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -62,6 +62,8 @@ source "$script_dir/scripts/patches/quick-window.sh"
 source "$script_dir/scripts/patches/claude-code.sh"
 # shellcheck source=scripts/patches/cowork.sh
 source "$script_dir/scripts/patches/cowork.sh"
+# shellcheck source=scripts/patches/buddy-ble-stub.sh
+source "$script_dir/scripts/patches/buddy-ble-stub.sh"
 # shellcheck source=scripts/patches/org-plugins.sh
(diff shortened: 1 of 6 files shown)


aaddrick:
Post-v3.0.0 triage heads-up before more work lands here: the scaffold targets the old `build.sh` patch wiring, which was replaced at the rebase — patches are now wired through the `active_patches` array in [`scripts/patches/app-asar.sh`](https://github.com/aaddrick/claude-desktop-debian/blob/main/scripts/patches/app-asar.sh), fail the build on anchor miss, and (the bigger bar) every new asar patch must justify itself against official bytes as compensating a genuine Linux gap under patch-zero ([D-002](https://github.com/aaddrick/claude-desktop-debian/blob/main/docs/decisions.md#d-002--rebase-onto-the-official-linux-deb-patch-zero)) — the default verdict is that the official build owns its own behavior.

Is the BuddyBleTransport work still coming? If so: rebase onto current main, wire through `active_patches`, and include the D-002 justification (what breaks on Linux in the official build, and why a patch is the only fix) — ideally as an issue first so the direction gets agreed before the patch. If it's no longer planned, no response needed and this can close as stale in a couple of weeks; rebase and reopen any time if you pick it back up.

---

---

PR #761, opened 2026-07-01
fix(patches): spawn a working terminal shell on Linux (#727, #728)

### LINKED ISSUE #727: Linux: node-pty spawn-helper not built → Code-tab integrated terminal shell PTY exits code 1
## Summary

On the Linux build, opening the **Code tab → Local session → integrated terminal** (Ctrl+`) never yields a shell — the main log shows:

```
LocalSessions … Shell PTY … exited with code 1
```

Root cause: node-pty's **`spawn-helper` binary is not built/shipped for Linux**, so node-pty cannot fork the shell.

This is distinct from the closed #401 (which was about a *wrong-architecture* Windows `pty.node`, PE32+ instead of ELF). Even with a correct Linux `pty.node`, node-pty 1.1.0 **also** forks the shell through a separate `build/Release/spawn-helper` executable (`lib/unixTerminal.js`), and that one is still missing.

## Environment

- **Claude Desktop:** `@ant/desktop` **1.13576.0** (built from `main`)
- node-pty **1.1.0**
- Linux x86_64 (reproduced on Gentoo)

## Steps to reproduce

1. Build from `main` and install.
2. Open the **Code** tab → **Local session**.
3. Open the integrated terminal (Views → Terminal, or Ctrl+`).

**Observed:** terminal panel never gives a shell; log shows `Shell PTY … exited with code 1`.
**Expected:** a working shell.

## Root cause

node-pty 1.1.0 spawns the PTY child via `build/Release/spawn-helper` (see `node_modules/node-pty/lib/unixTerminal.js`). The Linux pack ships `pty.node` but **not** `spawn-helper`: `prebuildify` only captures the `.node` addon, and node-pty's published prebuilds are **darwin-only**, so a Linux build that doesn't compile `spawn-helper` from source omits it entirely.

(Separately, the Linux pack also carries stray Windows node-pty artifacts — `winpty-agent.exe`, `winpty.dll`, `conpty.node`, `conpty_console_list.node` — which are dead weight on Linux and should be dropped; related to #401.)

## Fix

`spawn-helper` is standalone C++ with no Node headers, so it compiles directly:

```sh
# from the node-pty package dir
g++ -O2 -o spawn-helper src/unix/spawn-helper.cc
```

Stage the result next to `pty.node` (mode 0755):

```
app.asar.unpacked/node_modules/node-pty/build/Release/spawn-helper
```

i.e. in `build.sh`'s `install_node_pty()`, after placing the Linux `pty.node`, also compile and install `spawn-helper`, and remove the stray Windows artifacts.

## Verification

With `spawn-helper` compiled and staged (and the companion shell-selector fix, see the linked issue), the integrated terminal works — verified with `bash` and `fish`.

### LINKED ISSUE #728: Linux: integrated-terminal shell hardcoded to powershell.exe → shell PTY exits code 1
## Summary

On the Linux build, the **Code tab → Local session → integrated terminal** never yields a shell; the log shows `Shell PTY … exited with code 1`. One root cause is that the integrated-terminal **shell is hardcoded to `powershell.exe`** with no platform branch, so on Linux node-pty tries to `exec powershell.exe`, which doesn't exist.

## Environment

- **Claude Desktop:** `@ant/desktop` **1.13576.0** (built from `main`)
- node-pty **1.1.0**
- Linux x86_64 (reproduced on Gentoo)

## Steps to reproduce

1. Build from `main` and install.
2. Open the **Code** tab → **Local session**.
3. Open the integrated terminal (Views → Terminal, or Ctrl+`).

**Observed:** no shell; `Shell PTY … exited with code 1`.
**Expected:** a working login shell.

## Root cause

The main bundle's shell selector in `.vite/build/index.js` returns a fixed Windows shell with no platform branch:

```js
function Ltt(){return{shell:"powershell.exe",args:[]}}
```

On non-Windows there is no `powershell.exe`, so node-pty's exec fails and the PTY exits 1.

## Fix

Branch on platform / use the login shell:

```js
function Ltt(){return{shell:process.env.SHELL||"/bin/bash",args:[]}}
```

(When this is patched inside `app.asar` post-build, the file's ASAR integrity block must be refreshed or Electron rejects the modified archive; fixing it in the source bundle avoids that.)

## Verification

With this change (and the companion node-pty `spawn-helper` fix in the linked issue), the integrated terminal works — verified with `bash` and `fish`.

---
Companion issue (same symptom, the other root cause): #727

LiukScot:
## What

Makes the Code-tab integrated terminal actually spawn a shell on Linux.
Symptom: the panel shows "Shell exited" and the main log has
`Shell PTY … exited with code 1`.

## Root causes (both Linux-only, both required)

1. **`spawn-helper` never built (#727).** node-pty 1.1.0 forks the shell
   through `build/Release/spawn-helper`, but its `binding.gyp` gates that
   target to `OS=="mac"` — so on Linux neither `npm install` nor
   `node-gyp rebuild` produce it, while `lib/unixTerminal.js` execs it on
   every unix. `install_node_pty` now compiles it from source with `g++`.
2. **Shell hardcoded to `powershell.exe` (#728).** The selector had no
   platform branch; `patch_linux_terminal_shell` repoints it at
   `process.env.SHELL||"/bin/bash"`.

Distinct from #401/#597 (wrong-arch Windows `pty.node`) — that fix stays.

## Changes
- `scripts/patches/cowork.sh` — compile `spawn-helper` in `install_node_pty`
- `scripts/patches/claude-code.sh` — `patch_linux_terminal_shell`
- `scripts/patches/app-asar.sh` — call the new patch
- `scripts/cowork-patch-markers.tsv` — `terminal-shell-linux` marker
- `nix/node-pty.nix` — same `spawn-helper` compile for the Nix path
- `CHANGELOG.md` — Unreleased/Fixed entry

## Testing
Validated the #728 sed against the live upstream bundle (1.15962.1):
`shell:"powershell.exe"` appears exactly once (the other `powershell.exe`
occurrences are Windows shell-detection lists, not `shell:` selectors),
1→0 after the patch, marker present. `spawn-helper` compiles to ELF and
the terminal spawns bash/fish after staging into `app.asar.unpacked`.
Applied the same patched `app.asar` to a running RPM install (1.15962.1)
and confirmed the terminal works end to end. Full `./build.sh` not run
locally; `shellcheck`/`bats` not available in the dev environment
(`bash -n` passes on all changed scripts).

Fixes #727
Fixes #728

---
85% AI / 15% Human
Claude: root-cause analysis, patch implementation, marker, validation against the live bundle, Nix path, CHANGELOG/PR text
Human: reported the bug, confirmed the fix works on the running install, directed scope


EtherAura:
Original reporter of #727/#728 here. Confirming this matches what we've been shipping in a Gentoo overlay (`app-misc/claude-desktop`, a live ebuild built from this repo's `main`) since ~2026-06-18 — both fixes, essentially identical to this PR:

- **spawn-helper (#727):** node-pty 1.1.0's `binding.gyp` gates the `spawn-helper` target to `OS=="mac"`, so a from-source Linux build never produces `build/Release/spawn-helper`, yet `lib/unixTerminal.js` execs it on every unix → PTY exits 1. We compile it standalone with the toolchain compiler (`-O2`, no node headers needed) and stage it next to `pty.node` in `app.asar.unpacked/.../node-pty/build/Release/`. Same as this PR's `install_node_pty` change.
- **shell selector (#728):** rewritten to `process.env.SHELL||"/bin/bash"`. We patch the single `{shell:"powershell.exe",args:[]}` anchor in the already-built `app.asar` and refresh that file's ASAR integrity block — equivalent to patching pre-pack as this PR does.

Both verified deployed and working end-to-end with `bash` and `fish`.

One note re: the #727 scope discussion above (a prebuilt RPM's `pty.node` reportedly not needing `spawn-helper`): for the **build-from-installer** path this PR targets, the helper is definitely required — without it the PTY exits 1, with it the shell spawns. So keeping the compile here is correct.

One durability suggestion for `patch_linux_terminal_shell`: make it a **no-op (not a hard error) when the `shell:"powershell.exe"` selector is already absent**, since Anthropic's own Windows/macOS bundle has already moved to a `pWi()`-style resolver — a future installer may ship without the hardcoded selector, and you'd want the patch step to skip cleanly rather than fail the build. (Still hard-fail if `shell:"powershell.exe"` is present but under a shape your matcher doesn't expect — that's real drift.) We just hit exactly this edge and made our patch fail-soft on it.

Thanks for pushing this over the line 🙏

aaddrick:
Hey @LiukScot! Closing this as superseded, not rejected. The root-cause work here is solid and the write-up made both bugs easy to follow.

Here's what happened. Three days after you opened this, the v3.0.0 rebase onto Anthropic's official `.deb` merged. That rebase deleted 4 of the 5 files this PR edits (`scripts/patches/cowork.sh`, `scripts/patches/claude-code.sh`, `scripts/cowork-patch-markers.tsv`, `nix/node-pty.nix`) — the whole cowork/claude-code patch layer is gone. It also pulled a newer upstream bundle (`claude-desktop 1.18286.0`, node-pty 1.1.0 → 1.2.0-beta.13) that already fixes both underlying bugs.

The project is patch-zero now: we only carry a patch if it compensates a genuine Linux gap in the official bytes. Both #727 and #728 verify as fixed against the installed build, so there's no gap left to patch and no patch files left to apply it to.

I've closed #727 and #728 as fixed-by-rebase with the evidence in each. Thanks for the diagnosis work here. It held up. It just landed on the same spot upstream got to on its own.

---

PR #765, opened 2026-07-04
fix(nix): put the glvnd EGL dispatcher on every ELF's runpath

typedrat:
## Problem

The v3.0.0 Nix derivation built cleanly (`autoPatchelf` fully satisfied) but **crash-looped at startup on real NixOS**: the GPU process failed EGL init with `Could not dlopen native EGL: libEGL.so.1`, exited (`Exiting GPU process due to errors during initialization`), and relaunched forever.

## Root cause

Chromium's bundled **ANGLE** (the co-located `libEGL.so`/`libGLESv2.so`) `dlopen`s the glvnd dispatcher `libEGL.so.1` by bare soname at GPU init. A `dlopen` resolves against the **calling object's** `DT_RUNPATH` — which on those libs carries only their own `DT_NEEDED`, not `libGL`. And `runtimeDependencies` can't reach them: `auto-patchelf.py` appends it to dynamic **executables only** (`if file_is_dynamic_executable: rpath += runtime_deps`), so `libGL` landed on the main ELF but never on the `.so` that issues the dlopen.

## Fix

```nix
appendRunpaths = [
  "${lib.getLib libGL}/lib"
  "${addDriverRunpath.driverLink}/lib"
];
```

`appendRunpaths` is applied to *every* patched file. Once ANGLE can load glvnd's `libEGL.so.1`, NixOS-patched glvnd self-locates the vendor ICD under `/run/opengl-driver`, so the second hop needs no extra wiring.

Chosen over a `makeWrapper --suffix LD_LIBRARY_PATH` (what nixpkgs' `discord` does) because the app spawns MCP servers (`node`/`uv`/`docker`) and a wrapper would leak the driver tree into their environment. A runpath edit is scoped to the ELFs that need it; `bin/claude-desktop` stays a clean symlink.

## Verification (real NixOS x86_64, nvidia)

- Both flake outputs (`claude-desktop`, `claude-desktop-fhs`) launch with clean GPU/EGL init — **0** EGL failures / GPU exits (down from a nonstop loop).
- Locales load; `process.resourcesPath` intact (no wrapper indirection).
- Eval clean on x86_64 (both) and aarch64.

## Doc corrections (verified against a local nixpkgs checkout)

- `signal-desktop` is **not** a vendored-Electron precedent — it's a *source* build on nixpkgs `electron_42`. The real precedents are **discord** and **vscode**. Corrected in `nix.md` + the file header.
- The `resourcesPath` blocker is broader than documented: `virtiofsd`, `cowork-linux-helper`, and the `smol-bin.*.img` Cowork VM images all resolve from `process.resourcesPath` with no fallback — not just the locale JSONs. Recorded, along with the ANGLE GL trap, as a new learning.

---
90% AI / 10% Human
Claude: root-caused the EGL crash-loop on real NixOS, proved the RPATH-vs-dlopen mechanism and the executable-only runtimeDependencies behavior against nixpkgs source, implemented and verified the appendRunpaths fix on both outputs, corrected the verified doc errors.
Human: ran the build, reported the crash, steered away from the LD_LIBRARY_PATH wrapper toward a runpath edit, challenged the nixpkgs-Electron and signal-desktop claims.


aaddrick (approved):
Approving. The fix is right and the root-cause writeup holds up. A dlopen of a bare soname resolves against the calling object's runpath, and nix's `runtimeDependencies` only reaches dynamic executables, never the co-located ANGLE `.so`. So `appendRunpaths` is the correct knob. Since it appends, it's dangling-safe and can't shadow a lib that already resolves. I walked the mechanism and it checks out, and the live Cowork session you got on your box is a solid shakeout of the FHS GL path on nvidia.

That said, keep the hedge in the CHANGELOG and nix.md before you merge. A working session confirms the path held on nvidia, but the box was nvidia, and the failing path is ANGLE's native-GL backend. Mesa (Intel/AMD, which is most NixOS users) picks GL backends differently, so that's still the one config that didn't get exercised. I'd word it as "verified x86_64 + nvidia; mesa and aarch64 unconfirmed."

Worth a breadcrumb too. If a mesa box ever lands on SwiftShader instead of hardware GL, the likely missing piece is the Vulkan ICD search path. Discord exposes that via `VK_ADD_DRIVER_FILES`, which a runpath edit can't reproduce. I'm not asking you to wire it now. That'd drag back the wrapper env-leak this PR deliberately sidesteps. Just a note so future-us isn't surprised.

This one went through a devil's-advocate pass along with the other two in the stack, and the mechanism survived it. Fold the narrowed hedge in when you revise and it's good to go.

PR #766, opened 2026-07-04
feat(nix): ship qemu in the FHS env so Cowork can boot its VM

typedrat:
**Stacked on #765** (base = `typedrat/nix-angle-egl-runpath`).

## Problem

The v3.0.0 FHS env bind-provides OVMF firmware at Cowork's hardcoded probe path, but that satisfies only *half* the gate. From the shipping bundle:

- **JS gate:** `E3A = {qemuPath, firmwarePath, virtiofsdPath}` → returns `requirement_missing` when `!qemuPath || !firmwarePath`. `qemuPath` comes from searching PATH for `qemu-system-x86_64` (arm64: `-aarch64`).
- **`coworkd` (static Go)** launches a real KVM guest: `-drive if=pflash,...,file=<OVMF>` + `accel=kvm` + `vhost-vsock-pci,guest-cid=` + virtiofsd `--shared-dir` + slirp usermode net.

So with firmware present but no `qemu-system-*` on PATH, Cowork's own check fails and the VM never boots.

## Fix

Add **`qemu_kvm`** (host-cpu-only — ships exactly the arch's `qemu-system-*`, ~1.5 GB closure vs 2.1 GB for the all-targets `qemu`) to `nix/fhs.nix` `targetPkgs`. No device wiring needed: `buildFHSEnv` already binds the whole `/dev` (`--dev-bind /dev /dev`), so `/dev/kvm` and `/dev/vhost-vsock` are reachable.

## Verification

- `qemu-system-x86_64` lands at the FHS `/usr/bin` (what Cowork searches) → `qemuPath` resolves.
- OVMF shim still at `/usr/share/OVMF/OVMF_CODE{,_4M}.fd` → `firmwarePath` resolves.
- Both gate requirements now pass. **Live VM boot is still unverified** — needs a real Cowork session with `/dev/kvm` on a KVM host; can't be proven headlessly.

## Note on the non-FHS output

The firmware path is hardcoded to `/usr/share/OVMF/...` with **no env override** (verified: the only Cowork env vars are `COWORK_CU_ONLY`, `COWORK_PROFILE_TURNS`, `COWORK_VM_DEBUG`, `CLAUDE_CODE_USE_COWORK_PLUGINS`). On NixOS the non-FHS `claude-desktop` output has no writable `/usr/share` and no bind layer, so Cowork can't be satisfied there — which is exactly why the FHS env owns Cowork. The app itself (GPU, chat, MCP) runs fine on the non-FHS output.

---
85% AI / 15% Human
Claude: traced Cowork's two-part VM-boot gate in the bundle + coworkd, confirmed /dev is fully bound by buildFHSEnv, picked qemu_kvm over qemu on closure grounds, wired it in and verified both gate requirements resolve.
Human: chose to add qemu to the default FHS env over a separate variant; asked whether the non-FHS output could support Cowork.


aaddrick (approved):
Approving. `qemu_kvm` over the all-targets `qemu` is the right call, and the two facts that mattered both hold. The host-cpu-only build still ships `qemu-system-x86_64` and `qemu-system-aarch64` named exactly the way Cowork's PATH search expects, and libslirp survives the slim build, so usermode net stays intact. And now that it's actually booted a VM on your machine, you can upgrade the CHANGELOG claim: usermode net and the qemu binaries are x86_64 live-verified, not just static-checked. Nice.

One reframe on the docs before merge. The old "`/dev/kvm` and `/dev/vhost-vsock` need no wiring" line still needs softening, because those two prereqs are exactly what had to be true for it to boot on your box. `/dev/kvm` is `root:kvm 0660`, so a user outside the kvm group gets EACCES, and coworkd's `accel=kvm` is a hard requirement. And `/dev/vhost-vsock` doesn't exist until the `vhost_vsock` module is loaded, which isn't a NixOS default. It worked for you because your host already satisfied both. They won't be present for every user.

`--doctor` already checks both, so the docs should spell out what the host still has to provide: "reachable inside the env, but the host must grant kvm-group access and load `vhost_vsock`." Reword that line and I'm happy.

typedrat:
Rebased onto latest `main` (the ANGLE/EGL commits it sat on are now merged, so those dropped out) and reworded the docs per review:

- **CHANGELOG**: the qemu bullet now claims **x86_64 live-verified** — a VM boots with KVM acceleration, and both the qemu binaries and usermode networking work — instead of the old static-checked wording. aarch64 stays flagged unverified.
- **`/dev/kvm` + `/dev/vhost-vsock`** (CHANGELOG, `docs/learnings/nix.md`, and the `nix/fhs.nix` comment): dropped the "need no wiring" framing. All three now say the nodes are reachable inside the env (buildFHSEnv binds `/dev`) **but the host must still grant kvm-group access** (`/dev/kvm` is `root:kvm 0660`, else `EACCES`) **and load `vhost_vsock`** (no node until the module is in) — with a note that `--doctor` flags both.
- `nix.md`'s "Current state" paragraph also picks up the x86_64 Cowork VM boot alongside the existing nvidia note.

---

PR #767, opened 2026-07-04
fix(nix): ship the OVMF VARS template so Cowork's efivars seed exists

typedrat:
**Stacked on #766** (base = `typedrat/nix-cowork-qemu`).

Started as the aarch64 firmware-naming open item; the investigation surfaced a **bigger, arch-independent bug** in the same shim.

## The bug

Cowork needs two firmware files per VM: a read-only **CODE** pflash and a writable **VARS** store. It finds CODE at the hardcoded probe path, then derives the VARS *template* beside it by renaming:

```js
Qgi = A => A.replace("OVMF_CODE","OVMF_VARS").replace("AAVMF_CODE","AAVMF_VARS")
// firmwareCodePath = firmwarePath;  firmwareVarsTemplatePath = Qgi(firmwarePath)
```

`coworkd` reads that template and copies it per VM to seed efivars, aborting with **`no EFI variable-store template configured`** if it's missing. The `ovmfCompat` shim symlinked **only CODE**, so the derived VARS path (`/usr/share/OVMF/OVMF_VARS_4M.fd`, arm64 `/usr/share/AAVMF/AAVMF_VARS.fd`) didn't exist on **either** arch. deb/rpm hide this because the distro's edk2 package already drops `OVMF_VARS` beside the CODE file (CW-1 only ever needed the CODE symlink); on the Nix FHS the shim is the sole source.

## The fix

Ship the matched **CODE+VARS pair**:
- **x86_64:** alias both Debian names for each half onto the single 4M nixpkgs build (`OVMF_CODE{,_4M}.fd`, `OVMF_VARS{,_4M}.fd` → `FV/OVMF_{CODE,VARS}.fd`).
- **aarch64:** the 64 MiB pflash-padded `AAVMF_{CODE,VARS}.fd` (verified present in current nixpkgs). **Dropped the `QEMU_EFI.fd` fallback** — it's the unpadded 3 MiB raw image (qemu `-drive if=pflash` would reject it) and has no matching `QEMU_VARS` name for the rename derivation.
- A **build-time guard** now fails the build if a source `FV/*.fd` is gone, instead of shipping a dangling symlink that only surfaces at VM boot.

## Verification

- x86_64: all four `/usr/share/OVMF/*.fd` paths resolve to real files, and Cowork's exact derived VARS path (`OVMF_CODE_4M.fd` → `OVMF_VARS_4M.fd`) lands on a real file. FHS env builds.
- aarch64: built the shim against the aarch64 nixpkgs OVMF — guard passes, both `AAVMF_CODE.fd` and `AAVMF_VARS.fd` resolve to 64 MiB firmware.
- **Still unverified:** an actual Cowork VM boot (needs an aarch64 KVM host + live session). This closes the *firmware-completeness* half of that gap on both arches.

---
80% AI / 20% Human
Claude: traced Cowork's CODE→VARS template derivation (Qgi) and coworkd's firmwareVarsTemplatePath, found the shim shipped only CODE, fixed both arches with a guarded matched-pair shim, dropped the wrong QEMU_EFI fallback, verified shim output on both arches.
Human: asked me to take on the aarch64 firmware open item, which led here.


aaddrick:
just to confirm, these three PR's landed you on Cowork working locally?

aaddrick (approved):
Approving, and congrats. This is the strongest of the three and the one that landed the whole stack. It's a real, non-obvious bug (the CODE-only shim yielding "no EFI variable-store template configured"), and the CODE to VARS rename trace holds under scrutiny, including the AAVMF_CODE no-op edge in the first `.replace`. Failing the build loud instead of shipping a dangling symlink is the right instinct, and the live boot on your box is the proof it all wires together.

One doc note that still stands. The guard means the aarch64 FHS output fails the build on OVMF-layout drift, say a pinned nixpkgs missing the AAVMF pair, rather than shipping a dangling link that only bit at VM boot. That's the correct trade. Fail-loud on an unverified arch beats a boot-time firmware error, and there's no clean fallback since QEMU_EFI is unpadded and qemu rejects it for pflash. It's a build-behavior change for pinned aarch64 consumers, so the CHANGELOG should call it out.

Stack-level, the framing flips in your favor. The three now boot Cowork end-to-end live on x86_64, so the CHANGELOG can claim that win plainly. This is genuinely done there. Keep aarch64 and non-nvidia GPUs marked as the remaining open items, and let those entries keep their "unverified" hedge until one boots. All three cleared the devil's-advocate pass. Merge them in stack order whenever you're ready, and nice work landing the payoff.

---

This was Claude generated... where'd my disclaimer go?

typedrat:
Restacked onto `main` now that #766 has merged (base retargeted from `typedrat/nix-cowork-qemu` → `main`; the qemu commits it sat on are upstream, so they dropped out) and applied the review:

- **CHANGELOG** now calls out the aarch64 build-behavior change explicitly: the build-time guard **fails the build** on nixpkgs OVMF-layout drift (e.g. a pinned nixpkgs missing the AAVMF pair) rather than shipping a dangling symlink — flagged as fail-loud on the unverified arch, with the no-clean-fallback rationale (`QEMU_EFI` is unpadded, qemu rejects it for pflash).
- **Stack framing updated**: with the qemu FHS entry landed, the entry now states plainly that Cowork boots a VM end-to-end on x86_64 with this shim in place, and keeps aarch64 marked unverified.
- `docs/learnings/nix.md`'s design-contract section picks up the same two points so it doesn't contradict the CHANGELOG.

Ready to merge whenever.

---

Issue #772, opened 2026-07-05
[bug]: Cowork fails on Crostini (Debian Bookworm 12.14) despite all dependencies present and verified


Blackhawke:
### Version (`claude-desktop-unofficial --doctor` output)

```shell
claude-desktop-unofficial --doctor

returns NULL.

No ouptut!
```

### What happened

**Environment:**

Device: Chromebook (Crostini/Linux container)
Distro: Debian Bookworm 12.14
Architecture: amd64
Claude Desktop: 1.18286.0 (official Anthropic apt repo)
Also tested: claude-desktop-unofficial (same result)

**Problem:**
Clicking the Cowork tab shows the error: "Cowork runs in a secure local virtual machine. Cowork requires QEMU. Install it with 'sudo apt install qemu-system-x86 ovmf virtiofsd', then restart Claude."
All dependencies are present and verified:

qemu-system-x86 1:7.2+dfsg-7+deb12u18+b2 — installed
ovmf 2022.11-6+deb12u2 — installed
/dev/kvm — exists
User is in kvm group — confirmed via groups
virtiofsd — present at /usr/lib/qemu/virtiofsd (v1.13.2) AND /usr/lib/claude-desktop/resources/virtiofsd (v1.13.2), both executable
Symlink created at /usr/bin/virtiofsd pointing to bundled binary — no effect

**Note on virtiofsd:**
apt-cache show virtiofsd returns "purely virtual" — the package name exists as a dependency reference but has no installable candidate in Crostini's repos. However the binary itself is present and functional at both paths above.
**Conclusion:**
Claude Desktop's dependency check appears to be checking for virtiofsd as an APT package rather than checking for the binary, causing it to report missing dependencies despite everything being installed and functional. The --doctor flag is not implemented in the official package.

### Steps to reproduce

Start desktop, hover over Cowork tab.

### Expected behavior

**Expected behavior:**

Cowork should launch successfully. All required components are present and functional — QEMU, OVMF, KVM device access, kvm group membership, and virtiofsd binary in both the system path and Claude Desktop's own bundled resources. The dependency check should verify that the required binaries exist and are executable, not that they are installable as APT packages. A binary that is present, executable, and returning a valid version string satisfies the dependency regardless of its APT package status.

### Logs / errors

```shell
NOTE THAT THIS LOG CONTAINS NO CO-WORK SPECIFIC ERRORS -- WHICH IS ITSELF VERY DIAGNOSTIC!

[...]

PR #773, opened 2026-07-05
fix(#771,#772): un-gate Cowork's bundled virtiofsd; doctor mirrors the client probe

### LINKED ISSUE #771: [bug]: Cowork VM gated on Linux (yukonSilver unsupported) — regression from v2.x→v3.0.0 official-deb cutover

Note: on this AUR (Arch/CachyOS) package the binary is still named `claude-desktop`, not `claude-desktop-unofficial`. Full output of `claude-desktop --doctor`:

```
Claude Desktop Diagnostics

       Version drift: newest official pool version is 1.18286.0 (installed version unknown — AppImage?)
       Desktop: KDE
            (disables global hotkeys)
[PASS] Electron: v42.5.1 (/tmp/.mount_claudepUuEUK/usr/lib/claude-desktop/claude-desktop)
[FAIL] Chrome sandbox: perms=755, owner=root
       Fix: sudo chown root:root /tmp/.mount_claudepUuEUK/usr/lib/claude-desktop/chrome-sandbox
            sudo chmod 4755 /tmp/.mount_claudepUuEUK/usr/lib/claude-desktop/chrome-sandbox
       Password store: upstream os_crypt autodetect (default)
[PASS] Keyring: org.freedesktop.secrets reachable for credential encryption
[PASS] MCP config: valid JSON ($HOME/.config/Claude/claude_desktop_config.json)
       MCP servers configured: 0
[PASS] Node.js: v25.8.2
       Path: $HOME/.local/share/mise/installs/node/25.8.2/bin/node
[WARN] Desktop entry not found (expected for AppImage installs)
[PASS] Disk space: 127314MB free

Cowork Mode
----------------
[PASS] KVM: /dev/kvm present and accessible
[PASS] vsock: /dev/vhost-vsock present
[PASS] QEMU: qemu-system-x86_64 found
[PASS] Firmware: /usr/share/OVMF/OVMF_CODE_4M.fd
[PASS] virtiofsd: found at /usr/lib/virtiofsd (not on PATH)
       Cowork isolation: KVM (official)
[PASS] Log file: 70KB ($HOME/.cache/claude-desktop-debian/launcher.log)

1 check(s) failed.
See above for fixes.
```

Package: `3.0.0+claude1.18286.0-1` (AUR build), installed 2026-07-04, same day as the v2.x→v3.0.0 official-deb cutover (#762). Distro: CachyOS (Arch-based), KDE Plasma, Wayland session (XWayland default, the app's own default mode).

### What happened

Cowork's UI still shows the "requires QEMU" install prompt even though every local prerequisite `--doctor` checks for (KVM, vhost-vsock, QEMU binary, OVMF firmware, virtiofsd) passes. The chrome-sandbox FAIL above is unrelated/pre-existing and not what's blocking Cowork.

Digging into `~/.config/Claude/logs/cowork_vm_node.log` shows the real gate — the app spawns its native Linux helper, then aborts before using it:

```
[info] [linux-vm] Spawning helper: <mount>/usr/lib/claude-desktop/resources/cowork-linux-helper
[info] [linux-vm-helper] ... starting ... listening on /run/user/1000/claude-cowork-vm.sock
[info] [cleanupVMBundleIfUnsupported] yukonSilver not supported (status=unsupported), checking for stale bundle...
[info] [deleteVMBundle] Deleting reinstall files from .../vm_bundles/claudevm.bundle
[warn] [startVM] VM not supported (linux/x64), skipping
```

This is the same yukonSilver evaluator gate fixed twice before in #742 and #601 (both via `cowork.sh` Patch 1b, injecting an early `linux` return into the support evaluator). But those patches applied to the old v2.x Windows-repack lineage, which #762 says was frozen/obsoleted in the v3.0.0 official-deb cutover — the new lineage ships the official `app.asar` byte-identical aside from two unrelated compensating patches (KDE focus fix, plugins path). So the yukonSilver gap is back in the new, intentionally-unpatched lineage.

For comparison, this same machine's log from before the July 4 rebuild (old v2.x install) went through a different, generic VM startup path shared with the macOS build (`download_and_sdk_prepare` → `load_swift_api` → `sdk_install`) that reported "Startup complete" despite missing VM image files (`rootfs.vhdx`/`vmlinuz`/`initrd` all logged "missing") — so it was non-functional in a different way back then too. The native `cowork-linux-helper` code path is new to this lineage; it's just gated off.

No local override exists — `~/.config/Claude/config.json` has no `yukonSilver` key, so this is evaluated live on every launch rather than cached, and isn't something a local file edit can flip.

### Steps to reproduce

1. Install the v3.0.0 official-deb-based build on a Linux x86_64 host with a real working KVM/QEMU/OVMF/virtiofsd setup (confirm via `--doctor`, Cowork section all-PASS).
2. Launch Claude Desktop, open Cowork.
3. Observe the "Cowork requires QEMU" install prompt regardless of doctor's PASS state.
4. Check `~/.config/Claude/logs/cowork_vm_node.log` for the `yukonSilver ... unsupported` / `VM not supported (linux/x64), skipping` lines on that launch.

Not yet tested: reproducing against a clean `~/.config/Claude` (moved aside) as CONTRIBUTING.md asks, since #768's own thread flags that Cowork's `spaces.json`/trusted-folders/`mcpServers` have no backing store outside that directory and an empty-state relaunch can permanently stub them. Didn't want to risk that on a daily-driver install without checking first — happy to do it if it's useful and there's a safe way to snapshot/restore first.

### Expected behavior

Expected: Cowork's KVM-backed sandbox to actually start, given every local prerequisite doctor checks for is present and the app itself spawns the correct native helper for it.
Got: the yukonSilver evaluator short-circuits `startVM` before it uses that helper, so the UI shows the generic "install QEMU" prompt even on a machine where QEMU/KVM/OVMF/virtiofsd are all present and correct — which is actively misleading, since it tells the user to fix something that's already fixed.

### Logs / errors

See `cowork_vm_node.log` excerpt above. Also relevant, `cowork_vm_daemon.log` shows a separate `cowork-vm-service` (Unix-socket daemon, distinct from `cowork-linux-helper`) starting/stopping cleanly on every launch with no errors — that one isn't the blocker, just noting it in case it's a confusingly similarly-named component to whoever picks this up.

### Anything else

Given #762's stated design intent for v3.x — byte-identical to official `app.asar` except for patches compensating for "a real Linux gap" — this looks like it now qualifies as one, in the new lineage. Is a Patch 1b equivalent (or a `setYukonSilverConfig` intercept per the discussion in #601) planned for v3.x, or is this expected to clear up server-side without a local patch, the way it did in #601?

Full `--doctor` output and log files available if useful.

---
Written by Claude Sonnet 5 via [Claude Code](https://claude.ai/code)

aaddrick:
## What

Un-gates Cowork's bundled virtiofsd so hosts with a working KVM stack stop getting the "Cowork requires QEMU" prompt for packages they already have. Also fixes `--doctor` to mirror the client's real probe, so it stops passing machines the app rejects.

Fixes #771. Addresses the shared root cause in #772 (Crostini has a further vsock wall past this fix — see caveats below).

## Why

The official Linux bundle resolves virtiofsd from exactly two absolute paths, `/usr/libexec/virtiofsd` then `/usr/bin/virtiofsd`, checked with `fs.access` for read. No PATH search. It falls back to its own bundled copy at `resources/virtiofsd` only when `/etc/os-release` is `ID=ubuntu` with `VERSION_ID` starting `22.`. Anywhere else, `virtiofsdPath` stays null, the evaluator (`yukonSilver`) returns `virtualization_tools_missing`, and the UI tells the user to install QEMU packages that are already installed.

This breaks on distros with complete KVM stacks:

- **Arch / CachyOS** ship virtiofsd at `/usr/lib/virtiofsd` — not probed (#771).
- **Debian Bookworm** ships it at `/usr/lib/qemu/virtiofsd` — not probed (#772).
- **Pop!_OS 22.04** reports `ID=pop`, so a jammy base never qualifies for the bundled fallback. This reproduced on Anthropic's own official `.deb` from the official apt repo, which proves it's an upstream defect, not our repackaging.

The Linux VM bundle is fully built — the static manifest carries `unix/x64` and `unix/arm64` rootfs entries with checksums, and every install ships a working bundled virtiofsd (v1.13.2). Linux Cowork is gated off by the path resolution alone.

Two things made this hard to diagnose. The `startVM` path logs `VM not supported (linux/x64), skipping` for *any* unsupported status, so it reads like an architecture gate when it's the tools gate — that misled our triage bot into `unsupported_architecture`. And `cleanupVMBundleIfUnsupported` deletes the downloaded VM bundle whenever status ≠ supported, so affected users' bundles get repeatedly cleaned.

## How

**`scripts/patches/virtiofsd-probe.sh`** (new, wired into `active_patches`): un-gates the bundled fallback with one bounded regex rewrite anchored on the path-array literal — `return e||(A?bundled:null)` becomes `return e||bundled`. System paths stay preferred; the bundled copy becomes the universal last resort instead of a jammy-only one. Verified against the real minified bytes and the beautified copy, idempotent, and hard-fails the build on anchor miss.

I deliberately did **not** widen the probe array. `/usr/lib/qemu/virtiofsd` on qemu < 8 hosts (e.g. Ubuntu 22.04 jammy) can be the legacy C virtiofsd, whose CLI is incompatible with the Rust one. That's presumably why upstream's list is conservative. Un-gating the bundled copy avoids that trap entirely.

**`--doctor` fix:** the old check searched the broad distro path list and PASSed machines the app rejects — the exact false confidence in #771's report (doctor all-green, app says install QEMU). It now mirrors the client's probe order: the two system paths, then the bundled copy. A binary found anywhere else WARNs with the symlink one-liner instead of passing silently.

This clears the D-002 patch-zero bar. It compensates a genuine Linux-environment gap (Cowork unusable on non-Ubuntu-22 hosts with working KVM), the gap is reproduced on the official `.deb`, the patch is a bounded rewrite justified against official bytes, and it's filed upstream at [anthropics/claude-code#74605](https://github.com/anthropics/claude-code/issues/74605) so the downstream patch retires when the fix lands.

## Testing

- 5 new bats tests for the doctor probe: client-probed path PASS, bundled-copy PASS, non-probed path WARN + symlink hint, nothing-found WARN + package hint, and read-only access suffices (R_OK, matching the client).
- Patch verification against minified bytes, beautified copy, idempotency (re-run is a no-op), and negative (anchor miss hard-fails the build).
- **Live differential on the VM fleet's Arch guest** (Hyprland, virtiofsd only at `/usr/lib/virtiofsd`, qemu + OVMF-at-probed-path + `/dev/kvm` + `/dev/vhost-vsock` all present): the unpatched 1.18286.0 build logs `[cleanupVMBundleIfUnsupported] yukonSilver not supported (status=unsupported)` and deletes the VM bundle; the patched build on the same machine and session state takes the supported path instead (`[Bundle:status] rootfs.img missing` — bundle management, pending a logged-in session for the download). `--doctor` on that guest now reports `virtiofsd: bundled copy (…/resources/virtiofsd)` where it previously false-PASSed on `/usr/lib/virtiofsd`.

## Docs

- Upstream report draft at `docs/upstream-reports/771-cowork-virtiofsd-probe.md`.
- CHANGELOG entries under Fixed.

## Caveats for #772 (Crostini)

Same root cause, but Crostini hits a harder wall right after virtiofsd resolves: `/dev/vhost-vsock` (`vsock_module_missing`) and working KVM. The ChromeOS container typically can't `modprobe vhost_vsock`, so Cowork may stay unavailable there even after this fix. This PR fixes the misleading error message for them; the vsock gate is a platform limit we can't patch around. Tracking the doctor/logging follow-up in #772.

---
95% AI / 5% Human
Claude: root-cause analysis of the official bundle, patch + doctor implementation, tests, upstream report, fleet verification
Human: direction, scope decisions (file upstream + patch + doctor + reply), review


aaddrick:
@sabiut, merging. Please follow up if I missed something

sabiut:
Post-merge verification all clear.

- Full bats suite locally: 179/179, shellcheck clean.
- Exercised the patch against the pinned official 1.18286.0 bundle: the probe-array anchor is unique, the applied rewrite is byte-for-byte exactly the one advertised change, re-run is a clean no-op, and a simulated upstream reshape hard-fails the build as designed.
- Confirmed the R_OK (system paths) / X_OK (bundled) probe semantics in the minified bytes — doctor's `-r`/`-x` mirrors are faithful — and the bundled virtiofsd ships 0755 in the official data.tar.

Two small follow-ups in my lane I'll pick up separately: coverage for the `_cowork_incomplete` flag (the `run` subshell makes the current assignments dead code) and an artifact-level exec-bit assert on `resources/virtiofsd` so a repack mode-loss can't silently kill the bundled fallback.

sabiut:
The two follow-ups from the verification above are up in #774.

PR #774, opened 2026-07-06
test(#771): artifact exec-bit assert for bundled virtiofsd + _cowork_incomplete coverage

sabiut:
## What

The two test follow-ups promised in the [#773 post-merge review](https://github.com/aaddrick/claude-desktop-debian/pull/773#issuecomment-4888320969).

## Why

**Artifact exec-bit tripwire.** The #771 un-gate makes `resources/virtiofsd` the universal fallback, and the client resolves it with `X_OK`. A repack that drops the exec bit (the mode-loss trap in `docs/learnings/packaging-permissions.md`) would silently kill Cowork on every host without a client-probed system virtiofsd — with all artifact tests green. `validate_app_contents` now asserts the bundled copy is present and executable, covering deb, rpm, and AppImage through the shared helper.

**`_cowork_incomplete` coverage.** The five #773 doctor tests set `_cowork_incomplete=false` before `run`, but `run` executes in a subshell, so those assignments were dead code and the WARN branches' flag flip (which feeds the Cowork readiness summary) was never asserted. The dead assignments are removed; new tests call the helper directly (output to a file) and assert the flag for both WARN branches plus a PASS-leaves-false guard. Also added: client-path-over-bundled precedence, and a mode-stripped bundled copy falling through to WARN instead of PASSing.

## Testing

- doctor.bats 75 → 80; full suite 184/0. Shellcheck clean.
- Both new guards are mutation-tested: the artifact assert was run against the real official 1.18286.0 resources tree (PASS), then with the exec bit stripped and with the file removed (both FAIL); removing `_cowork_incomplete=true` from the helper fails exactly the two flag tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ The format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) — 
 
 <!-- Updated automatically by check-claude-version; will be current at release time. -->
 
+### Changed
+
+- The artifact tests now assert that Cowork's bundled `resources/virtiofsd` is present and executable in every package format: the [#771](https://github.com/aaddrick/claude-desktop-debian/issues/771) un-gate makes it the universal fallback and the client resolves it with `X_OK`, so a repack that drops the exec bit would silently kill Cowork on hosts without a client-probed system virtiofsd. The doctor's virtiofsd probe tests also grew coverage for the `_cowork_incomplete` readiness flag (the WARN branches were previously unasserted — `run` subshells discard the mutation), the client-path-over-bundled precedence, and the mode-stripped bundled copy falling through to WARN. ([#774](https://github.com/aaddrick/claude-desktop-debian/pull/774))
+
 ## [v3.0.1] — 2026-07-05
 
 ### Added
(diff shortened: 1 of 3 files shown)


aaddrick (approved):
Approving — clean, precise test PR.

- Confirmed the four removed `_cowork_incomplete=false` lines were genuinely dead: `run` subshells the helper, so the mutation never reached the test shell. The new direct-call tests capture it correctly (`_pass`/`_warn` echo to stdout, non-`run` keeps the flag visible).
- The mode-stripped bundled test matches the helper: bundled resolution uses `-x` while client paths use `-r`, so a 644 file correctly falls through to the `not found` WARN.
- The artifact exec-bit assert mirrors the client's `X_OK` resolution.

One thing worth a sanity check before merge (not blocking): `validate_app_contents` runs on every arch, and the new assert hard-`fail`s if `resources/virtiofsd` is absent. Mutation testing was against the amd64 1.18286.0 tree — if the official **arm64** deb doesn't bundle virtiofsd, this would break the arm64 leg. If arm64 does bundle it, this is exactly the tripwire intended.

---

PR #775, opened 2026-07-06
fix(#772): --version launcher fast-path; finish the -unofficial rename

aaddrick:
Fixes the silent `--version` reported in #772, plus a repo-wide rename audit for leftovers from the v3.0.0 package rename.

## `--version` printed nothing

The launcher only intercepted `--doctor`. Everything else, including `--version`, fell through to the full launch path. And `run_electron_and_cleanup` redirects all Electron stdout/stderr into `~/.cache/claude-desktop-debian/launcher.log`:

```sh
"$@" >> "$log_file" 2>&1 &
```

So Electron did print its version. It just went to the log file, not the terminal. The user saw nothing.

The fix adds a `--version` fast-path to all three launcher generators (`scripts/packaging/deb.sh`, `rpm.sh`, `appimage.sh`). It prints `<package_name> <version>` and exits 0 before any of the launch/redirect logic runs. The value is baked in at build time, so it reads something like:

```
claude-desktop-unofficial 1.18286.0-3.0.1
```

The AppImage path needs one extra step. Its AppRun heredoc is single-quoted, so shell expansion doesn't happen at generation time. I used the same placeholder-then-sed pattern the existing `WM_CLASS` substitution uses: the heredoc emits `@@PACKAGE_NAME@@` and `@@VERSION@@`, and sed stamps the real values after generation.

### Verification

- **AppImage**: end-to-end via `./build.sh --build appimage --clean no`. The generated AppRun prints `claude-desktop-unofficial 1.18286.0` and exits 0. No `-3.0.1` suffix locally because local builds have no release tag; CI passes the tag.
- **deb/rpm**: harness-verified only. I confirmed the heredoc expansion produces the right fast-path with a generation harness. I did not install a real deb or rpm and run the flag, so treat those two as static-verified until someone confirms on a real install.
- Shellcheck clean on all changed scripts.

## Rename audit (post-v3.0.0)

The v3.0.0 rebase renamed the package to `claude-desktop-unofficial`, and a few references to the old `claude-desktop` name were still around. This PR sweeps them.

**`scripts/doctor.sh`** — the hardcoded chrome-sandbox default still probed `/usr/lib/claude-desktop/chrome-sandbox` (the official tree) instead of `/usr/lib/claude-desktop-unofficial/chrome-sandbox`. That's a real functional bug for anyone on our package. Also fixed two "Reinstall claude-desktop package" hints and two stale comments.

**Docs telling users to run `claude-desktop`**: quickstart in `docs/index.md`, five commands in `docs/configuration.md`, `docs/testing/runbook.md` (which also had stale CI artifact names like `claude-desktop-deb-amd64` → `package-amd64-deb`), `docs/testing/cases/launch.md`, `docs/testing/cases/distribution.md`, `docs/issue-triage/README.md` (mirrors the already-renamed `bug_report.yml`), `docs/testing/quick-entry-closeout.md` (old asar path), and `docs/learnings/packaging-permissions.md` (old install-tree paths).

While in `distribution.md` S05 and `launch.md` T13, I brought the doctor descriptions current. They still described the pre-#711 behavior. Doctor now probes rpm ownership of the binary first, then falls back to `dpkg-query claude-desktop-unofficial`.

### Deliberately left alone

Not everything named `claude-desktop` is a leftover. These are correct as-is:

- References to Anthropic's official package and install tree.
- The upstream ELF/process name `claude-desktop` inside our install dir.
- Icon source basenames.
- The transitional dummy package and the `Conflicts`/`Replaces` stanzas.
- The AUR package name.
- Nix attrs (owner @typedrat).

## CHANGELOG

`[Unreleased]` gets an Added bullet for `--version` and a Fixed bullet for the rename leftovers.

---
95% AI / 5% Human
Claude: root-cause trace of the silent --version, launcher fast-path across deb/rpm/AppImage, rename audit and doc updates, local AppImage build verification
Human: scoped the task, directed the rename audit, reviewed


sabiut (approved):
Verified locally: full bats suite green on dd22d01, shellcheck clean on all four changed scripts.

Traced the fast-path in all three generators — placement before the log redirect is correct, `$version`/`$package_name` are in scope in each, and the AppImage placeholder-sed matches the existing WM_CLASS precedent. The rpm leg printing the raw hyphenated version (rather than the `%{version}`-split `rpm_version`) keeps output format-consistent across deb/rpm/AppImage, which reads as the right call.

Also built the AppImage from this branch end-to-end and ran it: `./claude-desktop-unofficial-1.18286.0-amd64.AppImage --version` prints `claude-desktop-unofficial 1.18286.0` and exits 0 — independently reproduces your result.

The doctor chrome-sandbox path fix is a real bug fix — a review sweep I ran after the v3.0.0 merge flagged that exact line as a false-PASS risk independently. Heads-up that it overlaps with #745 and my held chrome-sandbox branch; whichever lands second needs a trivial rebase there.

One follow-up I'll take in my lane: a `--version` assert in the artifact tests, so the deb/rpm legs stop being static-verified-only — the fast-path exits before any launch logic, so it's a cheap assert on all three formats.

Approving.

PR #776, opened 2026-07-06
feat(#772): opt-in bubblewrap Cowork backend for KVM-less hosts

### LINKED ISSUE #772: [bug]: Cowork fails on Crostini (Debian Bookworm 12.14) despite all dependencies present and verified

aaddrick:
## Summary

This adds an opt-in bubblewrap Cowork backend for hosts where KVM isn't reachable. It's the fallback I promised @Blackhawke on #772. Set `COWORK_VM_BACKEND=bwrap` and Cowork runs Claude Code in a namespace sandbox on the host instead of a microVM.

This does **not** fully fix #772. Crostini blocks `vhost_vsock` at the Termina kernel level, so the official KVM gate can never pass there, and no amount of "install QEMU" fixes a kernel-level block. What it does is give #772's Crostini users a path that doesn't need vsock at all. The trade is real: bwrap is weaker isolation than a VM. That's the cost of running where KVM can't.

## The problem

The official client runs Cowork in a KVM microVM, gated on `/dev/kvm` + `/dev/vhost-vsock`. It drives QEMU through a bundled native helper (`cowork-linux-helper -socket <path>`) over a length-prefixed-JSON Unix socket. @Blackhawke confirmed in #772 that ChromeOS Crostini blocks `vhost_vsock` in the Termina kernel, so that gate is a hard wall on that platform.

## What this does

It reinstates the pre-3.0.0 bubblewrap backend as an opt-in path. Three flag-gated injections in a new asar patch (`patch_cowork_bwrap`):

1. Report the `yukonSilver` KVM support evaluator as supported.
2. Swap the native-helper spawn for `node cowork-vm-service.js -socket <path>`.
3. Suppress the unused multi-GB VM-image download (foreground fetch + warm prefetch).

Every branch is gated on `process.platform === "linux" && process.env.COWORK_VM_BACKEND === "bwrap"`. On an unflagged launch every branch evaluates false and the official KVM path runs unchanged. Per [D-002](docs/decisions.md) this clears the patch bar as an opt-in path compensating a genuine Linux-environment gap, same category as `virtiofsd-probe` and `org-plugins`.

The daemon (`scripts/cowork-fallback/cowork-vm-service.js`, the old 2.x daemon) was adapted to the official helper's socket protocol: honors the `-socket` argv, id-multiplexed non-blocking dispatch, and the new methods the official bundle expects (`sendGuestResponse`, `getSessionsDiskInfo`, `deleteSessionDirs`, `pruneSessionCaches`, `getNetworkDrives`). The wire contract I pulled out of the official bundle is written up in `scripts/cowork-fallback/PROTOCOL.md`.

## Runtime node dependency (flagged path only)

Worth calling out directly. The official Electron binary ships with the `RunAsNode` fuse OFF (verified empirically), so it can't run the daemon the way 2.x did via `child_process.fork()`. The daemon runs under a system `node`/`nodejs` instead. The launcher auto-detects it and exports `COWORK_NODE_PATH`; `--doctor` flags its absence. This adds a runtime dependency, but only on the flagged path. On Crostini `apt install nodejs` covers it.

I considered two alternatives and rejected both:

- **Flip the `RunAsNode` fuse.** Changes the official binary's security posture for every user, not just the opt-in ones.
- **Bundle a node binary.** Adds ~50-90MB per package.

The daemon ships in `resources/` (`process.resourcesPath`), staged by `patch_app_asar`. It has to live outside the asar (`child_process` can't exec from inside one) and outside `app.asar.unpacked` (keeps the repack invariant pinned to upstream's set).

## What I verified

- Local AppImage build applies all four injections. They survive the asar repack and the repacked bundle parses (`node --check`).
- Daemon passes all 146 existing bats tests plus a new 15-check length-prefixed-protocol smoke test: configure, id-multiplexed pipe, `subscribeEvents` ack-then-events, `startVM`/`isGuestConnected`, spawn with stdout/exit events, the new methods, and unknown-method rejection.
- A "join" test runs the exact spawn expression extracted from the built asar. Flagged, it spawns `node` plus the shipped daemon, which binds the socket and answers RPC. Unflagged, it spawns the native-helper path and no daemon. That proves the runtime join without the GUI.

## What I did NOT verify

I have not driven a full GUI Cowork session end-to-end on a real KVM-less host. The headless env here can't drive the Cowork tab plus login. So "a real Cowork session actually runs under bwrap on Crostini" is static- and harness-verified, not live-verified. This needs a real-hardware confirmation before we can call it done, ideally from the #772 reporter.

## Ownership

Cowork is your subsystem, @RayCharlizard, and this parked fallback was originally staged for your 3.1 investigation, so this lands squarely in your area. I'd like your review on the protocol adaptation and the flag gating. If you or anyone on KVM-less hardware can drive a real Cowork session under the flag, that closes the one verification gap I couldn't reach from here.

---
90% AI / 10% Human
Claude: bundle protocol mapping, daemon protocol adaptation, the flag-gated asar patch, packaging + launcher + doctor wiring, build/protocol/join verification
Human: scoped the task, made the system-node vs fuse-flip call, reviewed


sabiut (changes requested):
Reviewed from the testing/release-quality side; deferring the protocol adaptation and flag-gating review to @RayCharlizard as requested.

The structure is right — `_doctor_check_bwrap_node` follows the extracted-helper pattern, the load-bearing/cosmetic fail-mode split in the patch is exactly the virtiofsd precedent, and staging the daemon outside both the asar and the unpacked set keeps the repack invariant clean.

Verified locally on ff1b26a: full suite plus all three cowork-fallback suites pass (325/325), shellcheck clean. Also independently verified the patch against the pinned official 1.18286.0 bundle: all four injections apply, the patched bundle parses, re-run is a clean no-op, and a broken load-bearing anchor fails the build with the file left byte-for-byte untouched. The spawn anchor is unique in the current bundle.

Four asks before this lands, three of them cheap:

1. **The claimed verification isn't committed.** The PR body cites a "new 15-check length-prefixed-protocol smoke test" and a "join" test that runs the extracted spawn expression — neither is in the diff (`scripts/cowork-fallback/` gains only PROTOCOL.md, README.md, and the daemon changes). Those two tests are precisely the regression guards this revived subsystem needs; please commit them.

2. **Wire the cowork-fallback suites into CI.** The post-v3.0.0 accepted-risk for these suites running nowhere was explicitly "reference code, ships in no artifact." This PR ships the daemon in every package and makes it user-reachable, so that condition no longer holds — one line in tests.yml (`bats scripts/cowork-fallback/tests/`) plus the new protocol smoke closes it.

3. **Node version floor.** The daemon uses optional chaining (line 1409) and `??` (line 2611), so it needs Node ≥ 14 — but `_doctor_check_bwrap_node` PASSes any executable node and the launcher exports whatever it finds. `sudo apt install nodejs` on Ubuntu 22.04 yields Node 12: doctor green, launcher happy, daemon dies with SyntaxError at first Cowork use. Crostini bookworm (Node 18) is fine, but the check should enforce a major-version floor in both the doctor helper and `setup_cowork_bwrap_env`.

4. **Doctrine question, not a blocker:** the doctor's Cowork rule has been "absence is never a `_fail`." The new helper `_fail`s on missing node/daemon. Defensible since the user explicitly opted in — just confirm that's the intent, since it flips the doctor exit code for flagged users.

Nit: Patch B swaps the first regex match without asserting uniqueness the way virtiofsd-probe's probe-array check does. Single-site today (verified against the real bundle), but an exactly-1 assertion would fail loud instead of half-swapping if upstream ever duplicates the spawn shape.

I'll take the coverage side in my lane as follow-ups regardless: doctor.bats tests for `_doctor_check_bwrap_node` (already testable — resources_dir arg + `COWORK_NODE_PATH` env), launcher-common.bats for `setup_cowork_bwrap_env`, and a `resources/cowork-vm-service.js` presence assert in the artifact tests mirroring the virtiofsd one from #774.

sabiut (approved):
All four asks addressed, and the revision goes further than asked — re-verified locally on 4233478:

- **Tests committed and passing**: full suite + cowork-fallback suites now 333/333, and `protocol-smoke.js` passes standalone. The new `cowork-bwrap-patch.bats` covers the join semantics, the fail-untouched negative, *and* an exactly-1 duplicated-anchor guard — that's the full verification methodology as regression tests.
- **CI wired** with the exemption rationale documented in the workflow comment.
- **Node floor done right**: v18 for `fs.statfsSync` rather than the bare v14 syntax floor, enforced in all three layers (doctor, launcher log, daemon startup guard with an actionable log line + exit 1), kept in lock-step by name.
- **warn-not-fail** adopted with the doctrine rationale written into the helper comment, `_cowork_incomplete` still set.
- **Uniqueness nit**: adopted on both load-bearing anchors, fail-loud.

Also re-ran the updated patch against the pinned official 1.18286.0 bytes: all four injections apply, the bundle parses, re-run is a clean no-op.

One non-blocking note for a future pass: `fs.statfsSync` landed in **18.15**, and all three guards compare major only — a Node 18.0–18.14 runtime (e.g. Ubuntu 23.04's 18.13) passes every check but still lacks the call. Since the daemon already self-guards, the cheapest fix is feature-detection there (`typeof fs.statfsSync === 'function'`) instead of a version compare; doctor/launcher majors can stay as the human-friendly signal. Fine as a follow-up, not worth blocking on.

The GUI-session-on-real-hardware gap remains open as the PR states, and the protocol/flag-gating review stays with @RayCharlizard. From the testing/release-quality side: approving.

sabiut (changes requested):
The statfsSync feature-detection is exactly right — one capability probe shared verbatim across daemon, doctor, and launcher, with the 16.19 backport caught. No notes there.

The config-file flag is the right mechanism for GUI launches (allowlist, env-wins precedence, never executed as shell — good design), but this revision has two real defects and a coverage gap. Verified on 0ebdeb2 (suite 333/333, shellcheck clean — which is part of the problem, see 3):

1. **`CLAUDE_GTK_IM_MODULE` is silently un-allowlisted.** The `allowlist` string uses a line continuation *inside single quotes*, where `\` + newline are literal characters. Verified end-to-end with a fixture config through the real function: `CLAUDE_GTK_IM_MODULE=xim` logs "ignoring unrecognized key" — the string contains `…PASSWORD_STORE \<newline>CLAUDE_GTK_IM_MODULE…`, so the `*" $key "*` match fails for the key preceded by a newline instead of a space. Every other key happens to survive. Fix: build the allowlist without the continuation (single line, or `+=`).

2. **`--doctor` never loads the config file.** The generated launchers intercept `--doctor` before `setup_electron_env`, and doctor.sh never calls `load_launcher_config` — so a user with `COWORK_VM_BACKEND=bwrap` *only* in the config file (the exact persona this feature serves) gets zero bwrap diagnostics. That's the doctor-vs-launch disagreement class from #771 again: doctor must see the same environment the launch will. Note `load_launcher_config` currently assumes `setup_logging` ran (for `log_message`), so the doctor call site needs that handled.

3. **No tests for `load_launcher_config`.** It's new user-facing parsing with non-trivial semantics and the suite passes with bug 1 present — which is the demonstration that coverage is needed. My fixture harness confirmed the parts that work (LD_PRELOAD correctly rejected, env-wins precedence honored, quote-strip correct), so the tests just need to lock those in alongside the fixes. The function is cleanly testable (point `XDG_CONFIG_HOME` at a fixture). Happy to contribute the bats if you'd rather not grow this PR further.

Minor: values aren't whitespace-trimmed — `CLAUDE_DISABLE_GPU = 1` (spaces around `=`) exports `" 1"` (verified live), the consuming comparison fails, and the log line still reports it as set, which will be a confusing report someday.

Re-requesting changes on 1 and 2; 3 can be this PR or an immediate follow-up.

sabiut:
Post-merge verification of ff7dca2 all clear — the re-review is fully addressed, so consider the standing CHANGES_REQUESTED lifted.

- Allowlist fix verified end-to-end through the real function: `CLAUDE_GTK_IM_MODULE=xim` now round-trips from the config file (previously silently rejected), and `LD_PRELOAD` rejection + env-wins precedence still hold.
- Doctor/config parity verified live: with `COWORK_VM_BACKEND=bwrap` set *only* in `~/.config/claude-desktop-debian/environment`, `run_doctor` now emits the full bwrap diagnostics section (node capability PASS included).
- Value trimming confirmed (`CLAUDE_DISABLE_GPU = 1` exports `1`, not `" 1"`).
- The 9 new `load_launcher_config` bats tests cover exactly the semantics that were at risk — the round-trip test is the regression guard the allowlist bug needed. Full suite 349/349 locally on e7673c9.
- Nice catches beyond the review: the `log_message` `$1`→`$*` fix (wrapped multi-word calls were logging only their first fragment) and the function-replacement guard against the $-in-identifier substitution trap in the patch.

Remaining coverage I'll pick up in my lane as promised: doctor.bats for `_doctor_check_bwrap_node`, launcher-common.bats for `setup_cowork_bwrap_env`, and an artifact assert for `resources/cowork-vm-service.js` mirroring the #774 virtiofsd one (the "gone by design" comment at test-artifact-common.sh:82 is stale now too).

PR #781, opened 2026-07-10
test(#772,#775): bwrap runtime coverage + --version and daemon artifact asserts

sabiut:
## What

The coverage follow-ups promised in my [#775 approval](https://github.com/aaddrick/claude-desktop-debian/pull/775) and [#776 reviews](https://github.com/aaddrick/claude-desktop-debian/pull/776), combined into one tests-only PR.

## Why

**`--version` fast-path asserted end-to-end on all three formats.** #775's deb/rpm legs were static-verified only. The fast-path exits before any launch, log-redirect, or sandbox logic, so a shared `run_version_flag_test` helper can run the real launcher directly — no display, D-Bus, or privilege handling. Deb exact-matches the control `Version` (the same string `deb.sh` baked into the echo); rpm prefix-matches `%{VERSION}` since the launcher bakes the raw hyphenated build version that rpm splits into `%{VERSION}-%{RELEASE}`; AppImage matches the filename-derived version (both local and CI-tagged name shapes verified).

**Bwrap daemon presence asserted in every artifact.** `resources/cowork-vm-service.js` ships since #776 and an opt-in launch fails at spawn without it; presence (not exec bit — it's spawned via system node) is now checked in `validate_app_contents`, next to the #774 virtiofsd assert. Also fixes the stale "cowork-vm-service.js gone by design" comment from the 2.x cleanup.

**Direct coverage for the bwrap runtime plumbing.** 9 doctor tests for `cowork_node_has_features` / `_doctor_check_bwrap_node` (capability probe against a real node and a versioned-but-featureless stub, all three WARN branches, and the `_cowork_incomplete` flip via direct call — `run` subshells discard it) and 5 launcher tests for `setup_cowork_bwrap_env` (flag gating, explicit-path precedence, PATH auto-detect, no-node and capability-warning log lines).

## Testing

- Full suite 349 → 363, all green; shellcheck clean on the artifact scripts.
- Both new artifact guards mutation-tested against real artifacts: the daemon assert FAILs on the pristine official 1.18286.0 resources tree and PASSes once the file is staged; `run_version_flag_test` PASSes against a real locally-built AppImage (`claude-desktop-unofficial 1.18286.0`, exit 0) and FAILs on a wrong expected prefix.
- Filename→version derivation verified for both `…-1.18286.0-amd64` and `…-1.18286.0-3.1.0-arm64` shapes.


sabiut:
The BATS failure on 5cf6426 was a pre-existing flake, not this PR's changes: the #693 "live UI present" reaper test spawns its stand-in UI via `exec -a` and races the reaper against the child's fork→exec window — on the loaded runner the `--class` fingerprint wasn't in `/proc/<pid>/cmdline` yet, the reaper took the orphan path, and the direct call died on the `((_wait++))` errexit trap (`launcher-common.sh:401`). The identical test passed on three sibling branches within 30 minutes.

90db93f fixes it at the source: a bounded poll for the fingerprint to land before invoking the reaper. Stress-verified 20/20 locally.

aaddrick:
One line in the flake fix blocks merge — covered below — and everything else here is merge-quality. I ran the suite green from the PR head, and the coverage is real. The launcher and doctor tests call the actual merged functions (`setup_cowork_bwrap_env` at launcher-common.sh:828, `cowork_node_has_features` / `_doctor_check_bwrap_node` at doctor.sh:1101/1120), and every asserted log and hint string matches source byte-for-byte. The `--version` asserts hit the real fast-paths in all three packaging scripts. I checked the rpm prefix-match against rpm.sh's first-hyphen version split (rpm.sh:18-19 vs the raw `$version` baked at :104), so `%{VERSION}` is always a prefix and the assert holds. The daemon assert points exactly where app-asar.sh:191-193 stages `cowork-vm-service.js` under the resources/ root, and the `app.asar.unpacked` comment fix matches reality. CI wiring checks out too: test-artifacts.yml runs all three formats on both arches with node present, so nothing silently skips.

One thing blocks merge, and it's the flake-fix poll in 90db93f. The poll greps the tr'd cmdline for `--class=Claude` without the trailing space. The reaper's fingerprint it's guarding (`_claude_desktop_ui_cmdline_matches`, launcher-common.sh:337) requires `--class=Claude ` with the space. During bash startup, before `exec -a` lands, /proc cmdline reads `bash -c exec -a "--class=Claude" sleep 300`. The loose pattern matches inside the quotes, the reaper's strict one doesn't. So the poll can green-light the reaper while the reaper still can't see the child.

I dug into how often that window is live. On real hardware that pre-exec state is basically the whole race: 498/500 instantaneous samples post-spawn, with exec landing in ~3-4ms. Natural scheduling never tripped the mismatch here (0/700 across idle and loaded runs), which is exactly why it survives local stress-runs. But freezing the child in that state, as a stand-in for vCPU steal on a loaded CI runner, reproduced the miss 5/5: poll breaks, reaper takes the orphan path, same `((_wait++))` death at launcher-common.sh:401 you set out to fix. The catch is the two events are correlated. A starvation long enough to cause the original flake is the same condition that parks the poll in the lying window. So the fix closes the queue-latency mode but might be a coin flip in the regime that actually flaked.

The clean fix leans on what setup() already gives you: it sources launcher-common.sh with WM_CLASS substituted, so the poll can just call the reaper's own predicate instead of keeping a parallel pattern. `_claude_desktop_ui_cmdline_matches "$(tr '\0' ' ' < "/proc/$ui_pid/cmdline" 2>/dev/null)"` inside the `if` is errexit-safe there, and pattern drift becomes impossible by construction. I verified the strict predicate correctly waits through the pre-exec state. If you'd rather stay minimal, just add the trailing space to the grep. Either way, put an explicit failure after the 5s timeout. With ~4ms typical latency a silent fall-through would reproduce the exact confusing flake signature this poll exists to kill. And don't touch `((_wait++))` in launcher-common.sh: production has no `set -e`, so it's a non-defect there, and it wouldn't fix this flake anyway since a fingerprint miss fails the kills-file assert regardless.

Non-gating, whenever you're next in the branch. The CHANGELOG entry is one mega-bullet bundling three changes. Repo rule is one bullet per change, and tests-only work arguably belongs under Added rather than Changed. Two cheap hardenings if you're already touching it: a `[[ -n ... ]]` guard so an empty dpkg-deb/rpm-qp result can't make the version prefix-match vacuous, and point the COWORK_NODE_PATH precedence test at a symlinked node in `$TEST_TMP` so it proves precedence instead of just the log line. Fix the poll and this is good to go.

---

sabiut:
All landed in 2aa6f3c — took your preferred fix plus every non-gating item:

- **Poll drift closed by construction** — the poll now calls `_claude_desktop_ui_cmdline_matches` on the tr'd `/proc/$ui_pid/cmdline`, exactly as you sketched, so it can only green-light states the reaper itself will see. And it fails loudly after the 5s ceiling with the PID in the message, so a starved runner produces a named failure instead of the original flake signature. Stressed the test 20/20 locally at the new head.
- **Empty-metadata guard** — `run_version_flag_test` now rejects an expected prefix with no version component (empty `dpkg-deb -f` / `rpm -qp` result), exercised both ways: empty version FAILs with the new message, happy path unchanged.
- **Precedence proven** — the explicit-`COWORK_NODE_PATH` test pins a symlink in `$TEST_TMP` (distinguishable from what the PATH probe would find) and asserts the variable survives the call, not just the log line.
- **CHANGELOG** split one-bullet-per-change and moved to Added.

Suites at head: launcher-common 97/97, doctor 89/89. Left `((_wait++))` alone per your note. Like the rest of the campaign, self-merge waits on #792 making CI green.

PR #782, opened 2026-07-10
refactor(doctor): extract AppArmor userns check into a testable helper

sabiut:
Extracts run_doctor's inline user-namespace/AppArmor check into `_doctor_check_userns_apparmor` with four path hooks (`_DOCTOR_USERNS_PATH`, `_DOCTOR_DEB_ELECTRON`, `_DOCTOR_AA_PROFILE`, `_DOCTOR_AA_LOADED`; defaults are the real system paths) and 7 bats tests covering both silent gates, the loaded PASS, both WARN variants, and the unreadable-set INFO-never-PASS guard. Pure behavior-preserving refactor — doctor output verified byte-identical. Continues the run_doctor coverage campaign (#740/#744/#745).

**Sequencing note:** this conflicts with #744 and #745 in `doctor.sh` — all three insert their helper at the same anchor after `_doctor_check_bwrap_fallback()` (the bats side auto-merges). Resolutions are trivial both-sides-insert; landing order and rebases per the review discussion below. #740 merges clean against everything.


aaddrick:
The refactor checks out. I normalized the inline block on main (`scripts/doctor.sh:1326-1394`) against the helper body and the only difference is the four `_DOCTOR_*` hook indirections. Every condition, message, grep anchor, EUID split, and the empty-read-vs-mode-bits securityfs logic is string-identical, and the hook defaults equal the old literals. Full `doctor.bats` at PR head is 87/87 green. The hook convention matches the existing `_DOCTOR_KVM_DEV`/`_DOCTOR_VSOCK_DEV`/`_DOCTOR_OVMF_PATHS` precedent, set-but-empty semantics included.

One request before merge: the tests only half-pin the grep anchor. `grep -q '^claude-desktop '` fails the PASS test, but every permissive weakening survives all 7 tests. Dropping `-unofficial`, dropping the `^`, dropping the trailing space, all pass, because the not-loaded WARN fixture's loaded set is just `firefox (enforce)` with no near-miss line. The co-install state the anchor exists to disambiguate (official `claude-desktop` profile loaded, ours absent) is a real machine state per `deb.sh` and the official postinst, and a simplified grep false-PASSes exactly there. One-line fix: add `claude-desktop (unconfined)` to the WARN fixture's loaded set and keep asserting WARN. Since testability is this PR's stated purpose, worth doing here.

The body's "Independent of #744/#745 (disjoint blocks and test anchors)" is textually wrong. Trial merges show all three insert their helper at the same anchor, right after `_doctor_check_bwrap_fallback()`, so 782×744 and 782×745 both conflict in `doctor.sh` (the bats side auto-merges). The resolutions are trivial both-sides-insert, but the description should say so. All three are yours anyway, so pick your own landing order and rebase between them, or I can resolve the keep-both hunks at merge time — whichever costs you less. #740 merges clean against everything, in every order. Same-anchor sequencing lives here; the notes on #744/#745 point back to this comment.

Two optional notes, neither a blocker. Adding the four new hook unsets to `setup()`, mirroring the existing `_DOCTOR_SECRET_BACKEND` unset, would guard against an invoking shell exporting `_DOCTOR_*`. It's two lines, and bats process isolation already covers cross-test leakage since every test assigns the hooks its branch reads, so it's belt-and-suspenders. And for the CHANGELOG: rather than per-PR bullets, one Changed bullet covering the whole doctor-coverage campaign, added by whichever of the four merges last, matches the #774 precedent — treat that as the plan unless you object.

---

aaddrick:
@sabiut these are all yours to land. You've got write access, so once the ask on each PR is in and CI's green, self-merge them. No need to wait on another review round from me. The per-PR asks, compressed: #740 after the rebase plus the two comment fixes; #744 with the `_doctor_failures` assert; #745 with the two test asks; #782 with the fixture line and the description correction; #781 once the poll calls the reaper's predicate.

Landing order for the three same-anchor extractions stays your call per the sequencing note above. If you'd rather not juggle it, the offer stands to resolve the keep-both hunks at merge time instead. Either way works.

---

sabiut:
Both landed in f407213:

- **Near-miss fixture** — both WARN fixtures' loaded sets now carry `claude-desktop (unconfined)` (the official package's profile, i.e. the co-install state the anchor disambiguates). Mutation check: dropping `-unofficial` from the grep now fails 3 tests (the PASS test via the no-longer-matching strict line, both WARN tests via the near-miss) instead of surviving all 7. Suite green at 87/87. Also took the optional: the four `_DOCTOR_*` path hooks are now unset in setup(), mirroring `_DOCTOR_SECRET_BACKEND`.
- **Description corrected** — the "independent of #744/#745" claim is replaced with the same-anchor sequencing note pointing back here.

Landing order call: #744 → #745 → #782, rebasing each onto main after the previous merge (I'll handle the keep-both hunks myself — thanks for the offer). #740 rides wherever since it merges clean, and per your CHANGELOG plan the single campaign bullet goes in with whichever of the four merges last. All of this waits on #792 (the AU-1 tripwire re-point) going green first, since CI is red repo-wide until then.

---

aaddrick:
Hey @sabiut! CI cleared. #792 merged, so the AU-1 tripwire gate you flagged is gone.

One thing worth flagging: your sequencing comment predates a second wall. #793 landed after it. Upstream 1.19367.0 code-split the main-process bundle, so every asar patch had to re-point at the content-hashed `index.chunk-<hash>.js` instead of `index.js`. Your campaign branches were red on that break too, not just the tripwire. Both fixes had to land, and all five branches have now picked up both. (#795 is a docs follow-on to it, a chunk census of the 44 satellites the split introduced. Just context, not a blocker.)

Net state: #744, #745, #782, #740, and #781 are all green. Every build leg passes, which is the proof they carry both fixes. So your "waiting on #792" is satisfied and then some. You're clear to run the order you called: #744 → #745 → #782, rebasing each onto main after the prior merge, #740 riding wherever, and the single campaign CHANGELOG bullet going in with whichever of the four merges last.

The offer to sort the keep-both hunks at merge time still stands if you'd rather hand them off, but you said you've got them, so your call.

While you're in there, #794 retargeted onto main clean now that #793 merged. Diff is tests-only, the coverage for `_resolve_main_js`. I ran it: 11/11 green, and I mutation-checked the two load-bearing cases. Neutering the multi-chunk guard reds test 9, neutering the missing-chunk check reds test 10, so it's genuinely pinned, not decoration. One heads-up: CI shows "no checks reported on the branch" because it predates the retarget and never ran in Actions. A trivial empty push triggers a green check, or you merge on the local green since the diff is inert test code. Your call as tests owner.

Last thing: #796, the test-methodology learnings doc mined from your PRs and reviews. Docs-only, green. I spot-checked the concrete citations against live code and they hold. Worth knowing the stacking though: #797 (the AGENTS/CLAUDE resync) sits directly on top of it, carrying #796's exact commit plus one resync commit. So #796 is the "separate cleanup" its own caveat deferred, and order matters here. Land #796 first, then #797 rebases down to just the resync. I'd suggest #796 → #797.

---

Issue #779, opened 2026-07-08
[bug]: StartupWMClass is incorrect and produces a second generic icon


BashBandito:
### Version (`claude-desktop-unofficial --doctor` output)

```shell
v3.0.1+claude1.18286.2
```

### What happened

The StartupWMClass value in the debian package .desktop file is incorrect. On first install and each update, it reverts back to the incorrect value which on GNOME DE creates a second taskbar icon that will not attach to the actual pinned icon.

The deb package ships with `StartupWMClass=Claude` but this is incorrect. When I manually change it to `StartupWMClass=claude-desktop `it looks proper.

### Steps to reproduce

Check /usr/share/applications/ .desktop file in text editor.

### Expected behavior

Use proper `StartupWMClass=claude-desktop` so the taskbar icon consolidates with pinned icon when app is launched.

### Logs / errors

```shell

```

jpetzke:
Same on Fedora 44 (KDE Plasma, Wayland session, app on XWayland), but KWin sees a third value:

```
resourceClass  s "com.anthropic.Claude"
resourceName   s "com.anthropic.claude"
desktopFile    s ""
```

Empty `desktopFile` = no match, so the pinned launcher and the window don't get grouped. Second taskbar entry on launch, generic `?` icon after closing.

`StartupWMClass=com.anthropic.Claude` as a user override fixes it.

Version: 1.19367.0-3.2.1.fc42

BashBandito:
> Same on Fedora 44 (KDE Plasma, Wayland session, app on XWayland), but KWin sees a third value:
> 
> ```
> resourceClass  s "com.anthropic.Claude"
> resourceName   s "com.anthropic.claude"
> desktopFile    s ""
> ```
> 
> Empty `desktopFile` = no match, so the pinned launcher and the window don't get grouped. Second taskbar entry on launch, generic `?` icon after closing.
> 
> `StartupWMClass=com.anthropic.Claude` as a user override fixes it.
> 
> Version: 1.19367.0-3.2.1.fc42

I'm glad you commented this because the previous value I used is no longer working but the one you shared works.

In other words, it changed in a recent update but not for the better.

aaddrick:
Thanks @jpetzke and @BashBandito — both datapoints reproduced live on the VM fleet today, on GNOME (Ubuntu 24.04) and KDE Plasma (Fedora 43), against 1.19367.0. `xprop` shows `WM_CLASS` = `com.anthropic.Claude` on both.

[...]

Issue #647, opened 2026-05-25
[bug]: Arch .desktop incorrect StartupWMClass


isc30:
### Version (`claude-desktop --doctor` output)

```shell
❯ sudo claude-desktop --doctor
Claude Desktop Diagnostics
================================

[PASS] Display server: X11 (DISPLAY=:0)
       Menu bar mode: auto (default, Alt toggles visibility)
       Titlebar style: hybrid (default, native frame + in-app topbar)
[PASS] Electron: v41.5.0 (/tmp/.mount_claude0rJkHy/usr/lib/node_modules/electron/dist/electron)
[FAIL] Chrome sandbox: perms=755, owner=root
       Fix: sudo chown root:root /tmp/.mount_claude0rJkHy/usr/lib/node_modules/electron/dist/chrome-sandbox
            sudo chmod 4755 /tmp/.mount_claude0rJkHy/usr/lib/node_modules/electron/dist/chrome-sandbox
[PASS] SingletonLock: no lock file (OK)
[PASS] Password store: basic
         → using fixed-key fallback; tokens are protected by filesystem permissions only
       MCP config: not found at /root/.config/Claude/claude_desktop_config.json (OK if not using MCP)
[PASS] Node.js: v22.22.2
       Path: /usr/bin/node
[PASS] Desktop entry: /usr/share/applications/claude-desktop.desktop

Cowork Mode
----------------
[PASS] bubblewrap: found
[PASS] bubblewrap: sandbox probe succeeded
[PASS] KVM: accessible
[PASS] vsock: module loaded
       QEMU: not found
[PASS] socat: found
       virtiofsd: not found
       VM image: not downloaded yet
       Cowork isolation: bubblewrap (namespace sandbox)
       Log file: not yet created (OK)

1 check(s) failed.
See above for fixes.
```

### What happened

The arch version of claude-desktop creates the `.desktop` with `StartupWMClass=claude-desktop` but the window uses `claude`.

### Steps to reproduce

Open `/usr/share/applications/claude-desktop.desktop`

### Expected behavior

Update `.desktop` to `StartupWMClass=claude` so that desktop integration works as expected

```diff
[Desktop Entry]
Type=Application
Name=Claude Desktop
Comment=Claude Desktop Application
Exec=/usr/bin/claude-desktop %U
Icon=claude-desktop
Terminal=false
Categories=Network;Utility;
MimeType=x-scheme-handler/claude;
-StartupWMClass=claude-desktop
[...]

PR #786, opened 2026-07-11
fix(desktop): StartupWMClass must be ELF basename, not productName

diarized:
## Problem

- Runtime `WM_CLASS` (res_name, res_class) = `claude-desktop` — Electron derives it from the ELF basename.
- Launcher passes `--class=Claude` — not honored by Electron.
- Shipped `.desktop` `StartupWMClass` = `Claude` (from `WM_CLASS='Claude'`, `build.sh`).
- `StartupWMClass` != runtime `WM_CLASS` → shell can't map window to `.desktop` entry.

### Effect

- GNOME window tracker / dash-to-dock / appindicators / panel-dash extensions: fallback icon `application-x-executable` (gray gear/cog).
- KDE task manager: generic placeholder.
- Distro-independent — same `.desktop` + same binary across deb/rpm/AppImage/nix. All four generators emit `StartupWMClass=$WM_CLASS`.

### Verified

- Ubuntu 24.04, GNOME 46, X11.
- `xprop WM_CLASS` = `"claude-desktop", "claude-desktop"`; process argv carries `--class=Claude`.
- GTK `IconTheme` loads the SNI tray PNG fine → tray icon not the cause; cog = unmatched window.

## Fix

- `build.sh`: `WM_CLASS` `Claude` → `claude-desktop`. Single constant → all four `StartupWMClass=$WM_CLASS` writes + launcher `--class` + doctor + pgrep.
- `scripts/patches/app-asar.sh`: productName guard asserted `productName == WM_CLASS` (now false). Decoupled → compares productName against literal `Claude` (rename tripwire only).
- `scripts/launcher-common.sh`: two comments corrected (WM_CLASS source = ELF basename, not productName).

### Not changed

- Package name, ELF basename, upstream productName.
- `--class` mechanism — still passed; Electron still ignores it. Kept for the documented #647/#652 contract.

### Not tested

- Full package rebuild (deb/rpm/AppImage) — change is a constant value + guard literal.
- Non-GNOME shells (KDE, etc.) — inferred from shared `StartupWMClass` matching.

---
90% AI / 10% Human
Claude: diagnosed the WM_CLASS/StartupWMClass mismatch from live X11 state, wrote the constant + guard fix and CHANGELOG entry.
Human: reported the cog symptom, set scope (all generators), reviewed.


aaddrick:
Built the PR tip (7993581) as deb+rpm and ran both on the VM fleet. On Ubuntu 24.04 GNOME (Wayland session, app on XWayland, the shipping default) `xprop` shows `WM_CLASS "claude-desktop", "claude-desktop"`, the installed `.desktop` carries `StartupWMClass=claude-desktop`, and the dash shows exactly one properly-iconed Claude entry with no generic-gear duplicate. On Fedora 43 niri (native Wayland, wlroots) `niri msg windows` reports App ID `claude-desktop`, matching `StartupWMClass`. So the fix does what it claims on both backends, live.

For history: this is effectively a revert of #655, which flipped the same constant `claude-desktop`→`Claude` on the opposite rationale (#648 had set `claude-desktop` originally). What settles it this time is upstream's own `.desktop` shipping `StartupWMClass=claude-desktop` plus the ELF-derived runtime identity, and now the fleet confirms it. One thing worth flagging in that light: CI is green here because the bats/smoke corpus bakes its own `'Claude'` via sed (`tests/launcher-common.bats:64`, `tools/chromium-switch-smoke.sh:33`, `tools/chromium-switches.baseline`) rather than reading `build.sh`'s constant. A future flip back would sail through untested. Folding those three value updates into this PR would pin the settlement.

Two small pre-merge asks. Rebase onto current main. Three CHANGELOG bullets landed today and merge-tree shows `CHANGELOG.md` is the only conflict, mechanical keep-both. And add the #779 issue link to the CHANGELOG bullet. It currently links only the PR, but the sibling bullets link their issues.

A couple of follow-ups we'll take downstream, not blocking this PR. The niri test-harness filter at `tools/test-harness/src/lib/input-niri.ts:215` (`app_id !== 'Claude'`) fails open once the real app_id is `claude-desktop`, so soft-tagging @sabiut on that. And two docs this PR proves wrong: `docs/learnings/official-deb-rebase-verification.md:125-129` still calls the official value "likely an upstream bug", and the upstream-reports README row still lists the mismatch as a to-file report. Merging this closes #779.

---

jpetzke:
Heads up: on upstream 1.19367.0 the runtime WM_CLASS is `com.anthropic.Claude`, not `claude-desktop`.

Fedora 44, KDE Plasma, Wayland session, app on XWayland:

```
resourceClass  s "com.anthropic.Claude"
resourceName   s "com.anthropic.claude"
desktopFile    s ""
```

The reporter in #779 sees the same thing: `claude-desktop` worked for him until a recent update, `com.anthropic.Claude` works now.

Which upstream version did the fleet build against? If it's 1.18286.x, this fixes the value from two releases ago and lands as flip #4 on the same constant.

Might be worth deriving WM_CLASS from the packaged app at build time instead of hardcoding it. Upstream clearly feels free to change it.

aaddrick:
Thanks for chasing this down, @diarized — closing this as superseded by #799, but the credit here is yours and I want to lay out why.

First, a correction I owe you and this thread. My July 11 endorsement was verified against a 1.18286.2-pinned build, and it repeated the same ELF-basename mechanism your PR text used. That mechanism is wrong — it was only coincidentally right for that one release. Sorry for signing off with the wrong reasoning attached.

What's actually going on: Chromium/Electron derive the Wayland `app_id` / X11 `WM_CLASS` from the asar's `package.json` `desktopName` field, not the ELF basename and not the launcher's `--class` flag. Upstream 1.18286.0 shipped `desktopName` as `claude-desktop.desktop`, which happened to equal the ELF basename, so the theory held on my fleet run. Then 1.19367.0 renamed it to `com.anthropic.Claude.desktop`, renamed its own `.desktop` file to match, and set `StartupWMClass=com.anthropic.Claude` — with a literal comment in the entry saying it "Matches the Wayland app_id / X11 WM_CLASS Chromium derives from package.json desktopName." New reports on #779 (@jpetzke on Fedora KDE, @BashBandito on GNOME) confirmed the runtime class had moved. I re-verified live today on the fleet: the drift reproduces on both Ubuntu 24.04 GNOME and Fedora 43 KDE Plasma against 1.19367.0, `xprop` reporting `com.anthropic.Claude` while the launcher's `--class` is ignored.

Your core diagnosis was right and it still stands: our shipped `.desktop` `StartupWMClass` never matched the runtime window class, which is the whole bug. The only thing that changed under us was the upstream value, twice in two releases (recall #648 set `claude-desktop`, #655 flipped to `Claude`). That's exactly why #799 doesn't hardcode a third constant. It derives `WM_CLASS` at build time from the staged asar's `desktopName` via `_derive_wm_class` in `scripts/patches/app-asar.sh`, fails the build loudly if the field goes missing or odd-shaped, and asserts end-to-end that the installed `.desktop` `StartupWMClass` equals the asar `desktopName` (suffix stripped) across all three formats. It also keeps your `productName`-tripwire decoupling idea, narrowed to guarding `~/.config/Claude`, plus mutation-checked bats coverage and the launcher corpus updated to `com.anthropic.Claude`. You're credited in `ACKNOWLEDGMENTS.md`.

No ask of you here — #799 carries it from here, and review or pushback over there is welcome. Merging it still closes #779.

---

PR #792, opened 2026-07-12
fix(build): re-point AU-1 tripwire to upstream's renamed updater marker

sabiut:
Fixes the repo-wide CI failure since the 1.19367.0 pin (fcfb72f): every build leg exits on the AU-1 tripwire because upstream dropped the `apt_channel_pending` string.

## What upstream actually changed

I extracted and diffed the official 1.18286.2 and 1.19367.0 bundles before touching the tripwire. The Linux updater gate is **unchanged in shape** — a constant-folded early-return that never reaches `setFeedURL`/`checkForUpdates` scheduling. Only the wording moved:

Before (1.18286.2):
```
D.info("[updater] Linux: in-app updater off (apt channel not yet live)"),Ye("desktop_update_disabled",{reason:"apt_channel_pending"});return
```

After (1.19367.0):
```
v.info("[updater] Linux: in-app updater off (updates via apt)"),nt("desktop_update_disabled",{reason:"managed_by_package_manager"});return
```

The "pending" framing is gone because the APT channel is live: Linux updates are now *permanently* the package manager's job. Corroborating evidence in the new bundle: the updater-status accessor hardcodes `platformUnsupported:!0`, the renderer-facing `checkForUpdates` IPC on Linux just opens the claude.ai download page, and the only path to a real `autoUpdater.checkForUpdates()` is gated behind an Ed25519-signature-verified `CLAUDE_UPDATER_TOKEN` env override (Anthropic-internal test hook, defaults inert). So **D-001 is strengthened, not threatened** — this is the anchor-rename case, not the updater-flip case.

## The fix

- AU-1 now greps for `managed_by_package_manager` — exactly 1 occurrence in the asar, inside the gate itself, so it disappears iff upstream rewrites that gate (same property the old anchor had, literals-over-identifiers per docs/learnings/patching-minified-js.md).
- Tripwire comment block documents the rename and window (1.18286.2 → 1.19367.0).
- `tests/app-asar.bats` fixtures updated to the new marker (using the real bundle's `desktop_update_disabled` shape); 5/5 green.
- CHANGELOG entry under [Unreleased] → Fixed.

## Verification

- Ran `_check_upstream_tripwires` against both real bundles: 1.19367.0 → `Upstream tripwires clear` rc=0; 1.18286.2 → fires AU-1 rc=1. The check is re-pointed, not weakened.
- MB-1 confirmed still matching in 1.19367.0 (`menuBarEnabled:!0`, 1 occurrence) — it won't fire next.
- Full bats sweep green; shellcheck delta vs main is zero; 80-col clean.

Unblocks CI for #744, #745, #781, #782 once merged.

---
90% AI / 10% Human
Claude: bundle diff/investigation of both official .debs, anchor selection, tripwire+test+changelog edits, both-ways verification against real bundles
Human: direction, review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,16 @@ The format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) — 
 
 ### Fixed
 
+- Builds against upstream 1.19367.0 no longer fail on the AU-1
+  tripwire. Upstream renamed the Linux updater's disable reason from
+  `apt_channel_pending` to `managed_by_package_manager` — the APT
+  channel went live and Linux updates are now permanently delegated to
+  the package manager. The updater gate itself is unchanged (verified
+  against the official bundle: the constant-folded early-return,
+  `platformUnsupported:!0`, and the Linux `checkForUpdates` IPC stub
+  that just opens the download page all survive), so decision D-001
+  (`docs/decisions.md`) is unaffected and the tripwire anchor now
+  tracks the renamed literal.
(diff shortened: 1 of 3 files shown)


aaddrick (approved):
Good to go.

aaddrick:
Also, apologies about the merging of the pile of PR's yesterday that had you as reviewer. Claude worked around a hook. Working to correct that

aaddrick:
More failures after the fix. working through the chain now

PR #794, opened 2026-07-12
test(app-asar): cover _resolve_main_js code-split chunk resolution

aaddrick:
Adds unit-test coverage for `_resolve_main_js`, the code-split main-chunk resolver from #793. I split this out of #793 on purpose. It keeps the test surface in @sabiut's lane and lets #793 stay focused on the build fix.

This is stacked on #793. The base branch is `fix/code-split-main-chunk-resolution`, because `_resolve_main_js` only exists there. GitHub will retarget it to `main` once #793 merges.

Six new bats cases in `tests/app-asar.bats`, next to the existing tripwire tests:

- follows the stub's `require("./index.chunk-<hash>.js")` to the main chunk
- falls back to `index.js` on the old single-file layout
- ignores non-chunk requires (`node:`/`electron`/other) when falling back
- fails loud when the stub requires two main chunks (a future deeper split means you re-point manually)
- fails loud when the named chunk file is missing
- fails loud when `index.js` is absent

I mutation-checked these so they aren't hollow. Neutering the require-follow fails the "follows the stub require" case, and loosening the multi-chunk guard fails the "two main chunks" case. Full suite is 11/11 green.

@sabiut tagging you as the tests/doctor owner per CODEOWNERS whenever you have a minute.

---
90% AI / 10% Human
Claude: wrote the six resolver bats cases + the _make_build_tree helper, mutation-checked them
Human: direction (split tests into their own PR), review


sabiut (approved):
Reviewed and verified locally. Approving.

- **Suite**: 11/11 green at the PR head (retargeted onto main post-#793).
- **Branch coverage is complete**: all five resolver outcomes are pinned — happy path, pre-split fallback, non-chunk-require fallback, multi-chunk fail, missing-chunk fail, missing-index fail.
- **Re-ran both claimed mutations plus my own**: neutering the require-follow fails tests 6/9/10 (not just the happy path — the failure modes collapse into the fallback, which the exact-path asserts catch); loosening the multi-chunk guard to >99 fails exactly the two-chunks test. Not hollow.
- **Real-bundle check**: ran `_resolve_main_js` against the actual extracted 1.19367.0 tree — resolves to `index.chunk-CNXUb5h4.js`, matching the fixture's chunk name and the chunk the AU-1 investigation located the updater gate in. The fixture models reality, not a convenient shape.
- `_make_build_tree`'s cd is safe under bats per-test subshells, and shellcheck is clean modulo the file's pre-existing informational classes.

One observation, not blocking: a stub that requires the *same* chunk twice would trip the multi-chunk guard (mapfile keeps duplicates) even though it's unambiguous. That's fail-loud in a direction that can only cause a build stop, never a mispatch, so I'd leave it — noting it here so it's a known behavior rather than a surprise.

PR #796, opened 2026-07-12
docs(learnings): test methodology and coverage (from @sabiut's PRs)

aaddrick:
Docs-only. Adds a new learnings doc that writes up @sabiut's testing methodology and coverage discipline, mined from his test/CI/doctor PRs and his reviews on other contributors' work. No code, no build, no behavior touched.

## What's here

- **New doc:** `docs/learnings/test-methodology-and-coverage.md`. The spine is one rule: a green test proves nothing until you've shown it goes red when the code it guards is broken. That's the mutation check, and the doc treats it as mandatory.
- **The half-pinned-test failure class.** Tests that shipped green while pinning nothing. `run helper` subshells away `_doctor_failures` and flag mutations, so you assert on the state directly, not through `run`. Anchor tests with no near-miss fixture. A stub that mirrors the production call can't catch a change to that call. `[PASS]` printed over data nothing ever read. A poll predicate that has to match the reaper's own predicate. SC2314 negative-assertion no-ops. Each one cites the PR it came from.
- **Host-state isolation.** When to stub an in-shell call vs. PATH-shim a command-substitution call, and unsetting every `XDG_*` / `_DOCTOR_*` var in `setup()` with a `$HOME` or system-path fallback so the suite doesn't read the tester's real machine.
- **Artifact launch-smoke methodology.** `setsid` plus `kill -- -PGID` process-group reaping, a readiness-marker poll instead of a flat sleep, rpm privilege-drop, native-arch runners, one shared cleanup trap.
- **Review heuristics.** Mutation check required, claimed verification has to ship as a committed test, doctor-vs-launch parity, honest static-vs-live labeling.
- **Wiring:** the doc is added to `docs/index.md` and the Learnings list in both `CLAUDE.md` and `AGENTS.md`.

Every claim traces to a real PR (#520, #534, #592, #646, #671, #691, #692, #693, #740, #744, #745, #774, #781, #782, #713, #752, #775, #776, and others) and I checked each against the current code anchors, not stale ones.

This is @sabiut's methodology. I mined it and wrote it up. Tagging him for review as the tests/doctor owner per CODEOWNERS, since he's the source of record here.

## Caveat

`CLAUDE.md` and `AGENTS.md` are supposed to be byte-identical below the H1, but they've already drifted. `AGENTS.md` is missing the `config-wipe-guard` bullet and some of the newer wording from earlier PRs. This PR adds the new learnings bullet to both files, but it does not try to reconcile the pre-existing drift. That's a separate cleanup and should land on its own so this diff stays docs-add-only.

---
85% AI / 15% Human
Claude: Mined @sabiut's test/CI/doctor PRs and reviews, distilled the methodology, drafted the doc, verified each citation against current code anchors, and wired it into the index and Learnings lists.
Human: Requested the doc, set the scope and framing, flagged the CLAUDE.md/AGENTS.md drift as out of scope, and reviewed.


sabiut (approved):
Reviewed as the source of record — approving. This is an accurate statement of the methodology; nothing here is embellished or misattributed.

Verified rather than skimmed:

- **The four case studies I can check against this week's work are exactly right**: the #745 stat-stub gap (`$2 == '%a'`, the GNU `-f` free-block reinterpretation, real-0644 FAIL test + stub kept for un-fakeable 4755+root), the #782 near-miss (`claude-desktop (unconfined)`, weakening goes from survives-all-7 to fails-3 — the measured number), the #781 poll-predicate drift (pre-exec quoting window, predicate-identity fix, loud timeout), and the #744 run-subshell example, which mirrors the committed test line-for-line.
- **Spot-checked the code anchors live at the PR head**: `[[ $avail =~ ^[0-9]+$ ]]` + `avail=$((10#$avail))` in doctor.sh, `readiness_marker='Executing: '` in test-artifact-common.sh, and the SC2314 disables in launcher-common.bats all exist as cited. The 88/97 suite counts are exact at current main.
- **Wiring**: index.md + CLAUDE.md + AGENTS.md bullets present; the drift caveat is the right call — keeping this docs-add-only is cleaner than bundling a reconciliation.

One non-blocking suggestion: the exact test counts (88/97) in the source-files list will be stale within days — the five campaign PRs merging today alone will move both numbers. Consider "~90"-style approximations or dropping the counts; happy for it to land as-is and age normally otherwise.

Genuinely useful doc — the "question 6 is the one that matters" framing is the right compression of the whole discipline.

PR #799, opened 2026-07-12
fix(desktop): derive StartupWMClass from the asar desktopName

### LINKED ISSUE #779: [bug]: StartupWMClass is incorrect and produces a second generic icon

aaddrick:
Fixes #779. Supersedes #786.

## The bug

Our packages ship `StartupWMClass=Claude` while the runtime window class is something else entirely, so GNOME and KDE can't associate the running window with the launcher entry — second taskbar entry, generic gear/`?` icon (#779).

## Why a derivation instead of a new constant

Chromium derives the X11 `WM_CLASS` / Wayland `app_id` from the asar `package.json` `desktopName` field minus its `.desktop` suffix — not from `productName`, not from the launcher's `--class` flag, and not from the ELF basename. #786's `claude-desktop` was correct for 1.18286.0 only because that release's `desktopName` (`claude-desktop.desktop`) happened to coincide with the ELF basename; upstream 1.19367.0 renamed the field to `com.anthropic.Claude.desktop` (and renamed its own `.desktop` file and `StartupWMClass` to match), which is why @BashBandito's override stopped working after the update. The constant has now flip-flopped three times in project history (#648 → #655 → #786) while upstream moved the underlying value twice in two releases — so the fix reads the field Chromium reads:

- `_derive_wm_class` in `scripts/patches/app-asar.sh` derives `WM_CLASS` from the staged `app.asar`'s `desktopName` at build time and exports it for the packaging scripts; a missing or odd-shaped `desktopName` fails the build loudly (same tripwire philosophy as AU-1/MB-1).
- `run_packaging` refuses to package with an empty `WM_CLASS` (belt and braces).
- The `productName` check narrows to what it still actually guards: Electron's `~/.config/Claude` userData path.
- `--class` stays on the launcher cmdline: Chromium ignores it for the window class, but it's load-bearing as the `/proc` cmdline UI fingerprint (`_claude_desktop_ui_cmdline_matches`).

## Testing

- **New bats coverage** for `_derive_wm_class` (both upstream shapes, interior-`.desktop` near-miss, empty and suffix-less failure modes), mutation-checked per `docs` test methodology: no-strip breaks 3 tests, greedy-strip breaks the near-miss test, deleting the shape guard breaks the suffix test.
- **End-to-end artifact assert**: `validate_app_contents` now checks the installed `.desktop`'s `StartupWMClass` against the asar `desktopName` in all three formats (deb/rpm/AppImage), so a hardcoded value can never silently drift from the runtime class again.
- **Test corpus settlement**: the launcher bats corpus, chromium-switch smoke sed, and switch baseline now bake the current real value `com.anthropic.Claude` (near-miss fixture preserved as `com.anthropic.ClaudeDev`).
- **Full suites green locally**: app-asar 10/10, launcher-common 97, doctor 88, xrdp/gpu 118, switch smoke matches baseline; shellcheck finding-set identical to main.
- **Live fleet verification against 1.19367.0**: built this branch's deb/rpm, installed on Ubuntu 24.04 GNOME (Wayland session, app on XWayland) and Fedora 43 KDE Plasma Wayland. `xprop` reports `WM_CLASS "com.anthropic.claude", "com.anthropic.Claude"` on both (launcher's `--class=claude-desktop` from the #786 build was ignored, confirming the mechanism). With the derived `StartupWMClass=com.anthropic.Claude`, the GNOME dock shows the proper Claude icon with running indicator; with a mismatched value it shows the generic gear.

## Also in this PR

- `docs/learnings/official-deb-rebase-verification.md`: corrected the StartupWMClass paragraph — the official value was never an upstream bug; ours was the mismatch.
- `docs/upstream-reports/README.md`: dropped the now-moot mismatch row.
- `ACKNOWLEDGMENTS.md`: credited @BashBandito (report + retests), @diarized (#786 diagnosis and the productName-tripwire decoupling, which this PR keeps), and @jpetzke (the 1.19367.0 `com.anthropic.Claude` datapoint that turned the fix into a derivation).

CHANGELOG bullet follows in a second commit with this PR's number.

---
90% AI / 10% Human
Claude: mechanism root-cause (desktopName), static + live fleet verification on both upstream releases, derivation implementation, tests, docs
Human: direction (supersede #786 with a derivation, testing bar), review


aaddrick:
@sabiut - merging without review. Let me know if i need changes.

Issue #750, opened 2026-06-26
tray.sh in-place fast-path silently broken on 1.13576+ (yukonSilver rebuild refactor) → #515 duplicate-icon race re-arms


aaddrick:
## Summary

`patch_tray_inplace_update` in `scripts/patches/tray.sh` stopped applying on Claude Desktop **1.13576+** (the "yukonSilver"-era refactor; verified against the current `detect-host.sh` target **1.15962.0**). Because tray patches warn-and-continue (the [#429](https://github.com/aaddrick/claude-desktop-debian/issues/429) class), it ships through green CI — the in-place fast-path is simply gone, which **re-arms the [#515](https://github.com/aaddrick/claude-desktop-debian/issues/515) KDE/Plasma duplicate-icon StatusNotifier race** documented in `docs/learnings/tray-rebuild-race.md`.

Surfaced while investigating [#746](https://github.com/aaddrick/claude-desktop-debian/issues/746) (Cinnamon empty-icon). Note: this is **not** the root cause of #746 — the core icon-selection patch still applies on both 1.15200.0 and 1.15962.0 — but it's a real latent regression found in the same area.

## What changed upstream

Extracted the real `1.15962.0` `app.asar` and ran `tray.sh`'s extraction against the shipped minified `index.js`.

**1. The destroy/recreate guard restructured.** Old shape the injection anchored on:

```js
;if(TRAY&&(TRAY.destroy(),TRAY=null)…
```

New shape (state resets hoisted into the `if`):

```js
const t=X.join(toi(),e),i=!xrt;if(A9=[],e9=!1,vE&&(vE.destroy(),vE=null),!A){JhA();return}
```

The regex anchor `;if\(TRAY&&\(TRAY\.destroy\(\)` no longer matches.

**2. Context menu is now a prebuilt object, with a `null` decoy.** Old: `setContextMenu(BUILDER())`. New: the menu is built once (`yh=S5A()`) and set by name, and the **first** `setContextMenu` call site is a menu-*clear*:

```js
…vE.setContextMenu(null)…           // invalidation clear  ← FIRST in file order
…yh=S5A(),…vE.setContextMenu(yh)…   // the real one; builder is S5A()
```

The resolver's `head -1` latched the `null` clear, so the builder never resolved → the [#680](https://github.com/aaddrick/claude-desktop-debian/issues/680) `WARNING: could not resolve tray menu function` fires and the fast-path is skipped.

[...]

Issue #656, opened 2026-05-27
[bug]: Tray patch fails to extract electron variable on upstream 1.9255.0


aaddrick:
The `extract_electron_variable` function in `scripts/patches/_common.sh` fails on upstream 1.9255.0. Both grep patterns miss, so the build exits with:

```
Failed to extract tray variable name
```

## Affected builds

All formats, both architectures. Confirmed on:

- CI run [26485783980](https://github.com/aaddrick/claude-desktop-debian/actions/runs/26485783980) (main)
- CI run [26493252671](https://github.com/aaddrick/claude-desktop-debian/actions/runs/26493252671) (PR branch)

Local builds against the cached 1.8555.2 installer still pass. Only 1.9255.0 is affected.

## What's happening

The two extraction patterns in `_common.sh` anchor on `require("electron")` and `new <VAR>.Tray`:

```bash
electron_var=$(grep -oP '[$\w]+(?=\s*=\s*require\("electron"\))' "$index_js" | head -1)
# fallback:
electron_var=$(grep -oP '(?<=new )[$\w]+(?=\.Tray\b)' "$index_js" | head -1)
```

Anthropic re-minified the bundle in 1.9255.0 and these anchors no longer match. This is the expected breakage pattern documented in `docs/learnings/patching-minified-js.md` — minified identifiers and surrounding syntax churn between releases.

## Fix path

Pull the 1.9255.0 installer, extract `index.js` without beautifying, and find the new shape of the `require("electron")` assignment. Update the grep pattern in `extract_electron_variable` to match.

---
Written by Claude Opus 4.7 via [Claude Code](https://claude.ai/code)

PR #800, opened 2026-07-13
fix(tray): pick light Linux tray icon on Cinnamon dark panels (#604)

mondalaci:
## Summary

- Upstream already ships `TrayIconLinux.png` (dark glyph) and `TrayIconLinux-Dark.png` (light glyph) but picks between them from `nativeTheme.shouldUseDarkColors` plus a GNOME check. Cinnamon often uses a dark panel while GTK still reports a light colour scheme, so the black icon disappears on the tray ([#604](https://github.com/aaddrick/claude-desktop-debian/issues/604)).
- The launcher auto-detects Cinnamon themes whose `org.cinnamon.theme` name looks like a dark panel style (e.g. Mint-Y-Dark-Aqua) and exports `CLAUDE_TRAY_USE_DARK_ICON=1`.
- A small asar patch threads that flag into upstream's existing ternary — no PNG rewriting, no change to upstream icons.

## Test plan

- [x] `bats tests/tray-icon-selection.bats` — patch injects `CLAUDE_TRAY_USE_DARK_ICON`, idempotent, warns on missing anchor
- [x] `bats tests/launcher-common.bats -f setup_tray_icon_env` — preset honored, Cinnamon dark theme sets flag, light theme no-op, non-Cinnamon no-op
- [x] Manual: rebuilt `.deb`, `apt install --reinstall`, verified white glyph on Linux Mint Cinnamon dark panel
- [ ] CI green

---
Co-Authored-By: Claude <noreply@anthropic.com>
90% AI / 10% Human
Claude: diagnosis, launcher auto-detect, asar patch, tests, docs
Human: reported the Mint tray contrast bug, validated the fix locally

Made with [Cursor](https://cursor.com)


aaddrick (dismissed):
Thanks @mondalaci, and nice diagnosis. I verified it against the pinned 1.19367.0 reference: upstream's DE detector (`oPe()`, beautified chunk line 75724) splits `XDG_CURRENT_DESKTOP` and only ever returns "kde", "gnome", or "other", so Cinnamon falls to "other" and the icon selection collapses to `shouldUseDarkColors` alone. The anchor is the sole Linux selection site (`ymr` is baked to "png", and each `TrayIconLinux` literal appears exactly once). And because the env read lives inside the selector, which re-runs on nativeTheme "updated", the flag survives theme-change rebuilds. This is the shape @IvanTheGeek sketched in #604, and it clears the D-002 patch-zero bar since upstream already special-cases GNOME right here. Approach accepted. Four things before merge.

First, the semantics of `0`. docs/configuration.md and docs/troubleshooting.md say `CLAUDE_TRAY_USE_DARK_ICON=0` forces the dark `TrayIconLinux.png`, but the patch only ORs in the `==="1"` case, so `0` can't override the GNOME check or `shouldUseDarkColors=true`. As written, `0` means "suppress the launcher auto-detect," nothing more. Either make the injected condition tri-state so `"0"` forces the plain icon, or fix the docs to match what the sed actually does.

Second, the anchor hardcodes the minified identifiers `oPe` and `G`. Those names rotate on every upstream re-minify, and the tray code specifically has burned us on this repeatedly: #218/#219 (oe/Ae), #252 and #625 (the `$`-identifier trap), #656 (anchor rot). Re-anchor on the stable literals `"TrayIconLinux-Dark.png"`/`"TrayIconLinux.png"` and capture the identifiers with `[$\w]+`, per the literals-over-identifiers section in docs/learnings/patching-minified-js.md.

Third, the anchor-miss path warns and returns 0. That's the #429 failure class, broken patches shipping through green CI, and it bit us as recently as the yukonSilver rebuild (#750). It's worse now than it was then: check-claude-version auto-tags releases with no human in the loop, so a rotted anchor would ship silently forever. Make it fail the build (`return 1`) the way scripts/patches/virtiofsd-probe.sh does. The hard fail doubles as the tripwire that tells us when upstream teaches its own detector about Cinnamon and this patch can retire.

Fourth, the bats stubs. The gsettings stubs in tests/launcher-common.bats use `#!/bin/sh` but run `[[ ]]`. On Mint that passes because /bin/sh is bash, but CI is ubuntu-latest where sh is dash, so the stub exits 127: the dark-theme test goes red, and the light-theme test passes for the wrong reason (a broken stub returns early, which is indistinguishable from the theme check actually working). Change the stub shebangs to bash. See docs/learnings/test-methodology-and-coverage.md on half-pinned tests and the mutation-check discipline.

Two minor things. The branch predates #799, so the CHANGELOG hunk adds a second `### Fixed` under `[Unreleased]` — rebase and fold the bullet into the existing section. And consider renaming the function: `patch_tray_icon_selection` is the exact name of a deleted 2.x function that did the macOS template-icon rewrite, and it's referenced all through docs/reports/CDL-ANT-0009_patch-suite-history/dossiers/tray.md — a distinct name keeps the archaeology unambiguous.

Happy to commit these four changes straight to your branch if that's easier, or leave them to you — either's fine, just let me know. Separately I'll file the upstream report, since teaching the existing DE detector about Cinnamon is a small ask, and this patch is the interim fix pending that. On merge I'll add both you and @IvanTheGeek to ACKNOWLEDGMENTS.md — his #604 analysis is basically the design doc for this.

---

aaddrick:
One more thing while you're in there for the anchor rewrite. Re-anchoring on the stable literals also needs a near-miss fixture in the patch tests, per the rule we distilled from #782 in [`docs/learnings/test-methodology-and-coverage.md`](https://github.com/aaddrick/claude-desktop-debian/blob/main/docs/learnings/test-methodology-and-coverage.md). The current warn-path fixture (`case"png":t="TrayIconLinux.png";break`) is nowhere near the anchor, so if someone weakens the patch (drops the GNOME half, loosens the identifier capture) all three tests still go green. A good near-miss is the beautified-spacing form of the ternary, or a lookalike ternary with different icon literals. One character short of matching is the whole point.

Related: the inject test only greps for the `CLAUDE_TRAY_USE_DARK_ICON==="1"` substring. It never pins that the condition landed inside the ternary, so a sed that mangled the line but kept the marker would pass. Assert the full post-patch expression through to `?"TrayIconLinux-Dark.png"` so placement is pinned, not just presence.

Non-blocking, but worth folding in the same pass: a `--doctor` line reporting what the tray auto-detect decided (preset, auto-detected dark, or no-op), the way the password-store check already reports its mode. Without it the next #604-class report costs another env round-trip. With it the bug report carries the answer. Same doctor-vs-launch parity idea as #776.

The offer from the review still stands. Happy to fold all of this in on your branch if you'd rather I just push it, only say the word.

---

aaddrick:
Quick update, @mondalaci. The upstream report is now filed as anthropics/claude-code#77170, so this patch is formally interim-pending-upstream. That's exactly the framing from my review, and the hard-fail anchor doubles as the retirement tripwire once upstream fixes it.

While verifying on KDE I also found and filed anthropics/claude-code#77171. The tray goes deaf to theme changes after the first one. Worth knowing since it touches this fix's UX, though it doesn't make things worse. An auto-detected value only applies at launch anyway, so the env-var approach isn't affected by it. No change to my review asks. Still happy to push the fixes to the branch if you'd rather I take them.

---

mondalaci:
Thanks for reviewing and elaborating! Please commit the suggested changes. You have a much deeper understanding of this than I have, and I trust your judgment.

aaddrick:
Thanks for the trust, @mondalaci. I pushed to your branch. It's two commits now after a rebase onto current main: your original (authorship preserved, CHANGELOG bullet moved into the `### Fixed` section that landed post-#799), and a second commit with all the review fixes.

Here's what's in that second commit:

1. Patch is now a node-heredoc, same shape as `virtiofsd-probe.sh`, anchored on the `"TrayIconLinux-Dark.png"` / `"TrayIconLinux.png"` literals with captured identifiers instead of hardcoded `oPe`/`G`. It also tolerates the bundler indirect-call form `(0,i.oPe)()` and property-chain electron handles, so a benign re-minification won't burn a release tag.
2. True tri-state. The injected condition is `FLAG==="1"||FLAG!=="0"&&(<upstream condition>)`, so `0` genuinely forces the plain icon over the GNOME check and `shouldUseDarkColors`, which is what the docs promised. Non-0/1 values get a launcher log line and a doctor WARN instead of silently killing auto-detect.
3. Anchor miss fails the build now (`return 1`), with an ERROR that names the retirement path (upstream anthropics/claude-code#77170 plus the local `docs/upstream-reports` record) and the stopgap to unblock a release (drop from `active_patches`, file a follow-up).
4. Idempotency guard keys on the injected expression, not the bare env-var name. A contrarian pass caught that upstream adopting the same variable name (which our upstream report advertises) would otherwise silently disarm the patch.
5. gsettings stub shebangs are `bash` now (the dash-on-CI trap), and the light-theme stub writes a ran-marker the test asserts.
6. Function renamed to `patch_tray_icon_env_override` to dodge the archaeology collision with the deleted 2.x `patch_tray_icon_selection`.
7. `--doctor` reports the tray mode (preset / Cinnamon auto-detect / upstream default) by calling the launcher's own `setup_tray_icon_env`, so doctor and launch can't drift.

One thing worth flagging gently: your commit introduced a latent failure in `log_session_env`'s exact-line test, which wasn't updated for the new key. The full launcher-common suite was red at test 12. A filtered test run wouldn't have caught it, so I fixed it in the same commit. Tests otherwise cover the real 1.19367.0-minified fixture with the full post-patch expression pinned, beautified-spacing, bundler-shape, byte-identical idempotency, and the near-miss cases (no-GNOME half, Win32-ico lookalike, duplicate site) all asserting the hard fail with output pins.

Verification I actually ran: patch applied against the real pinned 1.19367.0 chunk and the beautified reference, `node --check` on the output, byte-identical re-run, the six-case truth table for the tri-state, all four bats suites green (8 tray + 107 launcher + 130 doctor + 16 app-asar), two mutation checks (reverting the soft-fail and weakening the anchor both turn tests red), and shellcheck clean against the main baseline. I did not test it live on a real Cinnamon panel with the rebuilt package. Your original Mint test predates the tri-state change, so if you get a chance, a quick re-test of `CLAUDE_TRAY_USE_DARK_ICON=0` on your box would be a nice confirmation. Totally optional.

Last thing: I rebased and force-pushed, so your local checkout is behind. To sync:

```
git fetch origin
git checkout fix/604-tray-icon-cinnamon
git reset --hard origin/fix/604-tray-icon-cinnamon
```

---

sabiut (approved):
All four test suites pass locally: tray-icon-selection 8/8, launcher-common 107/107, doctor 130/130, and app-asar 16/16.

I applied the patch to the actual pinned 1.19367.0 chunk (index.chunk-CNXUb5h4.js from the official .deb). It captures oPe and G correctly, passes node --check, and produces exactly one injection. I also checked the full six-case truth table: 1 selects dark, 0 selects plain and overrides upstream, while unset or invalid values preserve upstream's choice.

I reran the mutation checks myself. Changing the anchor-miss hard failure back to a soft failure correctly breaks all four failure-path tests. Relaxing the exactly-one requirement to at-least-one breaks only the duplicate-site test. The near-miss fixtures—the missing-GNOME half and the Win32 ICO lookalike—are both genuine one-edit-away cases.

Doctor behaviour matches the launcher. run_doctor loads the launcher config before _doctor_check_tray_icon, and the check calls setup_tray_icon_env itself. That means both config-file presets and auto-detection report the same result the launcher would use. Informational-only reporting, without showing PASS for unread state, is the right severity.

The Bash-shebang stubs, including the explanation for using Dash in CI, and the ran-marker on the light-theme stub close the remaining half-pinned gaps identified in the original review.

The only open item is the one already noted: there has not yet been a live retest of CLAUDE_TRAY_USE_DARK_ICON=0 on a real Cinnamon panel since the tri-state change. That remains an optional confirmation for @mondalaci. Given the committed truth-table and fixture coverage, I'm comfortable merging without it.

mondalaci:
Just retested, and the desired icon appears in the dark system tray of my Linux Mint installation.

PR #808, opened 2026-07-18
fix(nix): ship virtiofsd in the FHS env so Cowork's VM gate resolves

### LINKED ISSUE #806: [bug]: Cowork gated on NixOS — the Nix format applies no asar patches, so the #771 virtiofsd fix never reached it
## Description

Cowork is greyed out on NixOS with the **"Cowork requires QEMU. Install it with `sudo apt install qemu-system-x86 ovmf virtiofsd`, then restart Claude"** popup, with KVM, vhost-vsock, qemu and OVMF all present and healthy.

This looks like #771 reaching the Nix format only. The #773 fix (`patch_virtiofsd_probe`, v3.0.1) is an asar patch, and the Nix build path doesn't apply asar patches — since the v3.0.0 official-deb rebase `nix/claude-desktop.nix` repackages the stock `.deb` with ELF/Vulkan fixups only, and `scripts/patches/*.sh` are sourced from `build.sh` for deb/appimage/rpm but never from Nix. So deb/appimage/rpm/AUR got the un-gate and Nix didn't.

## Version

The Nix packaging has no `--doctor` — `scripts/doctor.sh` is wired into the deb/appimage/rpm launchers via `launcher-common.sh`, but the Nix `makeWrapper` doesn't call it, so `claude-desktop --doctor` just launches the GUI. Manual equivalent:

- **Package:** `claude-desktop.packages.x86_64-linux.claude-desktop-fhs`, flake rev `296395e`
- **App version:** 1.20186.1 (that rev now builds 1.21459.0; the gap is builder-side, not version-specific)
- **Host:** NixOS 26.11, kernel 6.18.38, `/dev/kvm` `crw-rw-rw-` (0666), `kvm-intel` loaded, `/dev/vhost-vsock` present

## Steps to reproduce

1. On NixOS, install `claude-desktop-fhs` from the flake (rev `296395e`, or HEAD — `nix/fhs.nix` is identical between them).
2. Make sure KVM is usable (`/dev/kvm` present and writable).
3. Launch Claude Desktop, open the Cowork tab.

## Expected behavior

Cowork works as it does on the other formats post-v3.0.1.

## Logs / errors

`~/.config/Claude/logs/cowork_vm_node.log`:

```
[cleanupVMBundleIfUnsupported] yukonSilver not supported (status=unsupported), checking for stale bundle...
[startVM] VM not supported (linux/x64), skipping
```

(Same `(linux/x64)`-prints-for-any-unsupported-status thing you flagged on #771.)

## Anything else

I read the probe out of the installed 1.20186.1 asar (`.vite/build/index.chunk-BcRhoXVd.js`) to check the diagnosis rather than assume it carried over. De-minified:

```js
const aRt = ["/usr/libexec/virtiofsd", "/usr/bin/virtiofsd"];

async function cRt() {                       // isUbuntu22
  const e = await vne();                     // parsed /etc/os-release
  return e?.id === "ubuntu" && (e.versionId ?? "").startsWith("22.");
}

async function rBe(e) {                      // first readable of a path list
  for (const t of e) try { return await G.access(t, ee.constants.R_OK), t } catch {}
  return null;
}

async function lRt(e) {                      // resolve virtiofsd (e = isUbuntu22)
  const t = await rBe(aRt);
  return t || (e ? vX(iRt, ee.constants.X_OK) : null);   // bundled fallback: Ubuntu-22 only
}
```

Matches your #771 analysis exactly. On NixOS all three resolution routes are closed at once: `nix/fhs.nix` `targetPkgs` has no `virtiofsd`, so neither probed path exists in the sandbox, and os-release inside the FHS env reports `ID=nixos` (confirmed — it's the host's), so the bundled fallback stays gated. The asar is byte-unpatched: probe paths and `yukonSilver` present (×1, ×1, ×10), `cowork-vm-service.sock` and `COWORK_VM_BACKEND` absent.

The evaluator:

```js
// gate 1
!t.helperBinaryPath || !t.smolBinPath ||
(!t.virtiofsdPath && !t.virtiofsdAptInstallable) ||
(x.app.isPackaged && !t.helperSocketPeerCredentialCheckAvailable)
  → "virtualization_entitlement_missing"   // "Reinstall Claude"

// gate 2
!t.qemuPath || !t.firmwarePath || !t.virtiofsdPath
  → "virtualization_tools_missing"         // "Cowork requires QEMU..."
```

`virtiofsdAptInstallable` is `!isUbuntu22`, so `true` on NixOS — gate 1 passes and it falls through to gate 2. Gate 2 builds its install string as `["sudo apt install", "qemu-system-x86", "ovmf", ...(virtiofsdPath ? [] : ["virtiofsd"])]`, so the popup containing the word `virtiofsd` is itself confirmation that `virtiofsdPath` resolved null.

Checked every leg the two gates need against the real `claude-desktop-fhsenv-rootfs` in the store:

| Probe leg | Resolution | Status |
|---|---|---|
| `qemuPath` | PATH search, `qemu-system-x86_64` | ✅ present (#766) |
| `firmwarePath` | `R_OK` on `/usr/share/OVMF/OVMF_CODE_4M.fd`, `OVMF_CODE.fd` | ✅ both readable (#767) |
| `helperBinaryPath` | `X_OK` on `resources/cowork-linux-helper` | ✅ present, `r-xr-xr-x` |
| `smolBinPath` | `R_OK` on `resources/smol-bin.x64.img` | ✅ present |
| `kvm` | `R_OK\|W_OK` on `/dev/kvm`, `/dev/vhost-vsock` | ✅ 0666 / present |
| **`virtiofsdPath`** | `R_OK` on the two paths above | ❌ **neither exists in the sandbox** |

virtiofsd is the only failing leg — #766/#767 landed qemu and OVMF, but the evaluator rejects on virtiofsd before they matter.

### Possible fix

Adding `virtiofsd` to `targetPkgs` in `nix/fhs.nix`:

```nix
targetPkgs = pkgs: [
  bubblewrap claude-desktop docker docker-compose glibc
  nodejs openssl ovmfCompat qemu_kvm uv
  virtiofsd
];
```

Tested the mechanism in isolation — a minimal `buildFHSEnv` with `virtiofsd` in `targetPkgs`:

```
--- the two paths the client probes ---
MISSING : /usr/libexec/virtiofsd
READABLE: /usr/bin/virtiofsd -> /nix/store/...-virtiofsd-1.13.3/bin/virtiofsd
--- version via the probed path ---
virtiofsd 1.13.3
--- os-release ID (the bundled-fallback gate) ---
ID=nixos
```

`nixpkgs.virtiofsd` ships the binary at `bin/virtiofsd`, so `buildFHSEnv` surfaces it at `/usr/bin/virtiofsd` — a path the probe already checks with `R_OK`, satisfying `rBe(aRt)` without the fallback needing to be un-gated. It's the Rust implementation at 1.13.3 (bundled copy is 1.13.2), so this shouldn't run into the legacy-C-virtiofsd CLI incompatibility you cited on #771 as the reason not to widen the probe array — the concern there was `/usr/lib/qemu/virtiofsd` on qemu < 8, not the path being populated with the Rust binary. Effectively the Nix analog of the `ln -s /usr/lib/virtiofsd /usr/bin/virtiofsd` workaround.

Two caveats:

- **Not confirmed end-to-end.** The chain is verified (probe reads `/usr/bin/virtiofsd` with `R_OK` → `targetPkgs` puts a readable Rust virtiofsd there → every other leg already passes), so I expect the gate to flip to `supported`, but the rootfs download needs a logged-in session — same wall you described on the Arch guest in #771. I haven't watched a session run.
- **Not tested against the real package.** `targetPkgs` is a hardcoded list inside `nix/fhs.nix` with no `.override` seam, and re-wrapping the non-FHS output locally means reproducing the internal `ovmfCompat` shim, which the flake doesn't expose — hence the isolated FHS test above rather than a patched build.

Separately, since the Nix path applies no asar patches, `patch_quick_window`, `patch_org_plugins_path` and `patch_cowork_bwrap` are absent from Nix builds too. Noting it as context in case it's useful; virtiofsd is the one causing a visible break.

---

nbCloud91:
Fixes #806

## What

Adds `virtiofsd` to `nix/fhs.nix`'s `targetPkgs`.

## Why

Cowork gates VM boot on three resolved paths, not two. #766 and #767 supplied `qemuPath` and `firmwarePath` for the Nix format; `virtiofsdPath` was still null, so the gate stayed shut.

The resolver in the official bundle probes `/usr/libexec/virtiofsd` then `/usr/bin/virtiofsd` for `R_OK`, and falls back to the bundled `resources/virtiofsd` only when the host is Ubuntu 22:

```js
const Ctt = ["/usr/libexec/virtiofsd", "/usr/bin/virtiofsd"];
async function Mtn(){ … (e?.id)==="ubuntu" && (e.versionId??"").startsWith("22.") }   // isUbuntu22
async function Dtn(e){ const t = await hrt(Itn); return t || (e ? Bre(Ptn, X_OK) : null) }
```

#773 un-gated that fallback through `patch_virtiofsd_probe`, but `nix/claude-desktop.nix` unpacks the official `.deb` directly and never invokes `scripts/patches/*.sh` — no asar patch reaches this format. So the Nix build falls between the two fixes: it gets neither the patched fallback nor a host `virtiofsd`.

With the probe unresolved, `_getSupport` takes the third disjunct of `!qemuPath || !firmwarePath || !virtiofsdPath` and reports `virtualization_tools_missing` — *"Cowork requires QEMU. Install it with `sudo apt install …`"* — on a host where QEMU and firmware were already present. (Note the first branch doesn't fire: `virtiofsdAptInstallable` is `!isUbuntu22`, hence `true` on NixOS.)

## Why the package, not a wider probe

On #771 the Debian `/usr/libexec/virtiofsd` was flagged as the legacy C implementation, as a reason not to widen the probe array. This PR doesn't widen it. nixpkgs' `virtiofsd` (1.13.3) is the Rust implementation and installs to `bin/virtiofsd`, which `buildFHSEnv` surfaces at `/usr/bin/virtiofsd` — the array's existing second entry. The probe is untouched; the path it already looks for simply exists now.

Closure cost is small: `virtiofsd-1.13.3` is a 1.6 MiB download / 3.9 MiB unpacked, fetched from cache.nixos.org — no local compile.

## Verification

**x86_64, against 1.21459.0 — end-to-end.** Control and treatment on the same machine one minute apart, differing only by these two lines:

```
17:03:11  [cleanupVMBundleIfUnsupported] yukonSilver not supported (status=unsupported)   ← stock
17:04:53  [Bundle:status] rootfs.img missing                                              ← patched
```

The patched build then booted a real guest — `[VM:start] Startup complete, total time: 87574ms` — with `qemu-system-x86_64` spawning `/usr/bin/virtiofsd --shared-dir /`, i.e. the path `targetPkgs` populates is the one actually used to run the guest, not merely the one that stops the evaluator objecting. A logged-in Cowork session opened, stayed responsive, and ran an agent session end-to-end. Build provenance was confirmed from the live process (`/proc/<pid>/root/usr/bin/virtiofsd` readable inside the sandbox, `/usr/libexec/virtiofsd` not).

Since #771 noted the logged-in-session boot couldn't be verified on the test fleet, that step is now covered for the Nix format at least.

**x86_64, against 1.22209.0 (current `main`) — mechanism re-verified.** The resolver above is quoted from this release, so the gate is unchanged two releases on. Building `.#claude-desktop-fhs` from this branch:

```
patched rootfs:  R_OK  /usr/bin/virtiofsd -> …-virtiofsd-1.13.3/bin/virtiofsd   (`--version` runs)
                 ABSENT /usr/libexec/virtiofsd
stock rootfs:    ABSENT /usr/bin/virtiofsd
                 ABSENT /usr/libexec/virtiofsd
```

**Not verified:** the aarch64 leg. `./build.sh --build appimage` was not run — it doesn't exercise the Nix path, and this change touches nothing else.

---

100% AI / 0% Human
Claude: static analysis of the 1.21459.0 and 1.22209.0 asar resolvers, root-cause of the deb-patch/Nix-format gap, the patch, the build verification, changelog and this description.
Human: hardware, and the interactive GUI steps no agent can drive headlessly — the logged-in Cowork session and its agent run.


aaddrick (approved):
The writeup here is the strongest part. Quoting the resolver out of two releases and pinning the failure on the third disjunct of the gate makes the whole thing easy to check rather than take on faith.

I verified the mechanism in a container today, following the `docs/learnings/nix.md` recipe (Fedora distrobox + Determinate Nix, no real NixOS). Rebuilt both flake outputs from the PR head. The base derivation resolved to `claude-desktop-1.22209.0`, which is the current upstream pin, so this holds on the bytes `main` ships now and not just the release the PR was written against. nixpkgs `virtiofsd-1.13.3` (the Rust impl, as the comment says) lands in the FHS runtime closure, and the fhsenv rootfs serves it at exactly the second probe path: `usr/bin/virtiofsd -> virtiofsd-1.13.3/bin/virtiofsd`. Nothing at `/usr/libexec/virtiofsd`, so the probe order lands on it the way you describe. One corroborating detail while I was in there: the base derivation autoPatchelf's the deb's own bundled `resources/virtiofsd`, the Ubuntu-22-only fallback the comment calls out, which lines up with the analysis of why the un-patched probe fails everywhere else.

What the container can't reach is the live KVM VM boot on real NixOS. That stays author-attested, and it's a good attestation. The aarch64 leg stays unchecked, but that's already true for the whole derivation per `docs/learnings/nix.md` and this PR doesn't widen the gap. CI is green across Shellcheck, BATS, and the full build.

Leaving the merge to @typedrat — nix is her subsystem and she'll have the better read on whether this is the right shape for the FHS env. The mechanism checks out on my end for whatever that's worth to the call.

---

PR #814, opened 2026-07-25
fix(ci): decouple reference-source tooling from the versions checkout

### LINKED ISSUE #813: Release job: reference-source tooling is gated on the claude-desktop-versions checkout
The `v3.2.1+claude1.24012.9` tag built and validated clean on both arches, then died in `Create Release` with `asar: command not found` (exit 127) — [run 30143548473](https://github.com/aaddrick/claude-desktop-debian/actions/runs/30143548473). Every publish job carries `needs: [release]`, so `mirror-official-deb`, `update-apt-repo`, `update-dnf-repo` and `update-aur-repo` all skipped and the repos stayed on 1.24012.0 until the job was re-run by hand.

The trigger was `claude-desktop-versions` going private on 2026-07-24. It's checked out with `continue-on-error: true`, which renders a **green ✓ in the run summary while `outcome` is `failure`** — so the four steps gated on `steps.checkout_versions.outcome == 'success'` silently skipped. One of them is `npm install -g @anthropic-ai/claude-code @electron/asar prettier`. `asar` is used by `Generate and upload reference source` at the end of the job, which has nothing to do with release notes and carries no such gate, so it ran without its tool and took the whole job down.

The repo is public again and the re-run is fully green, so nothing is broken right now. What's left is the coupling: tooling for one step is installed by a step gated on an unrelated repo's availability, and a convenience asset can fail the job that gates every publish channel. The reference-source step already degrades gracefully in two places (`::warning::` + `exit 0` for a missing AppImage or a missing `app.asar`) — it just had an unguarded hard dependency above them. Worth noting the gated steps exist to serve `Run compare-releases`, which is currently `if: false`.

Fix is to install `asar`/`prettier` unconditionally and add the third bail-out so the step degrades like its siblings. The green-checkmark behaviour of `continue-on-error` is the part that cost the most diagnosis time and is worth pinning in a test.

---

aaddrick:
Fixes #813.

`asar` and `prettier` move out of the versions-gated `Install Node.js tools` step into an ungated `Install asar and prettier`, and `Set up Node.js 20` is ungated with them — they serve `Generate and upload reference source`, not release notes. What stays gated on the `claude-desktop-versions` checkout is what actually needs that repo: Python, difftastic, and the `@anthropic-ai/claude-code` CLI, all of which exist for `Run compare-releases`.

The reference-source step gets its third bail-out. It already warned and `exit 0`'d on a missing AppImage and a missing `app.asar`; now it does the same on a missing `asar`, so the invariant is uniform — nothing in that step can fail a job that gates APT, DNF, AUR and the official-deb mirror.

`tests/ci-release-job.bats` pins both halves plus the premise (`needs: [release]` on all four publish jobs). It parses the release job with awk rather than adding a YAML dependency, which keeps it on the same toolchain as the rest of `tests/`; since a silently-matching-nothing parser is the obvious way that goes wrong, the first test asserts the parse's own shape and every other test asserts its step block was found before reading a field. Ten mutations, each reddening the tests it should and nothing else:

| Mutation | Reds |
|---|---|
| re-gate the asar install on the versions checkout | 3, 4 |
| drop the `command -v asar` guard | 7, 8 |
| bundle asar back in with `@anthropic-ai/claude-code` | 5 |
| delete the asar install step entirely | 2, 3, 4, 5, 6 |
| move the guard after `asar extract` | 7 |
| replace the guard with a comment that mentions it | 7, 8 |
| change a bail-out to `exit 1` | 8 |
| re-gate the node setup | 6 |
| drop `needs: [release]` from a publish job | 9 |
| rename the release job | 1–8 |

452 bats tests pass, `actionlint` is clean (the `if: false` warning on the disabled compare-releases step predates this branch). `tests.yml` picks the suite up automatically and now also triggers on `ci.yml`. The `continue-on-error`-renders-green trap and the parser-self-check discipline are written up in `docs/learnings/test-methodology-and-coverage.md`.

@sabiut — tests/doctor is yours, and this is a new test surface rather than an extension of an existing one, so the shape is the thing worth arguing with. Specifically: whether awk-on-indentation is an acceptable trade against a YAML dep, and whether the parser self-check is enough to keep it honest if someone reformats `ci.yml`.

---
Co-Authored-By: Claude <claude-opus-5> <noreply@anthropic.com>
85% AI / 15% Human
Claude: diagnosed the failed run, wrote the workflow fix, the bats suite and the mutation matrix, and the learnings entry
Human: caught the release failure, set the scope, rejected the Python dependency, and directed the review against the testing learnings


sabiut (approved):
Approved — verified hands-on.

Ran the new suite (9/9) and the full suite with the CI-identical invocation (452/452). Mutation-checked every invariant: deleting the install step reds 5, renaming the job reds 8, the comment-as-guard near-miss reds 2 — exactly matching the learnings doc's claims. Also tried re-gating the install on `checkout_versions` (2 red), moving the guard below `asar extract` (1 red), co-installing `@anthropic-ai/claude-code` (1 red), and dropping `needs:` from a publish job (1 red). Every test earns its keep, and the near-miss mutation proves the non-final `[[ ]]` asserts actually fail the test — no hollow assertions here.

Confirmed prettier needs no guard (`npx prettier ... || true` is already failure-proof), actionlint's only finding is the pre-existing `if: false` on compare-releases, and the `tests/*.bats` glob picks the new file up without workflow changes.

One non-blocking note: the "Every exit path here is a warning, never a failure" comment overstates slightly — `--appimage-extract`, `tar`, and the `gh release upload` itself still hard-fail the job under `bash -e` (a network flake on the upload would reproduce the #813 blast radius). Fine to leave for a follow-up since #813 scoped only the missing third bail-out; either soften the comment or give the upload its own warning bail-out.

CherrieTheShifter:

yannicklescure:
Flagging one thing to ride along with this PR rather than land as a conflicting follow-up: the `Set up Node.js 20` step this PR hoists out of the versions-checkout gate should become `node-version: "22"` at the same time.

### Why

`@electron/asar` moved its floor to `engines.node >=22.12.0` at 4.0.0, and every 4.x release since (4.0.0 … 4.3.0) declares the same and hard-refuses to start below it. npm reports that as an `EBADENGINE` **warning**, not an error, so the wrapper installs, lands on disk, and then fails on every invocation. This is the same trap [#839](https://github.com/aaddrick/claude-desktop-debian/issues/839) / [#841](https://github.com/aaddrick/claude-desktop-debian/pull/841) just fixed on the build side (`NODE_MIN_VERSION='22.12.0'` in `scripts/setup/dependencies.sh`, plus a `setup_asar` run-check); #841 deliberately left `ci.yml` alone because this PR was already rewriting the block.

An unpinned `npm install -g @electron/asar` does land on 4.x under Node 20 — the bare spec resolves through the `latest` dist-tag, which is not engine-filtered. Confirmed against the live registry with `npm-pick-manifest`:

```
20.19.2 -> 4.3.0
22.11.0 -> 4.3.0
22.12.0 -> 4.3.0
24.20.0 -> 4.3.0
```

So on `node-version: "20"` the release job installs 4.3.0 and gets a binary that exits 1 with an empty stdout and its complaint on stderr.

### Why it matters specifically for this PR

The `command -v asar` guard this PR adds to the reference-source step does not cover this failure mode. The binary *is* on `PATH` — it just refuses to run — so the guard passes, `asar extract` fails, and since that step carries no `continue-on-error`, the release job fails and the four publish jobs on `needs: [release]` skip. That is precisely the outage shape this PR exists to prevent, reached by a different door.

Worth noting the other consumer of the same step: `@anthropic-ai/claude-code@latest` now declares `engines.node >=22.0.0` too, so `Install claude-code CLI` is on the wrong side of the floor as well. `prettier` is `>=14`, so a bump to 22 is safe for all three.

### Suggested change

```diff
-      - name: Set up Node.js 20
+      - name: Set up Node.js 22
         uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
         with:
-          node-version: "20"
+          node-version: "22"
```

`ubuntu-latest` already builds on 22.x, and `tests/ci-release-job.bats` matches on `actions/setup-node` rather than the step name or the version, so neither the rename nor the bump reds it.

Happy to open this as a separate PR on top once this one lands if you'd rather keep the diff here focused — just seemed cheaper to fold it in than to conflict with it.

---

yannicklescure:
Opened as #845, stacked on this branch rather than on `main` so the diff stays to the Node bump and nothing here conflicts. GitHub should retarget it to `main` when this merges.

One thing from writing it that's worth knowing before anyone else annotates that block: spelling `@electron/asar` in a YAML comment above the setup step reds three tests in this PR's suite. `uncommented` is applied per step block, but the region-level greps in "node is set up before asar is installed" and "asar is installed by exactly one step" read the raw region, so the comment counts as a second step installing asar and as an asar reference preceding the Node setup. I treated that as the invariant working and kept the package names bare in the comment, but say the word if you'd rather the greps absorbed it instead.

---

PR #816, opened 2026-07-27
fix(doctor): judge only the running binary's chrome-sandbox

sabiut:
The held behavior-changing follow-up to #745, as discussed there — rebased onto current main now that the helper it modifies has landed.

When `electron_path` is set, `_doctor_check_chrome_sandbox` now validates ONLY that binary's adjacent chrome-sandbox and never falls back to the deb install path. The old first-existing-path-wins loop reported a green PASS from a stale deb tree while the actually-running AppImage/Nix binary's sandbox went unexamined — the doctor-vs-reality disagreement class. The deb path (still `_DOCTOR_DEB_SANDBOX`-overridable) remains the fallback only when no electron path is known.

The #745 precedence test is intentionally flipped to electron-wins — that flip is the behavior change, documented as such rather than silently edited.

Two regression guards, both mutation-verified against the pre-fix helper: deb-PASS-with-no-electron-sandbox and deb-PASS-shadowing-a-bad-electron-sandbox both fail without the fix and pass with it. doctor.bats 130 → 132, full suite 445/445, shellcheck clean, verified after the rebase onto 1.24012.9-era main.

### DIFF
--- a/scripts/doctor.sh
+++ b/scripts/doctor.sh
@@ -1437,9 +1437,12 @@ _doctor_check_electron_binary() {
 # Check the chrome-sandbox helper's setuid permissions (must be 4755,
 # owned by root). Official layout: chrome-sandbox sits at the package
 # root beside the ELF (no node_modules/electron/dist tree anymore).
-# Looks at the deb install path and, when an electron path is provided,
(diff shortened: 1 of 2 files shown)


CherrieTheShifter:

PR #824, opened 2026-08-22
fix: add CLAUDE_FORCE_SANDBOX escape hatch + doctor sandbox-state check

### LINKED ISSUE #804: [bug]: deb/rpm/nix launcher forces --no-sandbox on every Wayland session — Chromium sandbox silently disabled
### Summary

`build_electron_args()` in `launcher-common.sh` appends `--no-sandbox` for the deb/nix package types (rpm uses the `deb` path) whenever the session is Wayland, regardless of the backend mode (default XWayland or native Wayland):

```bash
# Wayland: deb/nix packages need --no-sandbox in both modes
[[ $package_type == 'deb' || $package_type == 'nix' ]] \
    && electron_args+=('--no-sandbox')
```

(v3.2.1, `launcher-common.sh` ~line 326)

Since Wayland has been the default session on GNOME and KDE for years, this effectively disables the entire Chromium sandbox for most desktop Linux users of the deb/rpm/nix packages — even though those packages correctly ship `chrome-sandbox` as root-owned 4755.

### Where it came from

The flag was introduced in 3e437088 ("fix: Add Wayland compatibility mode to prevent crashes") together with `--disable-gpu` / `--disable-software-rasterizer`, to fix a "Trace/breakpoint trap (core dumped)" crash that the commit message itself attributes to GPU buffer (dmabuf/EGLImage) incompatibilities. I couldn't find any recorded failure implicating the sandbox itself — it looks like `--no-sandbox` rode along with the GPU fix and outlived it (the launcher has since gained the sticky GPU-FATAL recovery logic, which handles the GPU side properly).

Notably, #687 already solved the one *real* sandbox blocker on record — the Ubuntu 24.04+ `apparmor_restrict_unprivileged_userns` launch crash — the right way (a scoped AppArmor userns profile, sandbox stays on), but only the X11 path benefits: as that PR's own description puts it, *"`scripts/launcher-common.sh` adds `--no-sandbox` for AppImage and any Wayland session, so deb-on-Wayland never reaches this path."* The Wayland branch never got the same treatment.

### Evidence it isn't needed (at least not universally)

Fedora 44, GNOME on Wayland (default XWayland mode), `claude-desktop-unofficial` 1.19367.0-3.2.1 (rpm), NVIDIA proprietary driver:

- Launching the Electron binary directly with the launcher's exact switches **minus `--no-sandbox`** runs stable: the setuid broker process is present, renderers show `Seccomp: 2`, no zygote/sandbox errors in the log. Works both with GPU acceleration on and with `--disable-gpu`.
- Same machine through the stock launcher: renderers run with `Seccomp: 0`.

### Impact

- Loses Chromium's defense-in-depth for rendered web content on the most common desktop configuration.
- `--doctor` reports `[PASS] Chrome sandbox: permissions OK` on these systems, which is misleading — the check validates file permissions but the launcher then disables the sandbox at runtime. (Related to the doctor-check discussion in #785, which covers the AppImage side.)

### Suggestion

1. Drop the unconditional Wayland `--no-sandbox` for deb/rpm/nix (or gate it on an actual probed failure rather than "is Wayland") — the #687 AppArmor userns profile already covers the only documented real blocker, and it is not Wayland-specific;
2. or at minimum add an env escape hatch symmetrical to `CLAUDE_DISABLE_GPU`, e.g. `CLAUDE_FORCE_SANDBOX=1`;
3. have `--doctor` warn when the composed `electron_args` include `--no-sandbox`, so the sandbox status reported matches reality.

Happy to test patches on Fedora/Wayland (X11-via-XWayland and native) with and without NVIDIA.

chrisdebian:
Fixes #804.

deb/nix packages pass `--no-sandbox` unconditionally on Wayland, even though the AppArmor userns profile #687 installs (Ubuntu-family deb installs only) already covers the one documented real blocker and isn't Wayland-specific.

Dropping the flag by default across all three package types isn't safe without more work though: rpm reuses the deb argv path (`rpm.sh` calls `build_electron_args 'deb'`) but has no equivalent AppArmor profile installer, and neither does nix — so this only adds the two lower-risk, additive pieces of the reporter's three suggestions, leaving the default-behaviour question to the maintainer:

- **`CLAUDE_FORCE_SANDBOX=1`**, an opt-in env var (mirrors the existing `CLAUDE_DISABLE_GPU` pattern) that keeps the sandbox enabled on Wayland for deb/nix. AppImage is unaffected — its `--no-sandbox` is unconditional (FUSE), not part of this Wayland branch.
- **`doctor.sh` gains `_doctor_check_effective_sandbox`**, reporting whether `--no-sandbox` actually ended up in the composed launch args, alongside (not replacing) the existing file-permission check — the two can genuinely disagree, which was part of the original report (the doctor PASSes on permissions while the sandbox is disabled at launch). Built by calling the real `build_electron_args` for the package type rather than re-deriving the same rule a second time, matching this project's own stated concern about switch-list drift (`tools/chromium-switch-smoke.sh`).

`run_doctor` gains a second, optional `package_type` parameter (default `'deb'`, matching every existing call site except `appimage.sh`) so the new check knows which argv-building path to check. All three generated launcher wrappers (`deb.sh`, `rpm.sh`, `appimage.sh`) updated to pass their package type through.

## Testing

- 11 new bats cases covering the escape hatch (both package types it affects, appimage unaffected, default-unchanged) and the new doctor check (X11 pass, Wayland warn, override pass, the appimage no-op, the standalone-doctor.sh guard, and a `run_doctor` integration test) — all pass.
- `shellcheck` clean on every touched file.
- `tools/chromium-switch-smoke.sh` passes — the default switch list is unchanged, this is opt-in only.

---
90% AI / 10% Human
Claude: investigated the root cause (traced deb.sh's AppArmor postinst against rpm.sh/nix's lack of an equivalent), implemented the escape hatch and doctor check, wrote the tests.
Human: scoped which of the reporter's three suggestions were safe to act on without maintainer input, reviewed the diff, decided to submit.


sabiut (approved):
Reviewed with the full local recipe: suite, shellcheck, and per-piece mutation checks. Approving — the two suggestions below are non-blocking.

**Verified:**

- Full suite green: 118/118 in `launcher-common.bats`, zero failures across all 13 bats files.
- shellcheck clean on every touched `.sh` file.
- Mutation checks all catch: reverting the escape-hatch hunk turns the 3 hatch tests red; removing the `detect_display_backend` call from `run_doctor` turns the integration test red; removing the appimage early-return turns the no-op test red. The coverage is honest.
- The dynamic-scope plumbing is correct: `build_electron_args` assigns `electron_args=()` unscoped, so the doctor check's `local electron_args` captures it without leaking globally.
- Real-flow safety: `--doctor` runs before `setup_logging` in all three wrappers, and every side-effecting path in `build_electron_args` is guarded in that context (`log_message` no-ops without `$log_file`, `_previous_launch_hit_gpu_fatal` returns 1 on unset `log_file`, `WM_CLASS` is baked at build time). No stray output in the doctor flow.
- Driving the *real* `build_electron_args` instead of a re-derived stub is exactly right given the switch-list-drift history — a stub mirroring the prod rule couldn't catch a change to it.
- The rpm.sh `'deb'` literal is correct (rpm reuses the deb argv path) and well-commented.

**Suggestions (non-blocking):**

1. `rpm` is an API trap in `run_doctor`'s doc comment. It lists `deb|rpm|nix|appimage` as valid, but a future call site passing `'rpm'` would get `[PASS] enabled at runtime` — wrong, since real rpm installs take the deb argv path and disable the sandbox. Either normalize `rpm` → `deb` inside `_doctor_check_effective_sandbox` (which also makes the rpm.sh call-site comment unnecessary), or drop `rpm` from the documented values.
2. `_info 'File permissions above are correct'` is asserted unconditionally — if `_doctor_check_chrome_sandbox` actually FAILed just above, this line contradicts it. Neutral wording ("regardless of the file-permission result above") avoids the disagreement.

Informational, no change requested: the current nix derivation (`nix/claude-desktop.nix`) doesn't use the launcher and explicitly never passes `--no-sandbox`, so the `nix` branch of the hatch is unreachable in the shipped Nix package today. Harmless future-proofing given the derivation is still evolving, but the CHANGELOG's "deb/nix packages ... otherwise pass `--no-sandbox` unconditionally there" slightly overstates the nix side.

aaddrick:
Let's wrap in @sabiut's non-blockers and then I'll merge. 

Thanks everyone!

chrisdebian:
Pushed a follow-up commit addressing both non-blockers from @sabiut's review:

- `_doctor_check_effective_sandbox` now normalizes `rpm` → `deb` internally, so the doc comment's listing of `rpm` as a valid `package_type` is genuinely accurate rather than a latent trap for a future caller.
- Reworded the "File permissions above are correct" line to "Regardless of the file-permission result above", so it no longer contradicts a preceding FAIL.

Added one new bats case covering the normalization directly, mutation-checked (reverted the fix, confirmed the test goes red, restored it). Full suite: 320/320 passing.

PR #825, opened 2026-08-25
fix(deb): remove legacy bwrap AppArmor profile on upgrade

yannicklescure:
Fixes the upgrade path left open by the rename, reported in #542.

## Problem

2.x installs shipped an AppArmor profile attached to the shared `/usr/bin/bwrap`. It is no longer installed, but **nothing removes it on an upgrade**: dpkg runs the *old* package's `postrm` with `upgrade`, and that cleanup arm only matches `remove|purge|abort-install`:

```sh
case "$1" in
    remove|purge|abort-install)
        for _profile in "/etc/apparmor.d/$package_name" \
            "/etc/apparmor.d/${package_name}-bwrap" \
            "/etc/apparmor.d/claude-desktop-bwrap"; do
```

`postinst` writes the new scoped profile but never touches the legacy names, so the stale file survives indefinitely on an otherwise-fixed package.

## Why it matters

On Ubuntu 25.10+ this is not cosmetic. The distro ships `bwrap-userns-restrict`, also attached to `/usr/bin/bwrap`, so two profiles claim the same path. AppArmor resolves neither:

```
apparmor="AUDIT" operation="exec" info="conflicting profile attachments"
        profile="unconfined" name="/usr/bin/bwrap" comm="gly-hdl-loader"
apparmor="DENIED" operation="capable" profile="unprivileged_userns"
        comm="bwrap" capability=8 capname="setpcap"
```

bwrap falls through to `unprivileged_userns`, which denies `setpcap`/`net_admin`. GNOME 47+ decodes **every image** through glycin inside a bwrap sandbox, so image loading fails system-wide:

- **GDM greeter unusable** — the logo texture returns null, `add_child(null)` throws inside `LoginDialog._init` (`loginDialog.js:1086`), and the constructor aborts before building the user list or background. No way to log in.
- **`gnome-terminal-server` SIGABRTs** on a GTK assertion at `gtkiconhelper.c:495`, loading the fallback `image-missing.png`.
- **gnome-keyring cannot prompt** (`gcr-prompter` fails to render), breaking passphrase-protected ssh keys and Claude Desktop's own login.

Confirmed on `claude-desktop-unofficial 1.30096.1-3.2.2`, where the correctly scoped Electron profile from #542's fix and the stale bwrap profile coexisted — the scoped one was fine, the leftover one broke the desktop. 560 bwrap denials in one boot; 0 after removing and unloading it.

## Change

Clean up both legacy names from `postinst`, gated on the same marker header already used for the Electron profile, so admin-authored files are preserved.

`apparmor_parser -R` is required **in addition to** `rm`: deleting the file leaves the profile loaded in the kernel, so the conflict persists until the next reboot. That is easy to miss — after removing the file and running `systemctl reload apparmor`, denials continued and `aa-status` still listed the profile.

## Testing

- `bash -n scripts/packaging/deb.sh` — passes
- Generated `postinst` extracted with `$package_name` expanded and syntax-checked — passes
- Marker guard exercised against a fixture pair: the marked file is removed, an admin-authored one is preserved
- Not able to run `shellcheck` locally; the addition lives inside the `<< EOF` heredoc, which shellcheck treats as data, so CI should be unaffected

Worth noting @clayg's point in #542 still stands: because these profiles are written by maintainer scripts rather than shipped as tracked package files, `dpkg -S` cannot attribute them and nothing reconciles them across a rename. Shipping via `dh_apparmor` would make this class of bug structurally impossible; this patch just closes the current gap.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ The format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) — 
 
 <!-- Updated automatically by check-claude-version; will be current at release time. -->
 
+### Fixed
+
+- `postinst` now removes the legacy `claude-desktop-bwrap` AppArmor profile that 2.x installs left behind, so it no longer survives an upgrade. The profile attaches `flags=(unconfined)` to the shared `/usr/bin/bwrap`, a path Ubuntu's own `bwrap-userns-restrict` also claims; AppArmor resolves neither attachment, logs `conflicting profile attachments`, and bwrap falls through to `unprivileged_userns`, which denies `setpcap` and `net_admin`. On Ubuntu 26.04 that takes out glycin — the sandboxed decoder GNOME 47+ routes *every* image through — so the symptoms are a GDM greeter with no user list and no background that cannot be logged into at all, `gnome-terminal-server` aborting with a GTK assertion at `gtkiconhelper.c` while loading the fallback `image-missing.png`, blank icons across the shell, and `gcr-prompter` failing to render so gnome-keyring stops prompting for ssh key passphrases. The profile is inert on 24.04 (no glycin), which makes this present as "the release upgrade broke my desktop" with nothing pointing at this package — and the 24.04 → 26.04 LTS upgrade path is exactly the population carrying it. `postrm` already handled the legacy names, but its case arm only matches `remove|purge|abort-install` and dpkg invokes the *old* package's `postrm` with `upgrade`, so the file survived every package upgrade; both profiles were confirmed coexisting on `claude-desktop-unofficial 1.30096.1-3.2.2`. Removal is gated on the marker header so admin-authored files of the same name are preserved, and `apparmor_parser -R` runs alongside the delete because the profile otherwise stays loaded in the kernel until the next reload. ([#542](https://github.com/aaddrick/claude-desktop-debian/issues/542), [#825](https://github.com/aaddrick/claude-desktop-debian/pull/825))
+
 ## [v3.2.4] — 2026-09-06
 
 ### Fixed
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -423,9 +423,15 @@ run with `--no-sandbox`, see
 ["AppImage Sandbox Warning"](#appimage-sandbox-warning) — so there's
 nothing AppImage-specific to add there either.)
 
-No package ships a bwrap profile as of v3.0.0+; the deb's `postrm` still
-removes the 2.x-era `/etc/apparmor.d/claude-desktop-bwrap` leftover (and a
-`claude-desktop-unofficial-bwrap` sibling, if one exists) on purge.
+No package ships a bwrap profile as of v3.0.0+. The deb's `postrm` removes
+the 2.x-era `/etc/apparmor.d/claude-desktop-bwrap` leftover (and a
+`claude-desktop-unofficial-bwrap` sibling, if one exists) on purge, and
+since [#825](https://github.com/aaddrick/claude-desktop-debian/pull/825)
+`postinst` also clears it on upgrade — `postrm` alone never fired on that
+path, so the leftover used to survive indefinitely. See
+["Blank icons / unloggable GDM greeter after upgrading to Ubuntu
+26.04"](#blank-icons--unloggable-gdm-greeter-after-upgrading-to-ubuntu-2604)
+if you are already in that state.
 
 **Credit:** [@hfyeh](https://github.com/hfyeh)
 ([#351](https://github.com/aaddrick/claude-desktop-debian/issues/351)) for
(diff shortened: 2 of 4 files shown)


yannicklescure:
@aaddrick — flagging timing on this one, since I think the window to land it is narrower than it looks.

Ubuntu 26.04.1 is out, and by Ubuntu's normal policy the 24.04 LTS → 26.04 LTS upgrade path opens with the first point release. Reporting I've seen says that migration is currently delayed rather than live, which is why I'm raising this now — the wave hasn't started yet.

That wave is precisely the affected population. A 24.04 user who installed claude-desktop during the 2.x era has `/etc/apparmor.d/claude-desktop-bwrap` on disk, and on 24.04 it is completely inert: GNOME 46 doesn't route image decoding through bwrap, so nothing collides and there is no symptom. The moment they upgrade to 26.04, glycin arrives, the profile starts colliding with the distro's `bwrap-userns-restrict`, and the desktop becomes unloggable.

My own box is a clean data point for that latency: the profile is dated 26 June, ran without a single symptom for seven weeks on 24.04, and broke on 16 August the moment the release upgrade completed. Nothing about the package changed in between — only the distro's image-loading path.

Two things make this worse than a normal regression:

1. **The failure mode is a locked-out desktop, not a broken app.** GDM renders no user list and no background, so there is no way to log in graphically. Recovery requires knowing to switch to a TTY, knowing to look at AppArmor, and knowing that `apparmor_parser -R` is needed on top of deleting the file. That's a lot to ask of someone whose machine just stopped booting to a desktop.

2. **Nothing points at this package.** The symptom appears immediately after an Ubuntu release upgrade, so it reads as "the upgrade broke my desktop." In my case it sent me through librsvg, shared-mime-info, the gdk-pixbuf loader cache, and the icon themes — all of which test clean, because the old in-process gdk-pixbuf API never touches glycin. Claude Desktop was the last thing I'd have suspected, and I only found it by chasing `conflicting profile attachments` in the audit log.

Upgrading the package doesn't save them either — that's the gap #825 closes. `postinst` writes the correctly-scoped profile but never removes the legacy name, and the old package's `postrm` cleanup arm only matches `remove|purge|abort-install`, so it never fires on `upgrade`. I confirmed both profiles coexisting on `claude-desktop-unofficial 1.30096.1-3.2.2`.

The patch is deliberately small — `postinst` only, +25/−0, marker-guarded with the same rule already used for the Electron profile, so admin-authored files are preserved. Happy to rework it however you'd prefer: move it into `preinst`, fold it into the existing loop, or restructure it entirely if you'd rather solve this via `dh_apparmor` and tracked files, which would make this whole class of bug structurally impossible (per @clayg's point in #542).

And if shipping cleanup in maintainer scripts isn't the direction you want at all, even a short entry in `docs/troubleshooting.md` naming `claude-desktop-bwrap` and the `apparmor_parser -R` step would help — it would at least be searchable for anyone who lands on a greeter with no user list after upgrading. Right now there's nothing to find.

No urgency on my own account, to be clear — my machine is fixed and I'm not blocked. I'd just rather this didn't land on a pile of people during the LTS migration.

CherrieTheShifter:

yannicklescure:
@sabiut — I see you're proposed as project lead in #836, so flagging this one ahead of that landing. It's small but time-sensitive, and it sits in `scripts/packaging/deb.sh`, which would be default-owned by you once that merges.

**What it does:** adds a cleanup loop to `postinst` that removes the legacy `claude-desktop-bwrap` AppArmor profile left behind by 2.x installs. `postinst` only, +25/−0, gated on the same marker header already used for the Electron profile so admin-authored files are preserved.

**The gap it closes:** the `postrm` cleanup already handles the legacy names, but its case arm only matches `remove|purge|abort-install`. dpkg runs the *old* package's `postrm` with `upgrade`, so it never fires on an upgrade — the stale profile survives onto current versions. I confirmed both profiles coexisting on `claude-desktop-unofficial 1.30096.1-3.2.2`.

**Why the timing matters:** the leftover profile is inert on 24.04 (no glycin), and fatal on 26.04, where it collides with the distro's `bwrap-userns-restrict` and takes out all glycin image decoding — unloggable GDM greeter, GTK apps aborting on icon loads, gnome-keyring unable to prompt. The 24.04 → 26.04 LTS upgrade path is the exact population carrying that profile. Full detail in #542.

You know this surface better than I do — #781 and #782 are yours. Happy to rework the patch however you'd prefer, move it to `preinst`, fold it into the existing loop, or drop it entirely in favour of shipping the profiles as tracked files via `dh_apparmor`, which would make this class of bug structurally impossible. A `docs/troubleshooting.md` entry naming the profile and the `apparmor_parser -R` step would also help on its own, since there is currently nothing searchable for someone staring at a greeter with no user list.

No rush on my account — my machine is fixed. Rebasing or splitting it is no trouble if that's easier for you.

sabiut (commented):
Verified this end to end. On main, the postrm case only matches `remove|purge|abort-install`, and the file is only deleted on purge. During an upgrade, dpkg invokes the old postrm with `upgrade`, so the legacy profile survives every package upgrade.

The v2.0.22 postinst created the bwrap profile with the header `managed by the claude-desktop package (postinst)`. The prefix check therefore identifies both the original legacy profile and its renamed variant without claiming unrelated files.

I rendered the postinst from this branch and executed it under bwrap against a fake `/etc/apparmor.d`, using a stub `apparmor_parser`. The results were correct:

- a legacy profile carrying our marker was unloaded and removed;
- an administrator-authored file with the same filename was preserved; and
- the distro-provided `bwrap-userns-restrict` profile beside it was untouched.

shellcheck is clean for `deb.sh`, and the rendered script passes `sh -n`.

Two small additions are needed before merge. Both are fine as additional commits on this branch:

1. Add a `CHANGELOG.md` entry under `[Unreleased]` → Fixed. Name the searchable symptom: the leftover `claude-desktop-bwrap` profile conflicts with `bwrap-userns-restrict` on Ubuntu 26.04 and breaks glycin image decoding.
2. Add a regression check to `tests/test-artifact-deb.sh`: extract the control archive with `dpkg-deb -e`, assert that postinst contains the legacy-profile cleanup loop, and validate it with `sh -n`. Checking presence and syntax in every package build is the practical baseline; the Ubuntu 24.04 runner prevents the full behavioural test because it blocks the required user namespace under bwrap.

The proposed troubleshooting entry naming the profile and the `apparmor_parser -R` recovery step would also be useful. I'm happy for that to land here or in a follow-up.

yannicklescure:
Both additions are in `bdf0410`, plus the troubleshooting entry since you said it could land here.

**CHANGELOG** — `[Unreleased]` → `Fixed`. Leads with the searchable symptoms rather than the mechanism: greeter with no user list or background, `gnome-terminal-server` aborting on the `gtkiconhelper.c` assertion while loading `image-missing.png`, blank icons, and `gcr-prompter` failing so gnome-keyring stops prompting. Names the `postrm` `upgrade` gap and why `apparmor_parser -R` is needed alongside the delete.

**Regression check** — in `tests/test-artifact-deb.sh`, placed in the metadata section before install since it only inspects the control archive. `dpkg-deb -e` into a temp dir, then three assertions: postinst references `claude-desktop-bwrap`, postinst runs `apparmor_parser -R`, and postinst passes `sh -n`. I added the `apparmor_parser -R` assertion beyond what you asked — the delete without the unload leaves the profile live in the kernel, so a future refactor that dropped it would still pass a presence-only check while leaving the bug in place on every already-broken machine.

**Troubleshooting** — new symptom-first section, "Blank icons / unloggable GDM greeter after upgrading to Ubuntu 26.04", with a TTY confirmation step (`journalctl -b | grep -c 'conflicting profile attachments'`) and the recovery commands. Written to be findable by someone who does not yet suspect this package, since the failure presents as an Ubuntu upgrade regression.

I also corrected the existing bwrap paragraph. It said `postrm` removes the leftover "on purge", which is accurate but reads as though the leftover is handled — and that path never fires on an upgrade, which is the entire bug. It now notes the `postinst` cleanup and links the new section.

Two notes:

- The `([#542])` suffix is deliberately out of the new heading: GitHub folds it into the generated anchor, which would have broken the in-page link. Issue link sits in the body instead.
- I could not run `shellcheck` locally (not installed here). `bash -n` passes on the test file, and the block is fully quoted with no unused variables, so I expect it clean — but that is reasoning, not a result. CI is the check.

Happy to split the docs commit out if you would rather keep the PR to code and tests.

yannicklescure:
Rebased onto `main` — the branch had gone to CONFLICTING after `0a9dedd` promoted `[Unreleased]` to v3.2.4. Now `69ba8eb`, mergeable again.

The only conflict was `CHANGELOG.md`, and the resolution is a judgement call worth stating: my entry stayed under the new empty `[Unreleased]` rather than moving into the v3.2.4 section, since the fix is not in that release. v3.2.4 and v3.2.3 are untouched.

Re-verified against the new base:

- `shellcheck -x` over the workflow's own file set (24 files) — clean, exit 0. That closes the caveat from my last comment; I have shellcheck locally now, so it was actually run rather than reasoned about.
- `bash -n` on `scripts/packaging/deb.sh` and `tests/test-artifact-deb.sh`
- the rendered postinst extracted with `$package_name` expanded, `sh -n` clean, and the cleanup loop plus `apparmor_parser -R` both present in the rendered output
- the `docs/troubleshooting.md` anchor still resolves to its heading after the rebase

One thing you may want to action: all three workflow runs on this PR are sitting at `action_required` (Shellcheck, BATS Tests, CI), so nothing has actually run in CI since the branch opened — GitHub is holding them behind first-time-contributor approval. My local results above are not a substitute for that.

sabiut (approved):
Both additions verified. Rendered the postinst from main and from this branch: the legacy-profile mention and the `apparmor_parser -R` call both read zero on main and are present here, so all three asserts are load-bearing. The extra unload assert is the right call. Changelog placement under the new `[Unreleased]` is correct, the troubleshooting anchor resolves, and the corrected postrm paragraph now says why the purge path never covered upgrades. Merging once the artifact-validation legs finish; they are the ones that run the new check against a real built deb.

Issue #709, opened 2026-06-09
Quit cleanup: MCP-process matching needs DE-agnostic scope evidence before it ships


aaddrick:
#682 originally included pgrep arms matching MCP-ish cmdlines (`-mcp`, `@modelcontextprotocol/`), fenced behind a systemd-scope gate. Review found the fence never fires. The gate greps for a literal `app-claude-desktop-*.scope`, but GNOME prefixes its units with `app-gnome-`, and systemd escapes the hyphens, so KDE renders the unit as `app-claude\x2ddesktop...` with `.service` variants for menu launches. Dead code on both desktops. I carved the MCP arms out of #682 rather than ship them ineffective.

This issue tracks what's needed before the slice comes back.

**1. Real scope captures from GNOME and KDE.** We need `/proc/PID/cgroup` or `systemd-cgls` output from actual claude-desktop launches on both desktops, menu-launch and terminal-launch. The matcher gets derived from those captures and pinned in BATS. No more guessed regexes: every pattern proposed so far has failed against systemd's actual escaping.

**2. A concrete repro.** Nobody has shown a non-cowork helper surviving an explicit quit on current main. The cowork daemon is handled: main's reaper catches it at next launch, and #693 added coverage for exactly that path. Quit cleanly, then show `pgrep -af` output of a helper still alive. If nothing survives, the MCP slice may be unnecessary and this issue closes itself.

**3. The risk to design against.** Matching `-mcp` or `@modelcontextprotocol/` without a working scope fence can kill a user's own MCP dev servers, like ones they run for Claude Code in a terminal. The original design was safe because the gate was an AND. But a dead gate means an untested gate, and a "fixed" gate that's wrong in the other direction turns a silent no-op into killed bystander processes. The captures from item 1 are what make the fence trustworthy.

---
Written by Claude Fable 5 via [Claude Code](https://claude.ai/code)

luohoa97:
[...]

PR #828, opened 2026-08-31
fix: reap chrome_crashpad_handler on quit so AppImage teardown can't SIGBUS it

abguymon:
No issue filed — small launcher fix, per CONTRIBUTING. Refs [#709](https://github.com/aaddrick/claude-desktop-debian/issues/709) and [`docs/learnings/quit-cleanup-scope-fence.md`](docs/learnings/quit-cleanup-scope-fence.md), which this supplies the missing evidence for.

## The bug

Every AppImage quit leaves a `SIGBUS` core dump behind:

```
Signal: 7 (BUS) si_code: BUS_ADRERR
Command Line: /tmp/.mount_claudentgET1/usr/lib/claude-desktop/chrome_crashpad_handler
              --monitor-self-annotation=ptype=crashpad-handler --no-rate-limit
              --database=/home/…/.config/Claude/Crashpad --url=https://f.a.k/e …
```

Journal, same second:

```
12:27:14 app-com.anthropic.Claude-4082638.scope: Consumed 15min 51s CPU over 4h 44min
12:27:14 Process 4082664 (chrome_crashpad) terminated abnormally with signal 7/BUS
12:27:14 tmp-.mount_claudentgET1.mount: Deactivated successfully
```

18 such cores on this machine going back to June, one per quit. The app itself exits cleanly; only the helper dies, so nothing is user-visible beyond the core dumps and any "application crashed" desktop notification the DE raises.

## Mechanism

`chrome_crashpad_handler` is built to outlive the browser it monitors — that is its job. Its text pages are `mmap`'d from `/tmp/.mount_claude*`. The AppImage runtime unmounts that squashfuse mount as soon as `AppRun` exits, so the still-live handler faults on its next instruction: `SIGBUS`/`BUS_ADRERR`, the signature of a file-backed mapping whose backing store went away.

The stack shows it dying twice over — crashpad's own handler catches the first fault and immediately re-faults at the identical address, because the handler code lives on the same vanished mapping:

```
#0  handler + 0x1a364d   ← crashpad's SIGBUS handler
#4  libc.so.6 + 0x3e6f0  ← signal return
#5  handler + 0x1a364d   ← original fault, same address
#6..#14 → __libc_start_main → main
```

## Why the existing reaper misses it

`run_electron_and_cleanup` already keeps `AppRun` alive precisely to reap helpers that outlive the main process, and `cleanup_stale_desktop_helpers` already SIGTERMs, waits 2s, then SIGKILLs — all before `AppRun` returns, i.e. before the unmount. The machinery to prevent this crash is already here.

Crashpad just slips through the filter. `_desktop_helper_candidate_pids` **does** list it (its `argv[0]` is under `/usr/lib/claude-desktop/`), but `_desktop_helper_cmdline_matches` then rejects it, because every Chromium arm requires a literal `--type=`:

```bash
*/usr/lib/claude-desktop/*--type=*)
```

Renderers, GPU and utility processes all carry `--type=renderer` and friends. Crashpad carries `--monitor-self-annotation=ptype=crashpad-handler` — that substring is `=ptype=`, not `--type=`, so the glob does not match. The one helper designed to outlive the browser is the one helper the outlive-reaper misses.

## Why this wasn't caught before

`quit-cleanup-scope-fence.md` concludes:

> the desktop-helper reaper has no demonstrated survivor to catch… Until a real survivor surfaces, the MCP slice stays out

That testing ran against a **deb** install (`/usr/lib/claude-desktop-unofficial/`). On deb a surviving crashpad is genuinely harmless — the tree stays on disk, nothing faults. The survivor only becomes fatal on AppImage, where the mount disappears underneath it. So the doc's conclusion was correct for what it tested; this is the AppImage-only corner it didn't cover.

## The fix

Two arms added to `_desktop_helper_cmdline_matches`, handing crashpad to the cleanup path that already exists. Deb/rpm behaviour is unchanged in effect (a lingering crashpad there is reaped rather than left, which is hygiene, not a fix).

The path prefix keeps the arm scoped to our tree — a bystander `/usr/lib/chromium/chrome_crashpad_handler` or Slack's must not match, and doesn't.

## Verification

`bats` isn't available on this host, so alongside the added bats cases I exercised the real function by sourcing it directly, against the verbatim cmdline from the core above:

```
BEFORE (upstream HEAD)                  AFTER
  FAIL  REAL crashpad cmdline             ok [match] REAL crashpad cmdline
  FAIL  unofficial-tree crashpad          ok [match] unofficial-tree crashpad
  ok    chromium's crashpad (miss)        ok [miss]  chromium's crashpad
  ok    slack's crashpad (miss)           ok [miss]  slack's crashpad
  ok    renderer --type= (match)          ok [match] renderer --type=
  ok    cowork helper (match)             ok [match] cowork helper
  ok    main UI not a helper (miss)       ok [miss]  main UI not a helper
  ok    user's own claude CLI (miss)      ok [miss]  user's own claude CLI
  ok    user's own MCP server (miss)      ok [miss]  user's own MCP server
```

- `shellcheck` 0.11.0 on `scripts/launcher-common.sh`: no new findings vs. HEAD (the only diff is a line-number shift on the pre-existing SC1091 info).
- `bash -n` clean; tabs, ≤80 cols with tab=8.
- The shipped `launcher-common.sh` inside my installed `3.2.1+claude1.24012.9` AppImage is byte-identical to HEAD's, so this reproduces on current main.

**Not run:** `./build.sh --build appimage --clean no`, and the bats suite itself. Worth a maintainer run before merge.

## Known limits (not introduced here)

- `cleanup_stale_desktop_helpers` bails entirely if any Claude UI is alive, so quitting one of two concurrent instances still won't reap that instance's crashpad. Pre-existing, and documented in-tree as "Safe, not complete."
- SIGTERM could in principle truncate an in-flight minidump. Low risk: reports are local-only (`uploadToServer:false`, `--url=https://f.a.k/e`), and SIGTERM gets a 2s grace before SIGKILL.

Environment: Arch (Omarchy), Hyprland/Wayland, systemd 258, `claude-desktop-appimage 3.2.1+claude1.24012.9-1`.

---
90% AI / 10% Human
Claude: diagnosed the crash from the core dump and journal, traced it to the `--type=` glob, wrote the patch, the bats cases and this description, ran shellcheck and the direct-source verification.
Human: hit the bug, provided the machine and its 18 core dumps, directed the investigation, reviewed and approved submitting.

### DIFF
--- a/scripts/launcher-common.sh
+++ b/scripts/launcher-common.sh
@@ -488,36 +488,31 @@ _desktop_helper_cmdline_matches() {
 	local cmdline="$1"
 	local config_dir="${XDG_CONFIG_HOME:-$HOME/.config}/Claude"
 
-	case "$cmdline" in
-		*cowork-vm-service.js*)
-			return 0
-			;;
+	# /proc/PID/cmdline is NUL-joined and rendered with tr, so argv[0] is
+	# everything up to the first space. Every arm that identifies a
+	# process by the *binary it is running* tests argv[0], never the whole
+	# string: a bystander that merely names one of our binaries as an
+	# argument -- `gdb .../chrome_crashpad_handler core`, `coredumpctl
+	# gdb`, an editor, a `less` -- must not land in the reaper's kill
+	# list. That scenario is not hypothetical: this reaper's own bug
+	# leaves crashpad cores lying around to be debugged.
+	#
+	# Argv[0] carrying a space degrades to a miss, never a false match.
+	# Our install prefixes and the AppImage mount point have none.
+	local argv0="${cmdline%% *}"
+
+	case "$argv0" in
 		*cowork-linux-helper*)
 			# Official Rust Cowork helper, spawned via
-			# process.resourcesPath (relocation-safe, so no fixed path).
-			return 0
-			;;
-		*"--user-data-dir=$config_dir "*)
-			return 0
-			;;
-		*"$config_dir/Claude Extensions/"*)
-			return 0
-			;;
-		*/usr/lib/claude-desktop/*--type=*)
-			return 0
-			;;
-		*/usr/lib/claude-desktop-unofficial/*--type=*)
-			# Phase 3 package rename, landing in v3.0.0: our package
-			# installs to /usr/lib/claude-desktop-unofficial while the
-			# official arm above keeps matching Anthropic's install
-			# (and the AppImage internal tree).
+			# process.resourcesPath (relocation-safe, so no
+			# fixed path).
 			return 0
 			;;
 		*/usr/lib/claude-desktop/*chrome_crashpad_handler*|\
 		*/usr/lib/claude-desktop-unofficial/*chrome_crashpad_handler*)
 			# Crashpad is built to outlive the browser it
-			# monitors, so it is the survivor the --type= arms
-			# above cannot catch: it is spawned with
+			# monitors, so it is the survivor the --type= arm
+			# below cannot catch: it is spawned with
 			# --monitor-self-annotation=ptype=crashpad-handler
 			# and carries no --type= switch at all.
 			#
@@ -531,6 +526,40 @@ _desktop_helper_cmdline_matches() {
 			# The path prefix keeps this scoped to our tree: a
 			# bystander /usr/lib/chromium/chrome_crashpad_handler
 			# must not match.
+			return 0
+			;;
+	esac
+
+	# Chromium's own helpers: an in-tree argv[0] *and* a --type=
+	# switch. Both halves are load-bearing -- the tree alone would
+	# also match resources/chrome-native-host, which Chrome (not
+	# Claude Desktop) spawns and owns the lifetime of.
+	#
+	# The claude-desktop-unofficial prefix is our package, renamed in
+	# v3.0.0; the claude-desktop one keeps matching Anthropic's own
+	# install and the AppImage internal tree.
+	case "$argv0" in
+		*/usr/lib/claude-desktop/*|\
+		*/usr/lib/claude-desktop-unofficial/*)
+			case "$cmdline" in
+				*--type=*)
+					return 0
+					;;
+			esac
+			;;
+	esac
+
+	# Argument-shaped arms: these identify a helper by what it was
+	# handed, not by which binary runs it (argv[0] here is node, or
+	# the Electron main). They stay whole-cmdline by necessity.
+	case "$cmdline" in
+		*cowork-vm-service.js*)
+			return 0
+			;;
+		*"--user-data-dir=$config_dir "*)
+			return 0
+			;;
+		*"$config_dir/Claude Extensions/"*)
 			return 0
 			;;
 	esac
--- a/scripts/launcher-common.sh
+++ b/scripts/launcher-common.sh
@@ -513,6 +513,26 @@ _desktop_helper_cmdline_matches() {
 			# (and the AppImage internal tree).
 			return 0
 			;;
+		*/usr/lib/claude-desktop/*chrome_crashpad_handler*|\
+		*/usr/lib/claude-desktop-unofficial/*chrome_crashpad_handler*)
+			# Crashpad is built to outlive the browser it
+			# monitors, so it is the survivor the --type= arms
+			# above cannot catch: it is spawned with
+			# --monitor-self-annotation=ptype=crashpad-handler
+			# and carries no --type= switch at all.
+			#
+			# Harmless on deb/rpm, fatal on AppImage. The
+			# runtime unmounts /tmp/.mount_claude* as soon as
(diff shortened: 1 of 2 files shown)


sabiut (approved):
Reviewed with the full local recipe — running the parts the PR noted as not-run on the author's host. Approving; the observations below are non-blocking.

**Verified:**

- Full bats suite green: zero failures across all 13 test files, including the new asserts (they land inside the existing `_desktop_helper_cmdline_matches: matches known Desktop helpers only` umbrella test — consistent with the file's style, though the whole scenario reports as one test).
- shellcheck clean on `scripts/launcher-common.sh`.
- Mutation check: reverting the launcher hunk while keeping the tests turns the umbrella test red — the coverage is real.
- Mechanism claims verified in-tree: the `--type=` glob genuinely misses `--monitor-self-annotation=ptype=crashpad-handler` (the substring is `=ptype=`); crashpad is a `_desktop_helper_candidate_pids` candidate via the `/usr/lib/claude-desktop(-unofficial)?/` pgrep arm; cmdline is composed with `tr '\0' ' '`, so the fixtures' trailing spaces mirror the real string; and `cleanup_stale_desktop_helpers` runs from `cleanup_after_electron_exit`, after Electron exits but before AppRun returns — before the unmount, which is the window the fix needs.
- Bystander scoping holds: the chromium-crashpad miss is pinned by a test, and the path prefix keeps Slack/chromium out.

**Non-blocking observations:**

1. **The arm is a substring match, not argv[0]-anchored** — like the `--type=` arms above it, it matches the path + binary name *anywhere* in a candidate's cmdline. The concrete new exposure: someone debugging one of these very core dumps (`gdb /tmp/.mount_…/usr/lib/claude-desktop/chrome_crashpad_handler core`, or `coredumpctl gdb`) has a cmdline that both passes the candidate pgrep and matches the new arm, so their debugger gets SIGTERM'd at the next launch or quit. This is the pre-existing pkill-broadness class (same thread as the #671 smoke-test follow-ups), and crashpad just makes the scenario likelier since this PR's own bug leaves cores lying around to debug. Fine to defer to a follow-up that anchors all the arms to argv[0] at once rather than special-casing this one.
2. **Official-install visibility**: `_claude_desktop_ui_is_alive` keys on `--class=$WM_CLASS`, which only our launcher injects — an official Anthropic deb running concurrently is invisible to the gate, and the arms (by documented design, including the pre-existing `--type=` ones) match its tree. The new arm extends that existing behavior consistently rather than adding a new class of risk; noting it here so the follow-up in observation 1 considers both.

CherrieTheShifter:

aaddrick:
Sorry errbody, been wrapped up in a project at work. Reviewing and merging shortly

aaddrick:
Thanks for the forensics. Tracing this to crashpad's text pages going away under the unmount was the hard part.

I pushed a commit to your branch (`a690c9c`) taking @sabiut's argv[0] point, but across all the arms rather than just crashpad: binary-shaped arms now test argv[0], argument-shaped ones stay whole-cmdline. That caught a real bystander on my box: `resources/chrome-native-host` is in our tree but owned by Chrome, and it stays out now. Then I rebased onto main for the #829 anchor fix. Your commit is `5b4739a`, authorship intact, diff unchanged. CI needs re-approval on the fork after the force-push.

Tests and shellcheck are green, and a real AppImage launches and quits clean. But I can't reproduce your SIGBUS here (this box has never logged a crashpad core), so I can't prove the fix kills it.

Can you build the rebased branch and confirm the cores stop, and that nothing you use got reaped? Happy to merge on that.

---

CochainComplex:
Heads-up from a neighboring PR: #833 touches the same file pair (`launcher-common.sh` + `launcher-common.bats`) and extracts a shared `_kill_pids_escalating` helper (SIGTERM → poll 20×0.1s → SIGKILL, with the standard log-string format) out of `cleanup_stale_desktop_helpers`. Merge-tested both orders against this branch: zero conflicts either way, so no coordination needed — but if #833 lands first, the crashpad reap here could reuse that helper instead of carrying its own escalation loop. Purely optional; mentioning it so the option is visible rather than discovered post-merge.

---

sabiut:
@abguymon, picking this up from @aaddrick now that I'm leading the project. I reviewed his follow-up commit a690c9c: the argv[0] anchoring across all the binary-shaped arms is what I was after, and the chrome-native-host case is a good catch. Suite is 119/119 on the rebased branch, shellcheck clean, and reverting either the argv[0] anchor or the `--type=` half turns the helper-matching test red, so both halves are pinned.

The one thing still open is his ask above. Can you build the rebased branch on the machine that logged the crashpad cores and confirm two things: the cores stop on quit, and nothing you use gets reaped at launch or quit? The unit tests prove the matcher, not the SIGBUS. Once you confirm, I'll merge.

CI is green on the rebased head.

PR #832, opened 2026-09-04
feat(launcher): kill replaced Claude Desktop UI before relaunch; fix(tests): resolve a real asar in the patch-stage harness

CochainComplex:
Two independent changes; happy to split into separate PRs if you'd prefer.

## 1. `feat(launcher)`: kill a replaced Claude Desktop UI before relaunch

When `dpkg`/`rpm` replaces the Electron binary underneath a running
instance, the old process keeps its deleted executable **and** Electron's
single-instance lock. The next launch silently loses the lock and appears
to do nothing — no window, no error.

Adds `cleanup_replaced_desktop_ui` to `launcher-common.sh`, wired into the
deb, rpm and AppImage launchers. It matches a live UI on the `--class`
fingerprint, checks whether `/proc/PID/exe` reads `(deleted)`, and
terminates only that process — `SIGTERM`, escalating to `SIGKILL` if it
survives. The `(deleted)` check is what keeps it from touching a
legitimately running instance.

Covered by new `launcher-common.bats` cases.

Also carries local WIP that is **inert in the current build**:

- `patches/config.sh`: parked `patch_asar_trusted_folder_guard` and
  `patch_asar_additional_dirs_guard` (`config.sh` stays sourced-but-not-wired).
- `cowork-fallback/cowork.sh`: treat YukonSilver as supported on Linux so the
  bwrap fallback backend doesn't delete the VM bundle before the local daemon
  starts (fallback path, not wired into the build).
- `docs/learnings`: update the cowork-vm-daemon and patching-minified-js notes.

## 2. `fix(tests)`: resolve a real asar in the patch-stage harness

`tests/test-patch-stage.sh` invoked `"$asar_exec"` as a single word, so its
fallback could never work:

```bash
asar_exec=$(command -v asar || command -v npx)
```

The first call site is `"$asar_exec" extract-file "$asar_path" package.json`,
which under the fallback becomes `npx extract-file app.asar package.json`.
**npx reads `extract-file` as a package name**, fetches an unrelated
`extract-file@1.0.0` from the registry and runs it. `package.json` is never
extracted, `desktopName` comes back empty, and the WM_CLASS tripwire fails
the build on garbage input rather than on anything real.

Net effect: on any machine without `asar` on `PATH`, the suite reports a
patch failure that does not exist — which reads like "the patches are
broken" to anyone running it for the first time.

Replaced with `_resolve_asar`: prefer a real `asar`; else `npm install
@electron/asar` into `$work_dir` the way `scripts/setup/dependencies.sh`
already does for the build; else fail loud. No `npx` path survives.

### Verification

Against the pinned **1.40609.1** official `.deb`:

| scenario | result |
|---|---|
| `asar` on PATH | used directly, no stray local install |
| `asar` absent | installs to `$work_dir`, resolves v3.4.1 |
| absent + npm broken | `rc=1`, loud message, no silent fallback |
| mutation check | restoring the old line re-selects `/usr/bin/npx` |

End-to-end with `asar` absent from `PATH`: patch stage exits 0, all 160 JS
files parse, 5 markers survive the repack, second pass byte-identical.

The full patch stage is also green on `1.40609.1` with all five active
patches binding (`quick_window` → `jU`, `org_plugins_path`,
`virtiofsd_probe`, `cowork_bwrap` A/B/C1, `tray_icon_env_override`), AU-1 /
MB-1 / WM_CLASS tripwires clear. `patch-anchors.bats` + `app-asar.bats` (39)
and `launcher-common.bats` (120) pass. `shellcheck` clean.

The pre-existing `C2: warm download anchor not present in this bundle`
warning is unchanged — C2 is best-effort by design.

## 3. Portability audit follow-ups (`0e0a723`, `1567f6b`)

The branch was audited for anything that only holds on the development
machine. Two things came out of it:

**`readlink -f` was a works-on-one-layout special case.** The replaced-UI
detection read `/proc/PID/exe` with `readlink -f`, which canonicalizes and
therefore requires every parent directory of the link target to still
exist. That's true for a plain file replace (dpkg/rpm upgrading in the same
layout — the shape the feature was developed against) but false when the
whole install directory goes away, e.g. a package migration such as this
repo's own `claude-desktop` → `claude-desktop-unofficial` transition. On
that shape `readlink -f` exits nonzero and the stale UI silently keeps the
single-instance lock — the exact bug the cleanup exists to fix. Verified
empirically both ways; plain `readlink` returns the raw link content with
the ` (deleted)` marker in both shapes. A new mutation-checked bats case
removes the whole install directory before the cleanup runs: restoring
`readlink -f` turns it red while the file-only-deleted case stays green.

Both test spawns also close fd 3 in the child (bats' run-coordination
pipe) — without that, a failure mode that leaves the blocked child alive
hangs the entire bats run instead of reporting red. And `_proc_state` now
applies `2>/dev/null` before the input redirect, so a PID vanishing
mid-scan no longer prints to the launcher log.

**Dedup refactor.** `cleanup_replaced_desktop_ui` had arrived carrying
verbatim copies of the SIGTERM→poll→SIGKILL escalation and the
fingerprint PID scan that already existed in the file; both are now shared
helpers (`_kill_pids_escalating`, `_claude_desktop_ui_pids`) with log
strings unchanged. `cleanup_orphaned_cowork_daemon` was deliberately left
out: its poll predicate is a `pgrep` re-scan rather than `kill -0` on
captured PIDs, and a reaper's poll predicate must equal its kill
predicate.

Nothing in the diff references machine-specific paths, users, or hosts;
`WM_CLASS` remains the build-time `@@WM_CLASS@@` substitution.
`tests/launcher-common.bats`: 121/121.

### DIFF
--- a/docs/learnings/cowork-vm-daemon.md
+++ b/docs/learnings/cowork-vm-daemon.md
@@ -91,13 +91,20 @@ even after rolling back the AppImage to a known-good version.
 
 ### Fix (extend delete list — Patch 6b)
 
-`scripts/patches/cowork.sh` now matches the `const NAME=["rootfs.img",...]` array at
-module level and appends `"sessiondata.img"` and `"rootfs.img.zst"` if
-they're not already present. The auto-reinstall path now wipes these
-too. Trade-off: the next successful startup re-downloads/re-extracts
-these files. Acceptable because auto-reinstall only runs after startup
-has already failed — biasing toward recovery over re-download
-avoidance is correct.
+`scripts/patches/cowork.sh` supports both upstream delete-list shapes:
+the older `const NAME=["rootfs.img",...]` array and the newer generated
+list returned by `deleteVMBundle()`. It ensures the list includes
+`rootfs.img`, its origin marker, `sessiondata.img`, `rootfs.img.zst`,
+and the compressed-cache origin marker. The auto-reinstall path now
+wipes these too. Trade-off: the next successful startup
+re-downloads/re-extracts these files. Acceptable because
+auto-reinstall only runs after startup has already failed — biasing
+toward recovery over re-download avoidance is correct.
+
+As of upstream `1.12603.1`, VM downloads create `.wvm-tmp-*` directly
+under the bundle directory. Patch 8 detects that shape and reports it
+as already fixed instead of warning that the older
+`mkdtemp(join(tmpdir(), "wvm-"))` anchor is missing.
 
 Not included in the delete list: `~/.config/Claude/claude-code-vm/`.
 That's CLI-binary storage (`2.1.x/claude`), unrelated to the VM
--- a/docs/learnings/patching-minified-js.md
+++ b/docs/learnings/patching-minified-js.md
@@ -242,11 +242,13 @@ in v3.0.0, lessons stand):
   The rewrite anchors on `"pop-up-menu"` (`quick-window.sh:17`), the
   `isWindowFocused` property name (`quick-window.sh:60`), and the
   `[QuickEntry]` log strings (`quick-window.sh:88-91`).
-- **Cowork spawn (PR #436).** Anchored on `,VAR.mountConda)`
-  (`cowork.sh:741`) — unique to the 12-arg call path, absent from the
-  10-arg one-shot. Asserts match count is exactly 1 and bails
-  otherwise (`cowork.sh:744`), so a future second caller surfaces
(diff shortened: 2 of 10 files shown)


CochainComplex:
Closing to re-file split into focused PRs: the launcher fix and the test-harness fix separately, without the unrelated parked WIP this branch carried.

Issue #812, opened 2026-07-23
[bug]: empty window on start after update to 1.24012.0-3.2.1 (regression of #600)


oldgranola:
### Version (`claude-desktop-unofficial --doctor` output)

```shell
$ claude-desktop-unofficial --doctor
Claude Desktop Diagnostics
================================

[PASS] Installed version: 1.24012.0-3.2.1
[PASS] Version: in sync with the official pool (1.24012.0)
[PASS] Display server: X11 (DISPLAY=:0)
[PASS] Electron: v42.7.0 (/usr/lib/claude-desktop-unofficial/claude-desktop)
[PASS] Chrome sandbox: permissions OK (/usr/lib/claude-desktop-unofficial/chrome-sandbox)
[PASS] SingletonLock: held by running process (PID 6005)
       Password store: upstream os_crypt autodetect (default)
[PASS] Keyring: org.freedesktop.secrets reachable for credential encryption
       Tray icon: Cinnamon dark panel auto-detected (CLAUDE_TRAY_USE_DARK_ICON=1, TrayIconLinux-Dark.png)
[PASS] MCP config: valid JSON (/home/john/.config/Claude/claude_desktop_config.json)
       MCP servers configured: 12
[PASS] Node.js: v24.13.1
       Path: /home/john/.nvm/versions/node/v24.13.1/bin/node
[PASS] Desktop entry: /usr/share/applications/claude-desktop-unofficial.desktop
[PASS] Disk space: 641648MB free

Cowork Mode
----------------
[PASS] KVM: /dev/kvm present and accessible
[PASS] vsock: /dev/vhost-vsock present
[PASS] QEMU: qemu-system-x86_64 found
[PASS] Firmware: /usr/share/OVMF/OVMF_CODE_4M.fd
[PASS] virtiofsd: /usr/libexec/virtiofsd (client-probed path)
       Cowork isolation: KVM (official)

       COWORK_VM_BACKEND=bwrap: Cowork routes through the bubblewrap fallback daemon (opt-in).
[PASS] bwrap daemon runtime: node v24.13.1 (/home/john/.nvm/versions/node/v24.13.1/bin/node)
[PASS] bwrap daemon: cowork-vm-service.js present
[PASS] bubblewrap: found
[PASS] bubblewrap: sandbox probe succeeded
[PASS] Cowork bwrap daemon: running (parent alive)
[PASS] Log file: 184KB (/home/john/.cache/claude-desktop-debian/launcher.log)

[...]

PR #833, opened 2026-09-04
feat(launcher): kill replaced Claude Desktop UI before relaunch

CochainComplex:
When dpkg/rpm replaces the Electron binary underneath a running instance,
the old process keeps its deleted executable **and** Electron's
single-instance lock. The next launch silently loses the lock and appears
to do nothing — no window, no error.

`cleanup_replaced_desktop_ui` (wired into all three launchers; a no-op on
AppImage by construction — the binary lives inside the FUSE mount, which
stays valid after the `.AppImage` file is replaced, so the marker never
appears there — it only ever acts on deb and rpm installs) finds a live UI matched on the `--class` fingerprint whose
`/proc/PID/exe` carries the kernel's ` (deleted)` marker and terminates
only that process — SIGTERM, escalating to SIGKILL if it survives. The
`(deleted)` check is what keeps it from ever touching a legitimately
running instance.

Design notes:

- **Plain `readlink`, not `readlink -f`.** `-f` canonicalizes, so it fails
  when the install *directory* is gone too (a package migration such as
  `claude-desktop` → `claude-desktop-unofficial`, or an rpm layout change,
  rather than a same-layout file replace) — and the replaced UI would be
  silently missed. The raw link content carries the marker in both shapes;
  verified empirically both ways.
- **Shared helpers instead of copy-paste.** `_kill_pids_escalating` does
  the SIGTERM → poll 20×0.1s → SIGKILL escalation for this cleanup and for
  `cleanup_stale_desktop_helpers`, which previously spelled it out inline;
  `_claude_desktop_ui_pids` is the fingerprint scan that both this cleanup
  and `_claude_desktop_ui_is_alive` layer their own predicate onto.
  `cleanup_orphaned_cowork_daemon` is deliberately *not* folded in: its
  poll predicate is a `pgrep` re-scan rather than `kill -0` on captured
  PIDs, and a reaper's poll predicate must equal its kill predicate.
- `_proc_state` replaces the previous `awk '/^State:/ …'` parse with a
  pure-bash read loop and applies `2>/dev/null` before the input redirect,
  so a PID vanishing mid-scan can't leak "No such file or directory" into
  the launcher log.

Tests (`launcher-common.bats`, 122/122) pin the predicate from both sides.
A **negative** case spawns a fingerprinted process whose binary is intact
and asserts it survives with no "Killed replaced" log line (the log file is
touched first so the assertion cannot pass vacuously on a missing file) —
replacing `_claude_desktop_ui_is_replaced` with an unconditional `return 0`
turns only it red, and `return 1` turns only the two positive cases red. A
mutation-checked positive case removes the whole install directory before
the cleanup runs — restoring `readlink -f` turns it red while the
file-only-deleted case stays green. Test spawns close fd 3 in the child (bats' run-coordination
pipe), so a failure mode that leaves the blocked child alive reports red
instead of hanging the run. `shellcheck` clean.

Replaces the launcher half of #832, re-filed without the unrelated parked
WIP that branch carried. Companion: the patch-stage harness fix is its own
PR.

**Artifact-level coverage (second commit).** The deb and rpm artifact tests
now drive the *installed* launcher: a fingerprinted stand-in whose executable
has been unlinked (the ` (deleted)` marker a real `dpkg`/`rpm` upgrade
produces) must be killed and logged; an intact one must survive with no kill
logged. Keyed on the WM_CLASS baked into the installed `launcher-common.sh`,
needs no Xvfb (the cleanup runs before the display check), drops to the
throwaway user on the rpm leg. Verified: CI's deb leg verbatim on
`ubuntu:24.04` (48 passed, 0 failed) and the rpm leg's `runuser` privilege
shape on `fedora:42`, each with the mutation flipping exactly its own pair.
No AppImage leg, per the no-op above.

Refs #812 — possibly, not certainly: that report's signature (silent exit 0
with nothing logged, beginning immediately after a dpkg update, doctor showing
`SingletonLock: held by running process`) matches the stale-replaced-instance
mechanism this PR fixes, though the reporter's `CLAUDE_DISABLE_GPU=1`
workaround points at the GPU theory instead. A diagnostic to distinguish the
two is posted on the issue.

Includes the Keep-a-Changelog `[Unreleased]` entry for this change.

---
50% AI / 50% Human
Claude: dedup refactor (`_kill_pids_escalating`, `_claude_desktop_ui_pids`, `_proc_state`), the plain-`readlink` portability fix with its mutation-checked install-dir-gone test, fd-3 test hygiene, CHANGELOG entry, cross-distro verification (7 systems), PR split and description.
Human: original `cleanup_replaced_desktop_ui` implementation, initial bats coverage, packaging wiring, direction and scope decisions throughout.


sabiut (changes requested):
Ran the suite on the branch: 121/121, shellcheck clean. The shared reaper is a faithful lift of the old inline escalation, and the launcher ordering is right (replaced UI first, so the helper reaper and stale-lock cleanup that follow see a dead UI).

One blocking gap. The description says the `(deleted)` check "is what keeps it from ever touching a legitimately running instance", but nothing pins that. I replaced the check in `_claude_desktop_ui_is_replaced` with an unconditional `return 0` and all 121 tests stayed green. Both new cases are positive. Add the negative one: spawn a fingerprinted process whose binary is intact, run `cleanup_replaced_desktop_ui`, assert it survives and the log has no "Killed replaced" line. That's the test that makes the safety claim true.

Two nits:

- The `_proc_state` comment justifies pure bash with "the launchers run before any PATH sanity check". There isn't one, and `_claude_desktop_ui_pids` already leans on `pgrep`, `tr`, and `id` from PATH. Keep the bash version if you like, but drop the rationale.
- On AppImage this is a no-op by construction: the binary is inside the FUSE mount, which stays valid after the `.AppImage` file is replaced, so `/proc/PID/exe` never carries the marker. Wiring it in is harmless, but the changelog line shouldn't imply AppImage is covered.

Also needs a rebase: `CHANGELOG.md` conflicts with `main` after the v3.2.4 promotion.

CochainComplex:
Thanks — all four points addressed in `77e9e0d` (rebased onto v3.2.4; CHANGELOG resolved).

**Blocking gap — negative test added.** `cleanup_replaced_desktop_ui: leaves an intact-executable UI alone` spawns a fingerprinted process whose binary stays on disk, runs the cleanup, and asserts the process survives and the log has no "Killed replaced" line. Two details to make it non-vacuous: the log file is `touch`ed after `setup_logging` (which only creates the directory), and the assertion requires `grep` exit **1** specifically, so a missing file (exit 2) can't pass by accident. Verified with your exact mutation and its inverse:

| mutation in `_claude_desktop_ui_is_replaced` | deleted-exe | intact-exe (new) | dir-gone |
|---|---|---|---|
| fix in place | ok | ok | ok |
| `return 0` | ok | **not ok** | ok |
| `return 1` | **not ok** | ok | **not ok** |

So the predicate is pinned from both directions now, not just the positive side.

**`_proc_state` rationale** — you're right, there's no such check and the scan already leans on `pgrep`/`tr`/`id`. Kept the bash version, dropped the rationale from the code comment, commit message, and PR body.

**AppImage** — agreed it's a no-op by construction (FUSE mount stays valid after the file is replaced, so the marker never appears). Wiring left in place as harmless; the CHANGELOG line, the function's comment, and the PR body now say explicitly that it only ever acts on deb/rpm.

**Rebase** — done, single commit on top of `c37cd9c`. 122/122, shellcheck clean.

CochainComplex:
Deployment check on a clean `debian:13` container, exercising the **installed** package rather than the bats fixtures — `77e9e0d` built into a `.deb` (Node 22 toolchain, per the #839 decision), installed with `apt`, then the real upgrade sequence as a non-root user:

```
fake UI pid=17122 exe=/usr/lib/claude-desktop-unofficial/claude-desktop
dpkg -i <same .deb>            # upgrade underneath the running instance
exe now: /usr/lib/claude-desktop-unofficial/claude-desktop (deleted)
relaunch via /usr/bin/claude-desktop-unofficial
launcher.log: Killed replaced Claude Desktop UI (PIDs: 17122)
old pid 17122 gone (PASS)

# production negative: intact instance, relaunch again
intact pid=17235 exe=/usr/lib/claude-desktop-unofficial/claude-desktop
intact pid 17235 survived (PASS)
Killed-replaced lines before=1 after=1 (PASS: no new kill line)
```

One honesty note: the running instance is a long-lived stand-in *at the installed binary path*, not Electron — Chromium's zygote refuses to start under Docker's default seccomp (`Failed to move to new namespace … Operation not permitted`). Everything the launcher actually keys on is real: the `--class` fingerprint via `pgrep`, the kernel's ` (deleted)` marker produced by a genuine `dpkg` replacement, and the SIGTERM→SIGKILL path in the shipped `launcher-common.sh`. `--doctor` on the installed package is clean apart from container-expected KVM/vsock warnings.

CochainComplex:
Added a second commit, `49cfb30`, on top of the review fixes: artifact-level coverage. The deb and rpm artifact tests now drive the **installed** launcher against a fingerprinted stand-in whose executable has been unlinked (the same ` (deleted)` marker a real `dpkg`/`rpm` upgrade leaves) and assert it is killed and logged, then against an intact stand-in and assert it survives with nothing logged. Keyed on the WM_CLASS baked into the installed `launcher-common.sh`; no Xvfb needed since the cleanup runs before the display check; drops to the throwaway user on the rpm leg. No AppImage leg, per the no-op.

Verified before pushing: CI's deb leg verbatim on `ubuntu:24.04` (build `77e9e0d`, then `tests/test-artifact-deb.sh`: 48 passed, 0 failed, the four new checks included) and the rpm leg's `runuser` privilege shape on `fedora:42`, each with a `return 0` / `return 1` mutation of `_claude_desktop_ui_is_replaced` flipping exactly its own pair. Kept as a separate commit so it's easy to drop if you'd rather keep this PR to the launcher change.

sabiut (approved):
All four points verified on the branch. Both guard mutations land where your table says: `return 0` reddens only the new intact-executable test, `return 1` reddens only the two positive ones. 122/122, full suite green, shellcheck clean.

The artifact commit is worth keeping. I traced the installed launcher's flow: the cleanup runs, then `check_display` exits before anything Electron-shaped, so the no-Xvfb claim holds, and the XDG overrides keep the config and cache out of the host's. Thanks for the honesty note on the Docker stand-in; the parts the launcher keys on are the real ones.

Approving. `CHANGELOG.md` conflicts with `main` again after #825 landed under `[Unreleased]`; one more rebase and I'll merge.

CochainComplex:
Rebased onto current `main` (`d2d7f6b`, after #840 and #825): head is now `8d0baaa`, still two commits. The `[Unreleased]` collision is resolved with all three entries under the one `### Fixed` heading. Re-verified on the rebased head with the CI toolchain (Ubuntu 24.04, Node 22, apt bats/shellcheck): shellcheck 0, 487 bats ok, `.deb` build 0, `test-artifact-deb.sh` 52 passed / 0 failed (#825's four postinst checks now run alongside the replaced-UI ones). Ready to merge.

PR #834, opened 2026-09-04
fix(tests): resolve a real asar in the patch-stage harness

CochainComplex:
`tests/test-patch-stage.sh` resolved its tooling as

```bash
asar_exec=$(command -v asar || command -v npx)
```

but invokes `"$asar_exec"` as a single word, so the fallback could never
work. The first call site is
`"$asar_exec" extract-file "$asar_path" package.json`, which under the
fallback becomes `npx extract-file app.asar package.json` — **npx reads
`extract-file` as a package name**, fetches an unrelated
`extract-file@1.0.0` from the registry and runs it. `package.json` is
never extracted, `desktopName` comes back empty, and the WM_CLASS
tripwire fails the build on garbage input rather than on anything real.

Net effect: on any machine without `asar` on `PATH`, the suite reports a
patch failure that does not exist — which reads as "the patches are
broken" to anyone running it for the first time.

`_resolve_asar` replaces it: prefer a real `asar` on `PATH`; else
`npm install @electron/asar@3` into `$work_dir`; else fail loud. No `npx`
path survives.

Two further traps close in the same helper, both of which otherwise surface
as that same phantom WM_CLASS failure (#839 tracks the build's own copy):

- **The major is pinned.** `@electron/asar` 4.x requires Node ≥ 22.12 and
  refuses to run (`CANNOT RUN WITH NODE 20.x`) on the Node 20 that Debian 13
  and other stable distros ship. Whether an unpinned install lands on 4.x
  depends on the npm version — npm 9 takes `latest` (4.3.0); npm 10's
  manifest picker skips engine-incompatible versions and falls back to
  3.4.1, which is why the earlier description reported 3.4.1 — so
  resolution can't be trusted across hosts.
- **`-x` is not "can run".** A 4.x binary under Node 20 is executable and
  still can't run, so the helper probes `--version` and stops with a message
  naming the Node requirement instead of letting a non-runnable binary fall
  through.

`--silent` is dropped so npm's `EBADENGINE` warning stays visible.

Verified on Node 20.20.2 / npm 10.8.2:

| scenario | result |
|---|---|
| `asar` on PATH | used directly, no stray local install |
| `asar` absent | installs `@electron/asar@3`, resolves v3.4.1, probe passes |
| **4.x `asar` first on PATH** | `rc=1`, "cannot run under node v20.20.2" + the `@3` / Node ≥ 22.12 hint |
| absent + npm broken | `rc=1`, loud message, no silent fallback |
| mutation: old line restored | re-selects `/usr/bin/npx` |
| **mutation: `--version` probe removed** | the 4.x binary passes `-x` (`rc=0`) |

End-to-end with `asar` absent from `PATH` against the pinned official `.deb`
(1.46388.2): patch stage exits 0, every JS file parses, all markers survive
the repack, second pass byte-identical. `shellcheck` clean.

Split out of #832; companion to the launcher PR.

Includes the Keep-a-Changelog `[Unreleased]` entry for this change.

---
90% AI / 10% Human
Claude: diagnosis of the `npx extract-file` package-name misresolution, the `_resolve_asar` fix, all four verification scenarios including the mutation check and the no-asar end-to-end run, CHANGELOG entry.
Human: surfaced the failing environment, reviewed and directed.


sabiut (changes requested):
I reproduced the issue on a Debian 13 host with no `asar` executable on `PATH`. The old command resolves to `/usr/bin/npx`, which downloads `extract-file@1.0.0`. The stage then fails on the empty `desktopName` tripwire. Good catch.

However, the proposed replacement also fails on the same host. `npm install @electron/asar` installs version 4.3.0, and all 4.x releases require Node.js 22.12.0 or newer. Debian 13 provides Node.js 20.19.2, so the installed binary reports:

`CANNOT RUN WITH NODE 20.19.2`

The harness then reaches the same `WM_CLASS` phantom failure. The `-x` check only confirms that the file exists and is executable; it does not confirm that the tool can actually run. Additionally, `--silent` hides npm's `EBADENGINE` warning, which would make the cause clear.

I could not reproduce the description's claim that npm resolves version 3.4.1. With npm 9 and the default configuration, it installs 4.3.0.

When I put `@electron/asar@3` (currently 3.4.1) on `PATH`, the harness passes every check:

* 160 JS files parse successfully
* All 5 markers are found
* The second pass leaves the asar byte-identical

This confirms that the resolver logic is correct; the problem is the version of `asar` it resolves to.

Two changes before this can land:

1. Verify that the resolved tool can run, not merely that it exists. For example:

   `"$asar_exec" --version >/dev/null`

   This exits 1 on 4.3.0 under Node 20 and 0 on 3.4.1. If it fails, stop with a clear error explaining the required Node.js version.

2. Do not install `@electron/asar` without pinning its major version. Either use the last Node.js 20-compatible release line, `@electron/asar@3`, or resolve Node.js in the same way as the build.

The build has the same latent issue: `setup_asar` installs an unpinned version, while `setup_nodejs` allows Node.js 20 and its fallback downloads 20.18.1. I'll open a separate issue for that so this PR can remain focused.

This PR also needs to be rebased because `CHANGELOG.md` now conflicts with `main` following the v3.2.4 promotion.

sabiut:
Follow-up for the build-side Node floor is #839.

CochainComplex:
Thanks — reproduced your finding exactly and fixed it in `9d9873d` (rebased onto v3.2.4; CHANGELOG resolved). Point by point:

**1. Verify the tool can run, not merely exists** — done. After the `-x` check, `_resolve_asar` runs `"$asar_exec" --version`; on failure it stops with the binary's own output plus a message naming the requirement:

```
asar at …/node_modules/.bin/asar cannot run under node v20.19.2:
  CANNOT RUN WITH NODE 20.19.2
  asar requires Node >=22.12.0.
  @electron/asar 4.x needs Node >= 22.12; use @electron/asar@3 on older Node, or a newer Node (#839).
```

**2. Pin the major** — done: `npm install … @electron/asar@3` (the Node-20-compatible line; `engines.node >=10.12.0`). I went with the pin rather than resolving Node the build's way, since the harness's job is to exercise the patch stage, not to own a Node bootstrap — and #839 is the right place for the build-side floor.

**3. `--silent`** — dropped, so `EBADENGINE` stays visible (kept `--no-audit --no-fund` to keep the noise to the warning that matters).

**On the 3.4.1 claim you couldn't reproduce** — it was real but host-dependent, which is itself the argument for the pin. `latest` is 4.3.0. npm **9.2** (Debian 13) takes `latest` → 4.3.0. npm **10.8** (my host) has a manifest picker that skips engine-incompatible versions and falls back to 3.4.1. So the same unpinned command yields different majors on different hosts. Description and CHANGELOG now say so explicitly.

**Verified on your exact host shape** — a clean `debian:13` container with distro `nodejs` 20.19.2 / npm 9.2.0 and no `asar` on PATH, on this branch:

```
=== unpinned (your repro) ===
npm WARN EBADENGINE   package: '@electron/asar@4.3.0',
CANNOT RUN WITH NODE 20.19.2

=== fixed harness, same host ===
No asar on PATH; installing @electron/asar@3 locally...
Using asar executable: …/node_modules/.bin/asar (v3.4.1)
  [PASS] patch stage exits 0
  parsed 181 JS files
  [PASS] all 181 JS files parse
  [PASS] marker present: ×5
  [PASS] second pass exits 0
  [PASS] asar byte-identical after re-run
All patch-stage checks passed
```

Plus, on Node 20.20 / npm 10.8: a 4.x binary placed first on PATH is rejected (`rc=1`) with the message above; and the mutation that removes the `--version` probe lets that same 4.x binary pass `-x` (`rc=0`) — so the probe is load-bearing, not decoration. `shellcheck` clean.

CochainComplex:
Alignment note re #839 (option 1, Node ≥ 22.12 floor): once that lands, the `@electron/asar@3` pin here can be dropped in favour of the same floor so the harness exercises the exact asar the build uses. Kept the pin for now because the harness has to work on today's `main`, where the build itself still fails on distro Node 20 — verified: `./build.sh --build deb` on a clean `debian:13` hits `CANNOT RUN WITH NODE 20.19.2` at the patch stage, and succeeds with a Node 22 toolchain.

sabiut (approved):
Both asks verified on a Debian 13 host with Node 20.19.2 and no `asar` on PATH. Against the pinned 1.46388.2 deb the harness installs 3.4.1, parses all 181 JS files, keeps every marker, and the second pass is byte-identical. Unpinning brings 4.3.0 back and the resolver now stops with the binary's own refusal and the #839 pointer; unpinning *and* removing the `--version` probe brings back the phantom WM_CLASS failure, so both the pin and the probe are load-bearing. The npm 9 vs npm 10 explanation for the 3.4.1 result closes the last open question. shellcheck clean.

Approving. `CHANGELOG.md` conflicts with `main` again after #825 landed under `[Unreleased]`; one more rebase and I'll merge.

CochainComplex:
Rebased onto current `main` (`d2d7f6b`, after #840 and #825): head is now `81d08dd`, single commit. `[Unreleased]` collision resolved, all entries under the one `### Fixed` heading. Re-verified on the rebased head with the CI toolchain (Ubuntu 24.04, Node 22): shellcheck 0; harness with no `asar` on PATH installs 3.4.1 and passes every check. Ready to merge.

CochainComplex:
Rebased onto `main` after #833 merged (`68369a1`): head is now `15dfd27`, single commit. The `[Unreleased]` overlap with #833's entry is resolved — all four bullets under the one `### Fixed` heading. `tests/test-patch-stage.sh` is byte-identical to the approved `9d9873d`; only the CHANGELOG merge changed. shellcheck clean. Ready to merge.

PR #841, opened 2026-09-07
fix(build): raise the Node floor to @electron/asar's engine constraint

### LINKED ISSUE #839: build.sh patch stage fails on Node.js 20 hosts: unpinned @electron/asar resolves 4.x, which requires Node >= 22.12
## Summary

Any local build on a host whose `node` is older than 22.12.0 fails at the asar patch stage. `setup_asar` installs `@electron/asar` with no version constraint, npm resolves the latest 4.x, and every 4.x release refuses to run on Node < 22.12.0. The failure surfaces as the `WM_CLASS` tripwire, not as a Node error, so it reads like a broken patch.

CI is unaffected because `ubuntu-latest` ships Node 22.23.2. Debian 13 stable ships Node 20.19.2, so this is the default experience for anyone building locally on Debian.

## Reproduction

On Debian 13 with the distro `nodejs` (v20.19.2) and no `asar` on `PATH`:

```
$ npm install --no-save @electron/asar
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: '@electron/asar@4.3.0',
npm WARN EBADENGINE   required: { node: '>=22.12.0' },
npm WARN EBADENGINE   current: { node: 'v20.19.2', npm: '9.2.0' }
npm WARN EBADENGINE }
added 8 packages

$ ./node_modules/.bin/asar --version
CANNOT RUN WITH NODE 20.19.2
asar requires Node >=22.12.0.
```

The patch stage then reports:

```
Error: package.json desktopName is missing/empty — cannot derive WM_CLASS. Upstream moved the field Chromium reads the window class from; re-verify before shipping (#779).
```

Found while reviewing #834, whose new `_resolve_asar` in the test harness hits the same wall. That PR is being fixed on its own; this issue covers the build.

## Where it goes wrong

Three places agree on a floor that `@electron/asar` no longer honours:

- `scripts/setup/dependencies.sh:120` accepts any system Node with major >= 20.
- `scripts/setup/dependencies.sh:153` downloads Node 20.18.1 as the fallback when the system Node is missing or too old, so the fallback path is broken too.
- `scripts/setup/dependencies.sh:224` runs `npm install --no-save @electron/asar` with no version pin.
- `docs/building.md:39` documents the requirement as "Node.js v20+".

Every `@electron/asar` 4.x release, from 4.0.0 (2025-05-14) through 4.3.0 (2026-08-18), declares `engines.node >= 22.12.0`. The last release that runs on Node 20 is 3.4.1.

## Options

1. **Raise the floor to match the tool.** Require Node >= 22.12 at `dependencies.sh:120`, bump the fallback download to a current 22.x LTS, and update `docs/building.md`. Keeps the build on current asar.
2. **Pin asar to the Node-20-compatible line.** `npm install --no-save @electron/asar@3`. Smallest change, but freezes the build on a release line that will stop getting fixes.

Either way, `setup_asar` should verify the tool runs after install (`"$asar_exec" --version`), not only that the bin exists. Right now the only signal is an npm warning that scrolls past.

## Decision

Going with option 1. CI already builds on Node 22, so raising the floor matches what ships. The fallback download bump needs a fresh SHA pin, so that lands as its own PR.

yannicklescure:
Fixes #839. Takes option 1 from the issue — raise the Node floor to match the tool — since CI already builds on Node 22 and 3.4.1 is the last `@electron/asar` release that runs on Node 20.

## The failure

Any local build on a host with Node older than v22.12 dies at the asar patch stage. `setup_nodejs` accepted any major >= 20 and downloaded v20.18.1 when it had to supply Node itself, but every `@electron/asar` 4.x release (4.0.0 onward) declares `engines.node >=22.12.0` and hard-refuses to start below it. npm reports that as an `EBADENGINE` **warning**, not an error, so the wrapper installs, lands in `node_modules/.bin/asar`, and then fails on every invocation.

Because the first thing the patch stage asks asar for is `package.json`, the operator sees:

```
Error: package.json desktopName is missing/empty — cannot derive WM_CLASS.
```

which reads as "upstream moved a field", not "your toolchain is broken". Debian 13 stable ships Node 20.19.2, so this is the default experience for anyone building locally there. CI is unaffected — `ubuntu-latest` runs 22.23.2.

## What changed

**`scripts/setup/dependencies.sh`**

- `NODE_MIN_VERSION='22.12.0'`, and the gate compares major **and** minor. A major-only test waves through the entire 22.0–22.11 band, which is below the engine floor.
- `NODE_FALLBACK_VERSION='22.23.2'` — the version CI already builds on, so the fallback path and the CI path agree.
- `setup_asar` now runs `asar --version` before handing the executable on, rather than trusting that the file exists.

**`.claude/hooks/install-build-tools.sh`** — same floor, reading the minor for the same reason, installing from NodeSource `setup_22.x`.

**Docs** — `docs/building.md` requirement table, the `setup-build-tools` skill table, and a `CHANGELOG.md` entry under `[Unreleased]` → Fixed.

## Two judgement calls worth flagging

**The fallback bump is in this PR, not a follow-up.** The issue said it needed its own PR because it "needs a fresh SHA pin". It doesn't — `setup_nodejs` already resolves the checksum from that release's own `SHASUMS256.txt` at build time (`scripts/setup/dependencies.sh`, the `shasums_url` block), so there is no hand-maintained SHA to refresh. Splitting it would have shipped a floor of 22.12 next to a fallback that downloads 20.18.1, i.e. a fallback path its own gate rejects.

**The run-check judges the exit code and the shape of stdout, and reports the merged streams.** On 4.3.0 under Node 20.19.2 the refusal exits 1 with an empty stdout and its complaint on stderr, so the exit code catches the shape that ships today. The stdout-shape arm is defense in depth for a future release that answers through `--version` and exits zero — an exit code is not a contract, and that shape would otherwise hand the patch stage a dead binary. Judging stdout rather than the merged stream is load-bearing either way: the refusal text is `CANNOT RUN WITH NODE 20.19.2`, which *contains* a version number, so a "does the output look like a version" match on merged output would pass the exact string the check exists to catch. So: judge `stdout`, anchor the match at the start of the reply, and show the operator the merged output when it fails. The failure message names the Node floor and the Node actually in use, so the diagnosis stops at the asar stage instead of at a patch anchor three stages later.

(Earlier revisions of this PR asserted the refusal path "prints its complaint through `--version` and is free to exit zero" as observed behaviour. That was wrong — @sabiut checked the real 4.3.0. The check is unchanged; the comment, the test comments and the changelog now describe the stdout arm as a forward guard rather than as something asar is seen to do.)

## Verification

- `tests/dependencies.bats`, 17 tests, added to the existing `tests/*.bats` glob in `tests.yml` — no workflow change needed.
- Full suite: 512 pass, 0 fail (`bats tests/*.bats scripts/cowork-fallback/tests/*.bats`).
- `shellcheck -x` over the workflow's own file set — clean, exit 0.

Mutation-checked, since a test that stays green when the fix is reverted is decoration. Ten mutations, each reddening only what it should:

| Mutation | Goes red |
|---|---|
| Collapse the comparison to major-only | `the 22.0-22.11 band is rejected`, `a bare major equal to the floor's` |
| Absent minor reads as 99 instead of 0 | `a bare major equal to the floor's` |
| Relax `NODE_MIN_VERSION` to 22.0.0 | `the floor is the asar engine constraint`, + 2 band tests |
| `NODE_FALLBACK_VERSION` back to 20.18.1 | `the shipped fallback download satisfies the floor` |
| Drop the `have` half of the unparseable-input guard | `input that arithmetic would clear the floor is rejected` |
| Drop the `want` half of the unparseable-input guard | `unparseable input is rejected` |
| Drop the exit-code arm (stdout shape only) | `a version-shaped reply on a non-zero exit still fails` |
| Drop the stdout-shape arm (exit code only) | `a refusal that exits zero would still fail the build`, `the same refusal on stderr` |
| Judge the merged stream instead of stdout | `noise on stderr does not fail a working asar` |
| Unanchor the version regex | `a refusal that exits zero would still fail the build` |

## Review round (thanks @sabiut)

Three things, all addressed in `9920bda` plus a rebase:

**The negated assertions could not fail.** `! cmd` is exempt from errexit, so a non-final `! cmd` line in a bats body reports nothing — only the last line is load-bearing. Every negated assertion now carries `|| return 1`, finals included, so appending a line to one of these bodies later can't silently disarm what's above it.

**Fixing that exposed a real gap rather than closing one.** Dropping the `have` half of the unparseable-input guard *still* left the suite green afterwards, and not because of the `!` problem: all four inputs the test used (`''`, `v22.12.0`, `nightly`, and an unparseable `want`) evaluate to 0 in a bare `(( ))`, so they are rejected with or without the regex — no input distinguished the guard from its absence. Added `node version: input that arithmetic would clear the floor is rejected`, using `0x18` (24 in base 16) and `22+1` (an expression, not a version). That mutation now goes red.

**The "exits zero" claim was wrong.** Corrected in the code comment, the test comments and the changelog, as described above. The check itself is unchanged, and `setup_asar: an asar that exits non-zero fails the build` now stubs the real shape (exit 1, empty stdout, complaint on stderr) rather than a generic failure.

**Rebased** onto `main` — the `CHANGELOG.md` conflict was #825 landing under `[Unreleased]`. Resolved by keeping every bullet from `main` and placing this one next to the `tests/test-patch-stage.sh` entry, which is the other half of the same #839 trap.

## Left out on purpose

Three callers hit the same wall from outside `build.sh`. Keeping them out so this stays on the build path — happy to take any of them as follow-ups:

- `tools/patch-necessity-audit.sh:74` and `tests/test-artifact-common.sh:139` — unpinned `npx --yes @electron/asar`, same silent-refusal behaviour, but they run outside `setup_nodejs`'s reach so raising the floor doesn't cover them.
- `.github/workflows/ci.yml:142` — the release job's `Set up Node.js 20` feeds the `npm install -g @electron/asar` two steps below it, so `Generate and upload reference source` gets an asar that won't run. Left alone because #814 is still open and rewriting that exact block. Worth a one-line `node-version: "22"` there whenever #814 lands.

## Commits

- `c206f0f` the fix, tests and docs.
- `15e91e8` a no-behaviour-change simplification pass over the same code: `node_version_at_least` splits the versions into arrays so `${_[1]:-0}` carries the absent-minor case instead of two identical branches, the 131-column failure message wraps, and `install_node`'s check folds into one arithmetic context.
- `9920bda` the review round above.

---
70% AI / 30% Human
Claude: diagnosed the engine-floor mismatch against the four cited call sites, wrote the version gate, the asar run-check, the bats suite and the mutation harness, ran the mutation pass and caught the surviving cases, drafted docs and CHANGELOG, handled the review round and the rebase.
Human: directed the issue selection, set the scope boundary against #814, called the fallback-bump split judgement, reviewed and approved the diff.


CochainComplex:
Cross-reference: #834 fixes the same trap in the test harness (`tests/test-patch-stage.sh`) and currently pins `@electron/asar@3` so it works on today's `main`. Merge-tested against this branch: zero conflicts either order. Once this lands, #834's pin can be dropped in favour of the same floor so the harness exercises the exact asar the build uses — I'll follow up there after this merges. Independent evidence for the failure this fixes: a clean `debian:13` container with distro Node 20.19.2 reproduces `CANNOT RUN WITH NODE 20.19.2` at the patch stage on `main`, and the same build succeeds with a Node 22 toolchain.

---

sabiut (changes requested):
Good PR, and you were right about the fallback bump: the checksum is resolved from `SHASUMS256.txt` at build time, so there was nothing to pin by hand, and splitting it would have shipped a fallback its own gate rejects. The issue text was wrong on that point. Both v22.23.2 tarballs and the checksum file are on nodejs.org.

Ran the suite on the branch: 16/16 and the full suite green. Dropping the run-check reddens 7 tests; collapsing the comparison to major-only reddens the two you named.

One mutation survives, and it's the test file rather than the guard. Drop only the `have` half of the unparseable-input guard and everything stays green. `node version: the 22.0-22.11 band is rejected` and `node version: unparseable input is rejected` stack negated assertions, and in bats a non-final `! cmd` line cannot fail the test: `! true` followed by `true` reports ok (three-line repro below). Only the last line in each of those tests is load-bearing. Guard each with `|| return 1`; I checked that form does fail.

```
@test "non-final negated assert" { ! true; true; }              # ok
@test "guarded"                  { ! true || return 1; true; }  # not ok
```

One wording fix. The comment and the changelog say the refusal "prints its complaint through `--version` and is free to exit zero". On the real 4.3.0 under Node 20.19.2 it exits 1 with an empty stdout and the message on stderr. Keep the stdout-shape check as defense in depth, but phrase it as a guard against a future change rather than as observed behavior.

Also needs a rebase: `CHANGELOG.md` conflicts with `main` after #825 landed under `[Unreleased]`.

yannicklescure:
Thanks — all three addressed in `9920bda`, plus the rebase.

**The `!` finding was right, and it was hiding a real gap rather than just decoration.** Every negated assertion now carries `|| return 1`, finals included, so a line appended to one of these bodies later can't silently disarm what's above it.

But guarding them didn't kill the mutation you found. Dropping the `have` half of the unparseable-input guard *still* left the suite green afterwards, and for a second reason: all four inputs that test used — `''`, `v22.12.0`, `nightly`, and the unparseable `want` — evaluate to 0 in a bare `(( ))`. `v22` and `nightly` are read as variable names, unset, so 0; the empty string is 0; each then loses the major comparison and is rejected with or without the regex. No input in the test distinguished the guard from its absence, so there was nothing for the `|| return 1` to make fail.

Closed with a test that uses inputs which *clear* the floor if they ever reach the arithmetic:

```bash
! node_version_at_least '0x18' "$NODE_MIN_VERSION" || return 1   # 24 in base 16
! node_version_at_least '22+1' "$NODE_MIN_VERSION" || return 1   # an expression
! node_version_at_least '22.12.0' '0x18' || return 1
```

That mutation now goes red. Worth flagging that your repro also showed the band test was never actually at risk — its final line fails under the major-only mutation on its own — so `unparseable input` was the only one where the stacking mattered, and it needed both fixes.

**The wording is corrected.** You're right that 4.3.0 under Node 20.19.2 exits 1 with an empty stdout and the message on stderr. The check is unchanged, but the code comment, the test comments and the changelog now describe the stdout-shape arm as a guard against a future release that answers through `--version` and exits zero, rather than as observed behaviour. `setup_asar: an asar that exits non-zero fails the build` now stubs that real shape instead of a generic failure, and the exits-zero test is renamed to the conditional (`a refusal that exits zero would still fail the build`) so the name doesn't assert something asar doesn't do.

**Rebased.** The `CHANGELOG.md` conflict was #825 landing under `[Unreleased]`. Kept every bullet from `main` and placed this one next to the `tests/test-patch-stage.sh` entry, since that's the other half of the same #839 trap.

Re-verified: 17/17 on `tests/dependencies.bats`, 512/0 on the full suite, `shellcheck -x` clean over the workflow's own file set. Mutation pass re-run at ten mutations (the two halves of the unparseable guard are now separable, and I added an absent-minor-reads-as-99 case) — each reddens only what it should, table in the description.

---

yannicklescure:
@sabiut ready for another look when you have a moment.

Everything from your review is in `9920bda`, plus the rebase — the rundown is in the [comment above](https://github.com/aaddrick/claude-desktop-debian/pull/841#issuecomment-5641200719). Short version:

- `|| return 1` on every negated assertion, finals included.
- Your surviving mutation needed a second fix: guarding the `!` lines didn't kill it, because all four inputs in that test read as 0 under a bare `(( ))` and were rejected either way. Added `0x18` and `22+1`, which clear the floor if they ever reach the arithmetic. Red now.
- The "exits zero" claim is corrected to what 4.3.0 actually does, in the comment, the tests and the changelog.
- Rebased; the `CHANGELOG.md` conflict was #825.

CI is green — 14 checks, `mergeable`/`CLEAN`. No rush.

---

sabiut (approved):
Approving. All three points from the first review are addressed, and the verification passes cleanly.

Verified on 9920bda after the rebase: tests/dependencies.bats passes 17/17, the full suite passes 512/512, and shellcheck -s bash is clean on both modified scripts.

Every negated assertion now includes `|| return 1`—12 in total. I also verified that the guard is doing real work: removing the `have` side of the parse guard causes the new arithmetic-input test to fail on its first non-final line, which is exactly the case the previous unguarded form could miss.

I re-ran all ten mutations from the table. Each fails only the test(s) listed for it, including the stdout-arm removal failing both refusal tests and the merged-stream mutation failing only the stderr-noise test.

The `0x18` and `22+1` inputs were the correct fix for the previously surviving mutation. The earlier inputs all evaluate to 0 inside `(( ))`, so they could not distinguish the guarded implementation from the unguarded one.

The wording around the non-zero case is also corrected, and the test now accurately models exit status 1 with empty stdout.

Issue #617, opened 2026-05-16
test(rpm): add --noscripts install pass to verify chrome-sandbox suid is scriptlet-independent


sabiut:
Follow-up from @aaddrick's pre-merge note on #595:

> `tests/test-artifact-rpm.sh:27` uses `rpm -ivh --nodeps` without
> `--noscripts`, so the test confirms `%attr` produces the right
> mode but doesn't exercise the scriptlet-skipped path that
> motivated the PR. A parallel `--noscripts` install (or a second
> matrix entry) would close that loop.

#595 moved the chrome-sandbox suid bit from a `%post chmod` to
`%attr(4755, root, root)` in `%files` so the bit survives paths
that skip scriptlets (`--noscripts`, rpm-ostree/Silverblue layered
images, etc.). The new regression guard at
[`tests/test-artifact-rpm.sh:44-48`](https://github.com/aaddrick/claude-desktop-debian/blob/main/tests/test-artifact-rpm.sh#L44-L48)
correctly asserts the installed mode, but installs with scriptlets
enabled — so if someone reverted to `%post chmod`, **the current
test would still pass**. The motivating regression mode isn't
actually exercised.

## Proposed coverage

Add a second install pass (or matrix entry) that runs with
`--noscripts` and re-asserts `assert_setuid "$chrome_sandbox"`.
The mode must come from `%attr` for that path to pass; a regression
to `%post chmod` would fail it.

Rough shape:

```bash
# Remove the prior install and re-install without scriptlets.
rpm -e claude-desktop
if rpm -ivh --nodeps --noscripts "$rpm_file"; then
    pass "rpm -ivh --noscripts succeeded"
    assert_setuid "$chrome_sandbox"
else
    fail "rpm -ivh --noscripts failed"
fi
```

## Design questions to resolve

1. **Shape: sequential vs matrix.** Two install/uninstall passes
   in one script vs a second job in the test-artifacts workflow
   parameterised on `--noscripts`. Matrix is cleaner but adds CI
   surface; sequential is the cheaper first cut.
2. **Order.** Run `--noscripts` first (so the suid check there
   doesn't piggyback on state from the scriptlet-run install) or
   second. Probably first — keeps the regression case isolated.
3. **Other scriptlet-skipped paths.** `--noscripts` covers the
   flag explicitly, but layered images (rpm-ostree/Silverblue)
   were also called out in the PR comment. CI can't easily
   simulate those, so flag explicitly in the test comment that
[...]

PR #842, opened 2026-09-10
fix(doctor): require the SingletonLock holder to be a Claude process

### LINKED ISSUE #784: doctor: SingletonLock live-PID check false-PASSes when a stale lock's PID is reused
Found during review of #740. Scoped out of that PR to keep it a pure rebase+merge, filing here as the follow-up.

The `--doctor` SingletonLock check's live-PID branch trusts `kill -0 $pid`. It pulls the PID from the `hostname-PID` symlink, signals it, and if that succeeds reports PASS: "held by running process (PID N)". The problem is that a stale lock whose PID has since been reused by any live same-user process false-PASSes. It's the mirror image of the non-symlink false PASS #740 fixes.

Chromium guards this exact case. In `process_singleton_posix.cc` it does an exe-basename check (`IsChromeProcess`) before it will treat the lock holder as live. Signalable isn't enough.

Suggested fix: after `kill -0` succeeds, verify the process is actually ours before calling it live. Read `readlink /proc/$pid/exe` and compare the basename against the claude-desktop/electron binary. When it doesn't match, downgrade to the existing stale-lock WARN.

The `kill -0` EPERM nuance is already handled right. EPERM from a same-user lock implies the PID got reused by another UID, which is stale. The gap is only same-user reuse.

Low probability in practice. It needs PID-space wraparound to land on a live process. But it's the same false-green class the doctor campaign has been stamping out (#740, and the empty-password-store and octal-disk-read fixes before it), so worth closing off.

---

CochainComplex:
Fixes #784

## What

`--doctor`'s SingletonLock check and the launcher's `cleanup_stale_lock` both decided "held by a live process" from `kill -0 $pid` alone, `$pid` being the number parsed from the `hostname-PID` symlink target. PIDs are recycled, so once a dead instance's number lands on any other same-user process the doctor prints `[PASS] SingletonLock: held by running process` on a provably stale lock, and the launcher leaves it in place. This is a false green rather than a launch blocker: Electron's own single-instance code connects to the `SingletonSocket` first and unlinks a lock whose holder does not answer (its `IsChromeProcess` compares the basename against `chrome`, which no Electron binary satisfies). So the doctor half is the fix #784 asks for; the launcher half is hygiene that keeps the log honest.

## How

A `kill -0` success is now additionally qualified by `_pid_is_claude_desktop`, a small predicate in `launcher-common.sh` next to the other `/proc/PID/exe` readers: the executable's path must carry `claude-desktop`.

- doctor: the reused-PID case downgrades to the existing stale-lock `[WARN]` with an accurate parenthetical (`PID N is not a Claude Desktop process`) and keeps the `Fix: rm ...` line.
- launcher: removes the lock and logs `Removed stale SingletonLock (PID N was reused by another process)`.
- The cross-UID EPERM path is untouched.

### Why the executable *path*, and not the basename the issue suggested or the launcher's `--class` fingerprint

The lock is shared by every Claude Desktop build on the machine, and each of them is a legitimate holder:

| holder | `/proc/PID/exe` | `--class` |
|---|---|---|
| official Anthropic `.deb` | `/usr/lib/claude-desktop/claude-desktop` | none |
| this project, deb/rpm | `/usr/lib/claude-desktop-unofficial/claude-desktop` | `com.anthropic.Claude` |
| AppImage / Nix | `…/usr/lib/claude-desktop/claude-desktop` | `com.anthropic.Claude` |
| pre-3.0 tree, still running across an upgrade | `/usr/lib/claude-desktop/node_modules/electron/dist/electron` | `Claude` |

The `--class` fingerprint exists to find *our* instances for cleanup and deliberately rejects the rest; an exe-basename match on `claude-desktop` rejects the legacy tree; the basename set the issue sketched (`claude-desktop`, `electron`) covers all four but also counts every unrelated `node_modules/.../electron` on a developer machine as live. Any of the narrower forms would have the doctor tell the user to `rm` a live lock. The last row is not hypothetical: it is the state of my own machine while writing this (legacy `claude-desktop 1.18286.0` running, `claude-desktop-unofficial 1.49585.0` installed, lock pointing at the `electron/dist` binary). The one thing all four share is `claude-desktop` in the path, so that is the test. An unreadable `/proc/PID/exe` (zombie, or a process that exec'd a setuid binary) counts as live, the harmless direction, and a replaced binary's ` (deleted)` marker needs no special case (the path still carries the name; `cleanup_replaced_desktop_ui` owns that process).

`doctor.sh` guards the call with `declare -F` like `load_launcher_config`, keeping the old verdict when sourced standalone; the SingletonLock live-PID tests in `doctor.bats` source the real `launcher-common.sh`, so the guard never hides the branch from them.

**Adjacent, deliberately not touched:** the `hostname` half of the lock target is discarded by both sites (pre-existing), so on a networked `$HOME` a lock written by another host is judged against local PIDs. Chromium treats a foreign hostname as `PROFILE_IN_USE`. Worth its own issue; folding it in here would rewrite eight existing tests.

## Tests

The stand-in spawners (a copy of `bash` at a `claude-desktop` path; a plain `sleep` for the recycled PID; an exec-aware `/proc/PID/exe` poll) and the teardown reaper live in a new `tests/test_helper.bash` loaded by both suites. The two "live PID" cases that used `$$` (bats' own bash, not a Claude executable) now point at the stand-in; each site gained a case pointing its lock at a real, signalable, non-Claude `sleep`; `cleanup_stale_lock` gained two more that pin the shape of the test: a pre-3.0 `electron/dist` path must be kept, and a deleted-exe holder must be kept.

Mutation-checked in `ubuntu:24.04` (bats 1.10.0 / shellcheck 0.9.0, the CI toolchain):

| mutation | red |
|---|---|
| predicate forced `true` (kill -0 only, today's code) | doctor 59, launcher 192 — the two reused-PID cases |
| predicate forced `false` | doctor 58, launcher 189/190/191 — the live-holder cases |
| basename-only match, the issue's literal suggestion | launcher 190 only — the pre-3.0 layout case |
| restored | 258/258, `shellcheck -x` clean |

A forced late `exec` on the stand-in (the fork still shows bash's exe on the first poll) stays green, so the poll is load-bearing. Same suites green on Ubuntu 24.04 / 26.04, Debian 12 / 13, Fedora 42, openSUSE Leap 15.6 (bash 4.4) and Arch (bash 5.3).

## Verification on an installed package

`debian:13`, `apt-get install ./claude-desktop-unofficial_1.49585.0_amd64.deb` (built from current `main`), the two scripts overlaid with `@@WM_CLASS@@` substituted the way `deb.sh` does, run as a non-root user with `DISPLAY` unset:

```
(a) lock -> plain sleep (/usr/bin/sleep)
[WARN] SingletonLock: stale lock found (PID 10901 is not a Claude Desktop process)
       Fix: rm '/home/tester/cfg/Claude/SingletonLock'
(b) lock -> /home/tester/claude-desktop (copy of sleep, 3.x layout)
[PASS] SingletonLock: held by running process (PID 10946)
(c) lock -> /home/tester/lib/claude-desktop/node_modules/electron/dist/electron (pre-3.0 layout)
[PASS] SingletonLock: held by running process (PID 10992)
(d) lock -> /home/tester/claude-desktop2 (deleted)
[PASS] SingletonLock: held by running process (PID 11037)
(e) launcher with the (a) lock: lock gone afterwards, launcher.log:
Removed stale SingletonLock (PID 11083 was reused by another process)
(f) launcher with a (b) lock: lock still present, no new "Removed" line
```

Local `./build.sh --build appimage --clean no` on this branch: exit 0. Includes the Keep-a-Changelog `[Unreleased]` entry. Merges cleanly together with the #617/#785 PRs in any order. Reviewed adversarially before opening (Electron `process_singleton_posix.cc` semantics, Nix `makeWrapper` target path, `buildFHSEnv` PID namespace, Yama scope 1); findings folded in above.

---
Co-Authored-By: Claude Fable 5.1 <noreply@anthropic.com>
90% AI / 10% Human
Claude: the `_pid_is_claude_desktop` predicate and both call-site changes, the layout analysis that replaced an earlier `--class`-fingerprint draft (Claude Opus 5), the adversarial review pass (Claude Opus 5), the `tests/test_helper.bash` extraction and the six bats cases with three-way mutation checks, the Debian 13 installed-package verification, the seven-distro matrix, CHANGELOG entry, this description.
Human: selected the issue, set the reuse-existing-code / no-`declare -F`-fallback / test-in-containers constraints, reviewed and directed.


sabiut (approved):
Reproduced end to end here. The lock on this box happens to be held by an official Anthropic Claude Desktop instance, which shares `~/.config/Claude` with our builds and so is a legitimate holder our tooling must recognise; its binary was replaced under it, so the exe reads `(deleted)`. Both main and this branch report it held. A temp lock at a plain `sleep`: main prints the #784 false PASS, this branch downgrades to the stale WARN and the launcher removes it with the new log line. A temp lock at a bash copy on a `claude-desktop` path: held, and the launcher keeps it. 260/260 across both suites, full suite green, shellcheck clean, and all three mutations redden exactly the cases your table lists; the basename-only one failing the pre-3.0 and deleted-exe cases is the argument for the path test in one line.

One non-blocking note: on a dev machine the checkout itself is named `claude-desktop-debian`, so a recycled PID landing on something running from inside it (the build's fallback `build/node`, say) reads as a holder. Harmless direction, Electron unlinks anyway; a comment in the predicate would be enough.

Approving and merging.

PR #843, opened 2026-09-10
test(rpm): install once with --noscripts to pin the chrome-sandbox suid bit

### LINKED ISSUE #617: test(rpm): add --noscripts install pass to verify chrome-sandbox suid is scriptlet-independent

CochainComplex:
Closes #617

## What

`tests/test-artifact-rpm.sh` installed with `rpm -ivh --nodeps` only, i.e. scriptlets enabled. #595 moved the chrome-sandbox setuid bit out of a `%post chmod`: the spec sets 4755 on the buildroot file in `%install`, so the payload itself carries the mode (recorded by the `%files` directory walk) and it survives install paths that skip scriptlets. `--noscripts` is the CI proxy for rpm-ostree/Silverblue layered images, which cannot be simulated. With scriptlets always running, the existing assertion could not tell that fix apart from a regression back to a `%post chmod`.

## How

A `--noscripts` install → `assert_setuid` → `rpm -e --noscripts` pass now runs *before* the existing scripted install, so the mode asserted there can only have come from the payload. The scripted install and every later check (including the CW-1 `%post` assertion, which needs scriptlets) run unchanged after it. `assert_setuid` in `test-artifact-common.sh` gained an optional label argument (same shape as `assert_contains`) so the two assertions read as distinct lines; the default message is unchanged.

## Mutation check

Spec reverted to set the bit via a `%post chmod` instead of the `%install` chmod (the pre-#595 shape), rpm rebuilt, test rerun in a fresh Fedora 42 container with the packages CI's rpm leg installs:

```
[PASS] rpm -ivh --noscripts succeeded
[FAIL] Setuid bit set (noscripts install, #617): /usr/lib/claude-desktop-unofficial/chrome-sandbox
[PASS] rpm -e --noscripts succeeded
[PASS] rpm -ivh succeeded
...
[PASS] Setuid bit set (scripted install): /usr/lib/claude-desktop-unofficial/chrome-sandbox
Results: 43 passed, 1 failed
```

Only the new assertion goes red. Spec restored, rebuilt, rerun: `44 passed, 0 failed`, identical to the baseline run before the mutation. `shellcheck -x` (0.9.0) clean on both files. The rpm was built with `./build.sh --build rpm --clean no` in an `ubuntu:24.04` + Node 22 container, so the patch stage ran on this branch.

Includes the Keep-a-Changelog `[Unreleased]` entry. Merges cleanly together with the #784/#785 PRs in any order.

---
90% AI / 10% Human
Claude: the `--noscripts` pass, the `assert_setuid` label argument, the spec-mutation rebuild and Fedora 42 verification, CHANGELOG entry, this description. Reviewed by Claude Fable 5.1.
Human: selected the issue, set the sequential-not-matrix and noscripts-first constraints, reviewed and directed.


sabiut (approved):
Verified the premise against the spec on main: the bit comes from the `%install` chmod on the buildroot file, recorded by the `%files` walk, and `%post` never touches chrome-sandbox — so a scripted-only install genuinely could not tell #595 apart from a `%post chmod` regression. The `--noscripts` pass running first, isolated from any scriptlet state, is the right shape. CI's rpm leg shows all four new lines passing on the built rpm (44/0), shellcheck clean, and the `assert_setuid` label mirrors `assert_contains`. Closes the loop on #617 exactly as asked. Approving and merging.

PR #844, opened 2026-09-10
fix(doctor): stop judging chrome-sandbox on AppImage runs

### LINKED ISSUE #785: doctor: chrome-sandbox check misreports on AppImage — the "expected for AppImage" WARN is unreachable
Found this during review of #745 (the doctor Electron/chrome-sandbox helper extraction). Scoped it out of that PR to keep the change behavior-preserving, so filing it here.

## What's wrong

The doctor's chrome-sandbox check has a WARN branch that prints "Chrome sandbox not found (expected for AppImage)". On AppImage that branch can never fire.

AppRun stages the full official tree into the mount, including chrome-sandbox right beside the ELF (`scripts/packaging/appimage.sh:30-32`). The check probes for chrome-sandbox adjacent to electron, so it always finds the file inside the mounted squashfs. The "not found" path is dead code on AppImage.

## What happens instead

Static analysis says the check then stats that file and most plausibly FAILs. SUID root generally isn't preserved through unprivileged tar extraction at build/staging time, so the staged chrome-sandbox won't carry the setuid bit. That produces a `Fix: sudo chown root:root /tmp/.mount_claudeXXX/.../chrome-sandbox` hint.

That hint is unactionable. The mount is transient and read-only, so the chown can't stick. And it's moot regardless: the AppImage launcher runs `--no-sandbox`, so the sandbox binary isn't in play.

I read this out of the code rather than running a live AppImage, so treat the FAIL verdict as the likely outcome, not confirmed.

## Not a #745 regression

The inline original has the identical blind spot. #745's extraction is behavior-preserving right down to it, so it didn't introduce anything. But the new helper docstring now repeats the wrong "expected for AppImage" claim in a second place, so the bad wording is easier to trust now.

## Suggested fix

Detect the AppImage case (mount path or the `$APPIMAGE` env var) and either skip the sandbox check or emit an INFO explaining that sandboxing is disabled by design there. Fix the docstring wording at the same time.

This should be verified against a live AppImage run before shipping, since the FAIL verdict above is static-analysis-only.

---

CochainComplex:
Fixes #785

## Observed (not inferred)

Locally built `1.49585.0` AppImage from current `main`, clean `debian:13` container, non-root user, `DISPLAY` unset. Both `./claude-desktop-unofficial-*.AppImage --appimage-extract-and-run --doctor` and `--appimage-extract` + `squashfs-root/AppRun --doctor`:

```
[FAIL] Chrome sandbox: perms=755, owner=tester
       Fix: sudo chown root:root /home/tester/squashfs-root/usr/lib/claude-desktop/chrome-sandbox
            sudo chmod 4755 /home/tester/squashfs-root/usr/lib/claude-desktop/chrome-sandbox
2 check(s) failed.
```

(The other failure is the expected no-display-server one.) Inside the image the file is `-rwxr-xr-x`, owned by the invoking user: `appimage.sh` stages the whole official tree, so `chrome-sandbox` is present — the "not found (expected for AppImage)" `[WARN]` is unreachable there — but the AppDir permission normalization drops the setuid bit, so the `4755`/root test fails by construction. The hint is unactionable (transient, read-only path) and moot: `build_electron_args` passes `--no-sandbox` unconditionally for `appimage`, so the helper is never invoked. That is why `_doctor_check_effective_sandbox` already short-circuits on this package type; the file-permission check simply had not caught up after #816 narrowed it to the running binary's own sandbox.

## Fix

`run_doctor` forwards its package type to `_doctor_check_chrome_sandbox`, which on `appimage` prints one informational line and returns before any `stat`:

```
       Chrome sandbox: not used (AppImage runs with --no-sandbox)
1 check(s) failed.
```

deb/rpm/nix are untouched. The not-found `[WARN]` stays (it is meaningful where the helper is supposed to be installed) but no longer claims "expected for AppImage"; it names the path it probed.

## Tests

Two `doctor.bats` cases (134 total). The appimage case uses a `stat` tripwire, so the "no stat" half is enforced rather than decoration; the second is a blast-radius guard that the same 0644 sandbox still FAILs for `deb`. Mutation check: dropping the gate turns the appimage case red on the tripwire assertion. `ubuntu:24.04` (bats 1.10.0 / shellcheck 0.9.0): `shellcheck -x` clean, 134/134 and `launcher-common.bats` 122/122 green. After overlaying the fixed `doctor.sh` into the extracted image, the same container run shows the info line and no FAIL/hint. Local `./build.sh --build appimage --clean no` on this branch: exit 0.

Includes the Keep-a-Changelog `[Unreleased]` entry. Merges cleanly together with the #784/#617 PRs in any order.

---
90% AI / 10% Human
Claude: the live-AppImage reproduction, the package-type gate and wording fix, the tripwire and blast-radius tests with mutation check, CHANGELOG entry, this description. Reviewed by Claude Fable 5.1 (style-guide pass, CI-toolchain re-run).
Human: selected the issue, required observation before any code change, reviewed and directed.


sabiut (approved):
Reproduced the fix end to end here: with a user-owned 755 chrome-sandbox beside the binary, main FAILs with the `sudo chown` hint for both types, this branch prints the single info line for `appimage` and keeps the identical FAIL for `deb`. 134/134, full suite green, shellcheck clean, and dropping the gate reddens exactly the tripwire test. The WARN rewording is right as well; the old text was unreachable on AppImage and misleading everywhere else. Approving and merging.

PR #845, opened 2026-09-14
fix(ci): raise the release job's Node to the asar engine floor

yannicklescure:
Stacked on #814 — this PR targets `fix/813-decouple-reference-source-tooling`, not `main`, so the diff shows only the Node bump. GitHub will retarget it to `main` automatically when #814 merges. Raised from the review note in https://github.com/aaddrick/claude-desktop-debian/pull/814#issuecomment-5657245963.

## The problem

#814 correctly hoists the `asar` / `prettier` install out from behind the `claude-desktop-versions` gate, but the `Set up Node.js 20` step feeding it keeps `node-version: "20"`.

`@electron/asar` has declared `engines.node >=22.12.0` since 4.0.0, and the claude-code CLI installed further down declares `>=22.0.0`. Neither mismatch fails the install — what it does instead depends on the runner's npm.

A bare `npm install -g <name>` is a **range** spec (`*`), not a tag, and `npm-pick-manifest` >=9 filters ranges by `engines`. npm 10.x bundles pick-manifest 9, and `setup-node` with `node-version: "20"` ships npm 10.8.2. So on the runner, Node 20 does not blow up — it quietly installs older packages:

| Package | Node 20 (npm 10) | Node 22 |
|---|---|---|
| `@electron/asar` | 3.4.1 | 4.3.0 |
| `@anthropic-ai/claude-code` | 2.1.197 (`>=18.0.0`) | 2.1.270 |
| `prettier` | 3.9.6 | 3.9.6 |

On npm <=9, which does not filter, the same command takes 4.3.0 and the `EBADENGINE` mismatch is only a **warning** — so the wrapper installs, lands on disk, and dies on every invocation. That is the shape #839 was reported as, from a Debian 13 host whose npm is packaged separately from its Node. Verified against the live registry across pick-manifest 6/7/8/9/10; the filtering appears at 9.

So there are two silent failure modes, not one, and which one a release gets is a property of the runner rather than of anything in this repo. Pinning the runtime is what makes the resolved versions ours. The claude-code CLI case bites on today's runners: it is 73 releases behind what Node 22 resolves.

## Why #814's guard doesn't already cover it

On the npm <=9 path, the `command -v asar` guard #814 adds checks for the wrong thing: the binary *is* on `PATH` — it just refuses to run — so the guard passes, `asar extract` fails, and since that step carries no `continue-on-error`, the release job fails and the four publish jobs on `needs: [release]` skip. That is the outage shape #814 exists to prevent, reached through a different door.

That path is not what GitHub's runners take today, so this is a latent hazard on that leg rather than a live break — but it costs one line to close, and the claude-code staleness above is live regardless.

## What changed

- `.github/workflows/ci.yml` — `node-version: "20"` → `"22"`, step renamed to match, with a comment recording the floor.
- `tests/ci-release-job.bats` — a new case asserting the setup step's major is ≥ 22. #814's suite has no assertion on the Node version, so before this a revert of the bump reds nothing.
- `docs/learnings/test-methodology-and-coverage.md` — one bullet on a trap I hit writing the workflow comment (below).
- `CHANGELOG.md` — bullet under `[Unreleased]` → Fixed.

## Verification

`tests/ci-release-job.bats` 10/10 green. Mutation-checked three ways, each reds the new test: revert the bump to `"20"`; delete the `node-version` key; comment the key out.

`actionlint` clean on the change — the one finding it reports (`constant expression "false" in condition` at the manually-managed release-notes step) is present on the base branch and untouched here.

## One trap worth flagging for the suite

The first version of my workflow comment spelled the packages as `@electron/asar` and `@anthropic-ai/claude-code`, and that **reds three of #814's tests**. `uncommented` is applied per step block, but the region-level greps in "node is set up before asar is installed" and "asar is installed by exactly one step" read the raw region — so a package name in a YAML comment counts as a second step installing asar, and as an asar reference preceding the Node setup.

That's the invariant working, not a parser bug, so I fixed it in the comment rather than the test: shared tooling is spelled bare (`asar`, `the claude-code CLI`) in that region, with the npm specifiers left to the `run:` lines that actually install them. The `ci.yml` comment says so inline and the learnings doc records it, since it's a live tripwire for whoever next annotates that block. Happy to loosen the greps instead if you'd rather the tests absorbed it.

---
85% AI / 15% Human
Claude: verified #814's state and re-read its rewritten block; worked out the npm resolution behaviour against the live registry with `npm-pick-manifest` — including getting it wrong first (asserting a bare spec resolves through the unfiltered `latest` tag) and correcting it to the range-spec/pick-manifest->=9 mechanism documented above; identified that the `command -v asar` guard misses the npm <=9 failure mode and that the claude-code CLI shares the floor; wrote the workflow change, the bats case, the mutation checks, the learnings bullet and the changelog entry.
Human: caught that the CI-side floor was left unfixed by #841, directed the stacked-PR approach over a conflicting edit to `main`, and reviewed.


yannicklescure:
Note on the absent check run: `tests.yml` gates its `pull_request` trigger on `branches: [main]`, so a PR stacked on #814's branch runs no BATS. The blank check list here is structural, not a skip — it will run once GitHub retargets this to `main` on #814's merge.

In the meantime the suite was run locally against this branch (bats 1.13.0), 10/10 green, and each of the three mutations was applied and reverted independently rather than in sequence:

| Mutation to `ci.yml` | Result |
|---|---|
| `node-version: "20"` | `not ok 7` — fails the `-ge 22` comparison |
| `node-version` key deleted | `not ok 7` — fails the regex match |
| key commented out | `not ok 7` — fails the regex match |

The third confirms `uncommented` is load-bearing in the new test rather than decoration.

---

yannicklescure:
Correcting the reasoning in the description above — I had the npm mechanism wrong, and the correction changes how live this is. Pushed as 7cf9f72; no change to the workflow or to any test.

**What I claimed:** a bare `npm install -g @electron/asar` resolves through the `latest` dist-tag, which is not engine-filtered, so a Node 20 runner installs 4.3.0 and gets a binary that cannot start.

**What actually happens:** a bare name is a **range** spec (`*`), not a tag, and `npm-pick-manifest` >=9 filters ranges by `engines`. npm 10.x bundles pick-manifest 9, and `setup-node` with `node-version: "20"` ships npm 10.8.2. My original check passed the literal string `latest` as the spec — a spec npm never uses here — which is why it looked unfiltered.

Verified against the live registry across pick-manifest majors; the filtering behaviour appears at 9:

```
pick-manifest 6.1.1:  node 20.19.2 -> 4.3.0   |   node 22.12.0 -> 4.3.0
pick-manifest 7.0.2:  node 20.19.2 -> 4.3.0   |   node 22.12.0 -> 4.3.0
pick-manifest 8.0.2:  node 20.19.2 -> 4.3.0   |   node 22.12.0 -> 4.3.0
pick-manifest 9.1.0:  node 20.19.2 -> 3.4.1   |   node 22.12.0 -> 4.3.0
pick-manifest 10.0.0: node 20.19.2 -> 3.4.1   |   node 22.12.0 -> 4.3.0
```

**So the release job is not currently broken**, and I should not have implied it was. On today's runners Node 20 silently installs asar 3.4.1 and `@anthropic-ai/claude-code` 2.1.197 rather than failing. The install-and-die shape belongs to npm <=9, which is how #839 was reported — Debian 13 packages npm separately from its Node.

What survives:

- The claude-code CLI resolves 73 releases behind what Node 22 gets, silently, on today's runners. That one is live.
- The npm <=9 path is a real hazard on this leg, and it is the path where #814's `command -v asar` guard does not help — but it is latent, not firing.
- Either way the resolved versions were a property of the runner's npm rather than of this repo, which the pin fixes.

Still worth the one line in my view, but on "make the toolchain deterministic" grounds rather than "CI is broken" — your call, and closing this is a reasonable read if the deterministic-versions argument doesn't carry for you.

---

yannicklescure:
Withdrawing the last line of my previous comment — I shouldn't have offered closing this. I found my stated mechanism wrong and let that slide into doubting the change, which are two different questions. The mechanism needed correcting; the change didn't.

Restating the case on the corrected facts:

- **Live on today's runners.** `Install claude-code CLI` runs under the same setup-node as the asar install. On Node 20, npm 10 resolves `@anthropic-ai/claude-code` 2.1.197 rather than 2.1.270 — 73 releases back — and that is the CLI generating the release notes. This one does not depend on the runner's npm at all; npm 10 picks the old version *because* its engine filtering works.
- **Latent but real.** On npm <=9 the same step takes asar 4.3.0, and that is the path where the `command -v asar` guard genuinely does not help: binary on `PATH`, fails only when invoked, no `continue-on-error` on the step, release job fails, four publish jobs on `needs: [release]` skip.
- **Structural.** Which asar major extracts the published reference-source tarball is currently a property of the runner's npm rather than of this repo.

One line, and it closes a live staleness bug plus a latent chokepoint failure. I'd still merge it.

---

sabiut (approved):
Approving. Verified on 7cf9f72, both stacked on the old #814 base and test-merged onto current main after #814 landed.

tests/ci-release-job.bats 10/10; the main merge runs the full suite at 522/0 with the release job on `node-version: "22"`. Each of your three mutations reds only test 7. I also tried `node-version: "22.12"` and an unquoted `22`; both pass, so the regex isn't over-tight.

The corrected mechanism in 7cf9f72 matches what I see here. Node 20.19.2 with Debian 13's npm 9.2.0 resolves asar 4.3.0 and claude-code 2.1.270, both under EBADENGINE warnings, and asar then dies with `CANNOT RUN WITH NODE 20.19.2`. The same Node with npm 10.9.9 resolves asar 3.4.1 and claude-code 2.1.197, and that asar runs. Under Node 22 the bare install lands 4.3.0 and runs. So the runner's npm is what decides which silent outcome you get, and pinning the runtime is the right fix.

Two docs nits, neither blocking:

1. The workflow comment in ci.yml still describes the npm <=9 shape as the only one ("an unpinned install on Node 20 succeeds and leaves a binary that exists and never runs"). 7cf9f72 corrected the test comment and the changelog but not this one, and it's the comment the next person annotating that block will read.

2. The learnings bullet says spelling `@electron/asar` in the comment reds three of #814's tests; here it reds two ("asar is installed by exactly one step" and "node is set up before asar is installed").

Keeping the greps as they are is the right call. I'll merge once the last build leg finishes.

yannicklescure:
Thanks for the independent measurement — your npm 9.2.0 vs 10.9.9 numbers on the same Node 20.19.2 are the cleanest statement of the mechanism yet, and `CANNOT RUN WITH NODE 20.19.2` is the exact failure string I never captured because I never had a npm <=9 host to hand.

Both nits are valid and are fixed in #848.

**Nit 1 was a real regression, not an oversight in wording.** The `ci.yml` comment edit *was* written in `7cf9f72` — it just never got committed. The mutation re-check in that same turn restores the workflow with `git checkout -- .github/workflows/ci.yml` between mutations, which reverted the still-uncommitted comment, and `git add -A` then committed only the test-comment and changelog halves. Commit before mutating, or the restore is indistinguishable from a revert. Good catch; it is exactly the comment the next person annotating that block would have read.

**Nit 2: you are right as the bullet was worded.** Verified both spellings:

```
--- comment names @electron/asar ONLY
    not ok 2 asar is installed by exactly one step
    not ok 6 node is set up before asar is installed, and is ungated

--- comment names BOTH packages
    not ok 2 asar is installed by exactly one step
    not ok 5 asar is not co-installed with release-notes-only tooling
    not ok 6 node is set up before asar is installed, and is ungated
```

My "three" came from the original annotation, which named `@anthropic-ai/claude-code` alongside asar and so tripped test 5 too. The bullet now gives both counts and which spelling produces which.

Also noted that you tried `node-version: "22.12"` and an unquoted `22` against the regex — that is the check I should have written myself rather than only testing the mutations that red it.

---

Issue #820, opened 2026-08-09
tray-icon-selection patch: harden the chain-callee capture (latent on current upstream)


aaddrick:
All six build legs fail on upstream 1.26832.0. That's amd64 and arm64 across deb, rpm, and AppImage, on both the `main` push run ([31291569939](https://github.com/aaddrick/claude-desktop-debian/actions/runs/31291569939)) and the auto-cut tag run ([31291571687](https://github.com/aaddrick/claude-desktop-debian/actions/runs/31291571687)), 2026-08-09.

Every leg stops at the same place, the "Patch app.asar" stage:

```
index.js requires 104 main chunks (...) — upstream split the main bundle
across files; patches need per-anchor resolution. Re-point
scripts/patches/*.sh before shipping.
```

That's our own guard in `_resolve_main_js` (`scripts/patches/app-asar.sh`, around line 180). It hard-fails on a multi-chunk split instead of picking a chunk and mispatching quietly. The guard is doing exactly what it was built to do.

## What changed upstream

Since 1.19367.0 the stub `index.js` required exactly one main chunk. `_resolve_main_js` returns that chunk and every patch operates on it.

1.26832.0 emits 104 `require()` calls across 83 distinct chunks.

The guard counts raw requires rather than unique ones, so its message overstates the split by 21. Worth fixing while we're in there.

## What's affected

All five active patches rest on the single-chunk assumption:

- `quick-window.sh`
- `org-plugins.sh`
- `virtiofsd-probe.sh`
- `cowork-bwrap.sh`
- `tray-icon-selection.sh` (`patch_tray_icon_env_override`)

Everything ahead of the patch stage passed. Download, SHA-256 verify, extraction, the productName guard, and both the AU-1 and MB-1 upstream tripwires are clear. Shellcheck and BATS are green. The failure is anchor resolution and nothing else.

## Blast radius

The auto-bump landed on main (13a7e14) and cut tag `v3.2.1+claude1.26832.0`, but no release came out of it. Latest release is still `v3.2.1+claude1.24012.11` from 2026-08-04, so APT and DNF are serving 1.24012.11. The tag is burned, same as v3.2.0 was burned by the 1.19367.0 split.

## Precedent

We've done this once already. All merged 2026-07-12:

- #793 fix(build): resolve code-split main chunk so patches land on 1.19367.0. This is the PR that added `_resolve_main_js` and the guard.
- #794 test coverage for the chunk resolution, by @sabiut
[...]

PR #846, opened 2026-09-14
fix(asar): share one pinned asar resolver across all callers (#839)

### LINKED ISSUE #779: [bug]: StartupWMClass is incorrect and produces a second generic icon

yannicklescure:
Follow-up to #841, which raised the build's Node floor to 22.12.0 and added a run-check to `setup_asar`. Two callers still reached `@electron/asar` through an unpinned `npx --yes @electron/asar` and sit outside `setup_nodejs`'s reach, so that floor does not cover them:

- `tools/patch-necessity-audit.sh`
- `tests/test-artifact-common.sh`

## The failure

Every `@electron/asar` 4.x release declares `engines.node >=22.12.0` and hard-refuses to start below it. npm reports that as an `EBADENGINE` **warning**, not an error, so the wrapper installs, lands on disk, and fails on every invocation. Because the first thing callers ask asar for is usually `package.json`, the failure surfaces as `desktopName is missing/empty — cannot derive WM_CLASS`, which reads like upstream moved a field rather than like the toolchain is broken. Debian 13 stable ships Node 20.19.2, so that is the default local experience there.

Whether an unpinned install even lands on 4.x depends on the npm version (npm 9 takes `latest`; npm 10's manifest picker skips engine-incompatible versions), so the symptom is host-dependent — part of why it is confusing.

In the audit tool the multi-word `asar_exec='npx --yes @electron/asar'` was also invoked **unquoted**, which turns `extract` into a package name.

## The change

`tests/test-patch-stage.sh` already solved this with `_resolve_asar`. Rather than let this become a third copy, that helper moves to `scripts/_common.sh` and all three callers share it: prefer a real `asar` on `PATH`, else `npm install @electron/asar@<major>` into a caller-supplied directory, else fail loud — no `npx` path survives, and `"$asar_exec"` is quoted at every call site.

**On where it lives.** `tools/patch-necessity-audit.sh` and `tests/test-patch-stage.sh` already source `scripts/_common.sh`. `tests/test-artifact-common.sh` did not, so it gains one `source` line resolved from its own `${BASH_SOURCE[0]}` rather than a caller's `$script_dir` (three entrypoints source it, each with its own). The helper takes the install directory and the pinned major as arguments, so nothing hardcodes a work directory.

The shared copy also inherits `setup_asar`'s two-arm run-check — exit code and stdout shape judged separately, merged streams reported, the match anchored at the start of the reply because the refusal text quotes the offending Node version.

**On pin policy.** Both new callers pin `@3`, not `@4`. They only ever *read* an asar, which every major does identically, and 3.4.1 declares `engines.node >=10.12.0` so it runs anywhere.

For the artifact tests that turned out to be load-bearing **in CI**, not only on a stable-distro laptop. `test-artifacts.yml` runs no `actions/setup-node` and takes whatever `apt-get install nodejs` (Ubuntu legs) or `dnf install nodejs` (Fedora rpm-leg container) hands it. The CI run on this branch confirms what that actually resolves to:

| Leg | Node from the package manager | Would `@4` (`engines.node >=22.12.0`) start? |
|---|---|---|
| Ubuntu — deb, appimage (amd64 + arm64) | **18.19.1** (`nodejs (18.19.1+dfsg-6ubuntu5`, noble) | **No** |
| Fedora container — rpm (amd64 + arm64) | 22.22.0 (`nodejs-1:22.22.0`) | Yes |

So four of the six artifact jobs run on Node 18. Pinning `@4` would have hard-failed CI outright, not just local runs on the Node 20 host from #839. All six jobs show the resolver taking its install branch (no `asar` on `PATH` in CI) and the assertions behind it executing:

```
[PASS] Using asar executable: /tmp/tmp.rX50h94Ej3/node_modules/.bin/asar (v3.4.1)
[PASS] package.json main points into .vite/build/
[PASS] package.json productName is Claude
[PASS] StartupWMClass matches asar desktopName (com.anthropic.Claude)
```

Which also puts a number on the silent-skip fix below: on the Ubuntu legs those three assertions were **never running** before this PR — `npx --yes @electron/asar` under Node 18 could not start, the extract failed, and the skip branch reported `[PASS]`.

The build keeps resolving through `setup_asar` under the enforced `NODE_MIN_VERSION` floor — that is the path that actually *writes* an asar, so that is where the version matters.

## A silent skip, fixed alongside

`validate_app_contents` reported `[PASS] Skipping asar extraction (tool not available)` when the tool was unusable, which silently dropped every assertion behind it — the `package.json` shape, `productName`, and the `StartupWMClass`/`desktopName` agreement that closes #779. An unresolvable asar or a failed extract now fails instead.

## Testing

`tests/asar-resolve.bats` (21 tests) pins the resolver and both call sites. Fourteen mutations each turn a different test red:

| Mutation | Test that goes red |
|---|---|
| Drop the `@$major` pin | pins a major instead of taking latest (+1) |
| Delete the exit-code arm | version-shaped reply on a non-zero exit |
| Delete the stdout-shape arm | refusal that exits zero (+1) |
| Unanchor the version regex | refusal that exits zero |
| Delete the `-x` check | non-executable resolved binary |
| Delete the empty-argument guard | empty install dir is refused |
| Delete the missing-directory guard | missing install dir is named |
| Ignore a `PATH` asar | 8 tests |
| Drop the Node floor from the message | the refusal names the Node floor |
| Audit tool back on `npx` | no script reaches asar through npx (+1) |
| Unquote `$asar_exec` in the audit tool | audit tool resolves before it extracts |
| Artifact tests back on `npx` + skip | 3 tests |
| Restore the skip-`[PASS]` | failed extract fails instead of passing |
| Capture the resolver with `$(...)` | a resolved asar actually reaches the extract |
| Leave a private copy in the patch-stage harness | patch-stage keeps no private copy |

That `$(...)` row is not hypothetical. The first draft of the artifact-test call site had `asar_log=$(_resolve_asar ...)`, and since `_resolve_asar` publishes `$asar_exec` as a global, the subshell discarded the assignment and the extract invoked the empty string. That still *fails*, so both negative tests stayed green on the wrong cause — it only showed up in a real end-to-end run, and needed a positive test that has to reach a successful extract before it was pinned. Same subshell-discards-the-mutation class as the `run`-wrapped assertions in `docs/learnings/test-methodology-and-coverage.md`.

Also verified:

- Full bats suite: **533/533**, including `scripts/cowork-fallback/tests/*.bats`.
- `shellcheck -x` clean over CI's file selection (24 files); no added line over 80 columns.
- Real `npm install` through `_resolve_asar` lands 3.4.1 with `engines: {"node":">=10.12.0"}`, and `--version` prints `v3.4.1` (the leading `v` is why the shape regex keeps `v?`).
- End-to-end in a clean `bash` env with no leaked `$asar_exec`: `validate_app_contents` against a real packed asar → 11 PASS / 0 FAIL.
- End-to-end `tools/patch-necessity-audit.sh --tree` against a packed stub tree → resolves, extracts, prints the matrix.

- `tests/test-patch-stage.sh` against the pinned official **1.52386.3** amd64 `.deb`, on a host with no `asar` on `PATH` — so the new `_resolve_asar "$work_dir" 3` call site ran its install branch end to end (`installing @electron/asar@3 … (v3.4.1)`). All four asserts green, 0 FAIL: patch stage exits 0, all 183 JS files parse, all five injected markers survive the repack, second pass exits 0 and the asar is byte-identical.

---
90% AI / 10% Human
Claude: read the #841 prior art, extracted `_resolve_asar` into `scripts/_common.sh`, rewired both `npx` callers and the patch-stage harness, decided and justified the per-caller pin policy, found and fixed the subshell bug via end-to-end verification, wrote `tests/asar-resolve.bats` and ran the 14-mutation check, wrote the CHANGELOG entry.
Human: identified the two uncovered callers and scoped the follow-up, set the extract-vs-copy and mutation-check requirements.


sabiut (approved):
Approving. Verified on eb693d0.

tests/asar-resolve.bats 21/21, full suite 533/0, CI-selection shellcheck clean, and it test-merges onto current main without conflict.

All fifteen mutation rows go red. Several go wider than the table says, which is fine: ignoring a PATH asar reds 8, dropping the `@major` pin reds 2, dropping the stdout arm reds 2, the audit-tool npx revert reds 2, the artifact-test npx+skip revert reds 4. Nothing reds that shouldn't.

I ran the one thing you left out. `tests/test-patch-stage.sh` against the real pinned 1.52386.3 deb passes end to end through the shared resolver: 183 JS files parse, all five markers survive, second pass byte-identical, asar 3.4.1.

End to end on the #839 host shape, Node 20.19.2 with Debian 13's npm 9.2.0 and no asar on PATH: `_resolve_asar` lands 3.4.1 and it runs; `validate_app_contents` then passes all eleven assertions including the StartupWMClass one that the old `[PASS]` skip was dropping; the audit tool resolves and extracts the same way. With a dead 4.3.0 asar first on PATH, `validate_app_contents` now FAILs naming the Node floor instead of passing.

One design note, not blocking: the resolver takes whatever asar is on PATH and hard-fails if it can't run. A host with a stale global 4.x from npm 9 gets a clear message but no fallback to the @3 install, even though that install would fix it. Falling through to the install on a failed run-check would make that host self-healing. Happy to take it as a follow-up.

Unrelated, pre-existing, noting for a separate issue: on a 1.49585.0 tree the audit tool's main-chunk resolution captures two chunk names because that index.js requires two, and aborts with "requires ... but it is missing". Same on main, so nothing to do here.

yannicklescure:
Thanks @sabiut — particularly for running `tests/test-patch-stage.sh` and the Node 20.19.2 / npm 9.2.0 host shape. That second one is the configuration the whole change is aimed at and the one I could not reproduce locally (this box is Node 24 / npm 11), so an independent pass on it is worth more than anything I could have added here.

Your wider mutation counts match mine — I reported the narrowest row per mutation rather than the full blast radius, which undersells a few of them. The PATH-asar one reding 8 is the honest number.

## On the fallback note

You're right, and I'd rather not fold it into this PR — it changes the resolver's failure semantics and needs its own test design rather than riding an already-verified diff.

Concretely, the shape I'd want:

```
PATH asar fails the run-check
  → warn, naming the refusal
  → fall through to the @3 install
  → if THAT also fails, report both: the PATH asar's refusal first
    (the actual root cause), the install failure second
```

That last line is the part that needs care. The entire point of #839 is that the operator learns it is a Node problem at the point of failure. A naive fallback replaces the accurate `asar at /usr/bin/asar will not run: CANNOT RUN WITH NODE 20.19.2` with `Failed to install @electron/asar@3` on any host that is offline or behind a proxy — strictly worse diagnosis for the same underlying cause. So the both-fail path has to carry the Node floor message, and that deserves its own mutation row.

There's also real test churn, which I checked rather than guessed. Seven of the current tests stub a broken `asar` on `PATH` and assert a refusal, with no `npm` stub:

```
resolve asar: an asar that exits non-zero is refused
resolve asar: a version-shaped reply on a non-zero exit is refused
resolve asar: a refusal that exits zero is refused
resolve asar: the same refusal on stderr at exit zero is refused
resolve asar: the refusal names the Node floor, not just the tool
artifact tests: a failed extract fails instead of passing
artifact tests: an unresolvable asar fails instead of skipping
```

Under a fallback every one of those would reach a *real* `npm install` against the registry — they'd go red, and worse, they'd go red by hitting the network from a unit test. Each needs a `_stub_npm_install` variant that fails, so the refusal path is still reachable and still asserted. Not hard, but it is the bulk of the work and it is exactly the kind of thing that should not be appended to a PR you have already verified at a specific SHA.

Happy to open it as a follow-up. `resolve asar: an asar on PATH wins over an install` stays valid unchanged either way, since it stubs a *working* PATH asar.

## On the 1.49585.0 audit-tool abort

Confirmed, and confirmed out of scope — that block is byte-identical to `main` (`git show main:tools/patch-necessity-audit.sh`, lines 100-107), untouched by this PR.

Worth flagging that the root cause is a bit deeper than the two-line `grep -oP` capture, though, and a fix should probably not just take the first match. The audit tool still assumes the 1.19367.0 model of *one* content-hashed main chunk behind a stub:

```bash
main_chunk=$(grep -oP 'require\("\./\Kindex\.chunk-[^"]+\.js(?="\))' "$index_js")
```

The build stopped believing that at #820. Per `_resolve_anchor_file` in `scripts/patches/app-asar.sh:152`, 1.26832.0 "dissolved that core entirely — `index.js` became a 190 KB file require()ing 83 chunks, across two chunk families (`index.chunk-*` and `index2.chunk-*`), with the tray anchor left behind in `index.js` itself", and there is "deliberately no 'the main JS file' any more" — each anchor resolves its own file.

So the audit tool's verdict matrix is keyed to a single resolved `$index_js` that no longer exists as a concept. On a multi-chunk tree it currently aborts; if it were patched to take the first chunk it would silently report `not-needed` for every anchor living in any other chunk, which is the worse failure since those verdicts feed the rebase decisions in `docs/learnings/official-deb-rebase-verification.md`. The real fix is probably to have it resolve per-anchor the way the patch stage does.

Want me to file that as an issue with the above, and open the fallback follow-up alongside it?

---

yannicklescure:
Both filed:

- #849 — `_resolve_asar` fallback on a stale global asar (enhancement). Carries the both-fail message design, the seven tests that would otherwise reach the live registry, and an acceptance checklist.
- #850 — `patch-necessity-audit.sh` multi-chunk abort (bug). Credited to you, noted as pre-existing on `main`.

One thing I put in #850 that wasn't in my earlier reply, since it seems like the part most likely to get lost: a `| head -1` fix would be worse than the current abort. The audit tool greps every anchor against one resolved `$index_js`, so first-match would silently report `not-needed` for anchors living in any other chunk — and a present-but-unmatched anchor is indistinguishable from one upstream already fixed. Those verdicts feed the matrix in `docs/learnings/official-deb-rebase-verification.md` that decides whether a patch gets deleted, so the failure mode is "argues for deleting a load-bearing patch". The abort is the safe direction.

#846 is unchanged and still at `eb693d0`.

---

PR #847, opened 2026-09-14
fix(ci): raise the triage workflows' Node to the tooling engine floor

yannicklescure:
Third and last site of the `EBADENGINE`-warning trap from #839. Targets `main` directly and is **independent of #814 / #845** — folding it into that stack would have blocked a triage fix behind an unrelated release-job PR.

## The problem

`issue-triage.yml` sets up `node-version: "20"` at four sites and then installs tools whose engine floors sit above it:

| Job | Installs | Declared floor |
|---|---|---|
| `gather-context` (L108) | `@anthropic-ai/claude-code` | `>=22.0.0` |
| `fetch-reference` (L219) | `@electron/asar prettier` | `>=22.12.0` (asar) |
| `investigate` (L295) | `@anthropic-ai/claude-code` | `>=22.0.0` |
| `apply-labels` (L484) | `@anthropic-ai/claude-code` | `>=22.0.0` |

`@electron/asar` has declared `engines.node >=22.12.0` since 4.0.0 and the claude-code CLI declares `>=22.0.0`. Neither mismatch fails the install — what it does depends on the runner's npm.

A bare `npm install -g <name>` is a **range** spec (`*`), not a tag, and `npm-pick-manifest` >=9 filters ranges by `engines`. npm 10.x bundles pick-manifest 9, and `setup-node` with `node-version: "20"` ships npm 10.8.2. So on the runner, Node 20 quietly installs older packages rather than failing:

| Package | Node 20 (npm 10) | Node 22 |
|---|---|---|
| `@electron/asar` | 3.4.1 | 4.3.0 |
| `@anthropic-ai/claude-code` | 2.1.197 (`>=18.0.0`) | 2.1.270 |
| `prettier` | 3.9.6 | 3.9.6 |

On npm <=9, which does not filter, the same command takes 4.3.0 and the `EBADENGINE` mismatch is only a **warning**, so the wrapper installs and dies on every invocation. That is the shape #839 was reported as, from a Debian 13 host whose npm is packaged separately from its Node. Verified against the live registry across pick-manifest 6/7/8/9/10; the filtering appears at 9.

Two silent failure modes, then, and which one a triage run got was a property of the runner rather than of anything in this repo. The live one on today's runners is the CLI: a triage job would have run `@anthropic-ai/claude-code` 2.1.197, 73 releases behind. Pinning the runtime makes the resolved versions ours.

**This was latent, not firing.** v1 has been `workflow_dispatch`-only since v2 took over production routing, and `issue-triage-v2.yml` is already on 22 — so nothing is broken in production today. It would have bitten whoever next reached for the documented v1 fallback, which is precisely when you least want the tooling to be the problem.

## What changed

- `.github/workflows/issue-triage.yml` — four `node-version: "20"` → `"22"`. The step name is plain `Set up Node.js`, so nothing else moves.
- `tests/triage-node-floor.bats` — new, 2 tests.
- `.github/workflows/tests.yml` — both triage workflows added to the push `paths`, so edits to them run the suite. (The `pull_request` trigger has no paths filter, so PR runs were already covered.)
- `CHANGELOG.md` — bullet under `[Unreleased]` → Fixed.

## On the test's shape

It asserts the floor **across both workflow files** rather than at the four repaired sites. A test pinned to the four line numbers would pass a fifth `Set up Node.js` step added later on 20 — and a defect that already appears four times in one file is exactly the kind that comes back. It covers `issue-triage-v2.yml` too, so v2 can't regress off 22 either.

The file list is explicit rather than derived from a grep for the install lines, because `ci.yml` installs the same tools and carries its own floor assertion in `ci-release-job.bats` (via #845). Deriving it would have coupled this PR's suite to that stack's merge order.

## Verification

Full suite 356/356 green. `actionlint` clean on every file this branch touches — the two findings it reports repo-wide (`ci.yml` `if-cond`, `update-flake-lock.yml` `shellcheck`) are pre-existing on `main` and untouched here. `shellcheck` clean on the new bats file.

Eight mutations, each applied and reverted **independently** against the committed fix, each redding exactly one test:

| Mutation | Reds |
|---|---|
| `issue-triage.yml` L111 → 20 | test 2 |
| `issue-triage.yml` L222 → 20 | test 2 |
| `issue-triage.yml` L298 → 20 | test 2 |
| `issue-triage.yml` L487 → 20 | test 2 |
| `issue-triage-v2.yml` L124 → 20 | test 2 |
| all four keys deleted | test 1 |
| all four keys commented out | test 1 |
| `issue-triage.yml` renamed | test 1 |

The commented-out case is what makes the comment-stripping in `node_majors` load-bearing rather than decoration, and the rename case is what stops the explicit file list from silently covering nothing.

Worth recording that my first pass at this was wrong: I ran the mutations with `git checkout --` restoring between them while the fix itself was still uncommitted, so the restore wiped the fix and every run after the first was measuring `main`, not the mutation. The numbers above are from the clean re-run against the committed baseline, with the live `node-version` state echoed before each run to prove exactly one site was mutated.

## Not in scope, flagged

`issue-triage.yml` installs `@anthropic-ai/claude-code` unpinned at all three sites. The `[Unreleased]` notes say v2's CLI install was version-pinned as part of the triage hardening; v1 never was. That's a supply-chain question rather than an engine-floor one, so I left it alone — happy to follow up if you want v1 brought in line.

---
85% AI / 15% Human
Claude: located the four sites and worked out the npm resolution behaviour against the live registry with `npm-pick-manifest` — including getting it wrong first, on a sibling PR, by asserting a bare spec resolves through the unfiltered `latest` tag, then correcting it to the range-spec mechanism documented above; wrote the workflow change, the bats file, the tests.yml trigger and the changelog entry; ran the eight mutations, caught that the first mutation run was invalidated by an uncommitted baseline, and re-ran them clean.
Human: directed that the triage workflows be fixed as well and that it ship separately from the #814/#845 stack; reviewed.


yannicklescure:
Correcting the npm mechanism in the description above — I had it wrong, and the correction changes how live this is. Pushed as 65c55ce (was e73bcd2 before this branch was rebased onto current main). No change to the workflow or to either test.

I claimed the mismatch leaves "a binary that exists and never runs" on Node 20. That is only true on npm <=9. A bare `npm install -g <name>` is a **range** spec (`*`), not a tag, and `npm-pick-manifest` >=9 filters ranges by `engines`; npm 10.x bundles pick-manifest 9, and `setup-node` with `node-version: "20"` ships npm 10.8.2. Verified across pick-manifest majors against the live registry — the filtering appears at 9:

```
pick-manifest 8.0.2:  node 20.19.2 -> 4.3.0   |   node 22.12.0 -> 4.3.0
pick-manifest 9.1.0:  node 20.19.2 -> 3.4.1   |   node 22.12.0 -> 4.3.0
```

So on today's runners a Node 20 triage job did not fail — it silently installed `@anthropic-ai/claude-code` 2.1.197 (`engines.node >=18.0.0`) instead of 2.1.270, and asar 3.4.1 instead of 4.3.0. The install-and-die shape belongs to npm <=9, which is how #839 was reported.

That makes the live defect the silent 73-release pin-back of the CLI the triage jobs actually run, with the npm <=9 break as a latent hazard behind it. Both were properties of the runner's npm rather than of this repo, which is what the floor fixes. The change and the tests are unaffected; only the justification moves.

---

yannicklescure:
Tightening one number in the mechanism above (6a38177, was c968de6 before this branch was rebased onto current main). No change to the workflow or to either test.

I said the engine filtering appears at `npm-pick-manifest` >=9, bundled by "npm 10.x". The boundary is **9.1.0**, and the original evidence sampled one version per major — `npm install npm-pick-manifest@9` resolves to 9.1.0, which is how you conclude "at 9" from a run that never tested 9.0.x.

Measured against the live asar packument at Node 20.19.2:

```
8.0.2  -> 4.3.0     9.0.0  -> 4.3.0     9.0.1  -> 4.3.0
9.1.0  -> 3.4.1     10.0.0 -> 3.4.1
```

And the npm side: 10.7.0 bundles 9.0.0, 10.8.0 bundles 9.0.1, **10.8.2 bundles 9.1.0**, 10.9.0 bundles 10.0.0. So npm 10.0.0 through 10.8.0 take the install-and-die branch, not the downgrade branch — "npm 10.x filters" was wrong.

The conclusion is unchanged and the fact is sharper: Node 20.19.2 bundles npm 10.8.2, so these workflows cleared the boundary by a single patch release. That is a better argument for pinning the runtime than the one I gave, not a weaker one — the behaviour was one npm patch bump away from flipping, in either direction, with nothing in the repo to say which.

---

yannicklescure:
@sabiut — when you have a moment, this one is the last of the #839 family still open, after #845, #846 and #848 landed.

Rebased onto `f4534cd` (current main, including your #848 approval) as `4bb9101`. The rebase re-conflicted only on `CHANGELOG.md`, where #848 had rewritten the release-job bullet under me; the resolution keeps main's corrected version of that bullet and adds the triage one, so the `>=9.1.0` wording you verified is untouched. Only four files differ from main: `issue-triage.yml`, `tests.yml`, `tests/triage-node-floor.bats`, `CHANGELOG.md`.

Same defect you already confirmed on the other two, at four more sites: `issue-triage.yml` set `node-version: "20"` and then installed `@electron/asar prettier` in `fetch-reference` and the claude-code CLI in the three Claude jobs. v1 is `workflow_dispatch`-only since v2 took over routing, so it was latent rather than firing, and v2 was already on 22.

Two things worth your eye specifically:

- **The test asserts across both workflow files rather than at the four repaired lines.** A line-pinned test would pass a fifth `Set up Node.js` step added later on 20, and this defect already appeared four times in one file. It covers `issue-triage-v2.yml` too, so v2 can't regress off 22 either.
- **The file list is explicit, not derived from a grep for the install lines.** `ci.yml` installs the same tools; deriving it would have coupled this suite to #845's merge order. Now that #845 has landed, folding `ci.yml` into the list would work — I left it out because `ci-release-job.bats` already asserts that one, but say the word if you'd rather have a single floor assertion covering all three.

Local: full suite 387/387, `actionlint` clean on both workflows, and eight mutations each red exactly one test (four triage sites, the v2 site, keys deleted, keys commented out, file renamed). CI on `4bb9101` is 7 pass / 2 pending as I write this.

---

yannicklescure:
@sabiut — took you up on the consolidation offer from my last comment, pushed as `035d70d`. Worth a re-read since it touches a file you already approved in #845.

**What changed.** `triage-node-floor.bats` is now `workflow-node-floor.bats` and lists all three floored workflows — `ci.yml`, `issue-triage.yml`, `issue-triage-v2.yml`. `ci-release-job.bats` loses its `the node version clears the asar engine floor` @test and gains a comment pointing at where the invariant moved.

**Why consolidate rather than keep both.** One invariant asserted in two files drifts, and this exact pair had already drifted: the npm mechanism was corrected in the release-job copy and missed in the `ci.yml` workflow comment, which is what #848 existed to repair. Keeping two copies of the assertion is the same hazard one level down.

**The replacement is stronger than what it removes.** The old @test scoped to the release job's own `setup-node` block. The consolidated one is file-wide, so a Node 20 `setup-node` added to any *other* job in `ci.yml` reds too — which the old one would have missed.

**Mutation results**, each applied and reverted independently against the committed state:

| Mutation | Reds |
|---|---|
| `ci.yml:157` -> 20 | test 2 (new coverage) |
| `ci.yml` key deleted | test 1 |
| `ci.yml` key commented out | test 1 |
| `issue-triage.yml` 111 / 222 / 298 / 487 -> 20 | test 2 each |
| `issue-triage-v2.yml:124` -> 20 | test 2 |
| workflow renamed out from under the list | test 1 |

Full suite 386/386, `actionlint` clean, `shellcheck` clean. `ci-release-job.bats` still 9/9 on its remaining structural tests.

One judgement call flagged rather than buried: removing an @test from a file you approved is the invasive half of this, and I did it without asking. If you'd rather `ci-release-job.bats` keep its own floor assertion for locality — the release job being the chokepoint it is — say so and I'll restore it; the duplication cost is real but so is having the release job's invariants readable in one file.

---

sabiut (approved):
Approving. Verified on 035d70d.

tests/workflow-node-floor.bats 2/2, ci-release-job.bats 9/9 with its floor test folded in here, full suite 544/0, actionlint clean on the touched workflows, clean test-merge onto main. All eight mutations red the test you name; I added the ci.yml site to the per-site sweep since 035d70d covers it, and it reds the same. A site written as `22.12` stays green.

Consolidating the ci.yml assertion into this suite is the right move. That pair had already drifted once.

One gap, not blocking: delete the `node-version` key at a single site, leaving the other three, and both tests stay green. Test 1 only asks for at least one version per file and test 2 only checks the versions present, so a setup-node step with no version at all goes unnoticed and takes the runner default. That is Node 22.23.2 on ubuntu-latest today, so nothing bites, but it is the "fifth step added on the wrong version" case this test exists for, reached by omission instead of by 20. Asserting setup-node step count == node-version key count per file would close it. Fine as a follow-up.

The unpinned claude-code CLI in v1 is a separate question, agreed; worth its own issue if v1 is going to stay as the documented fallback.

PR #848, opened 2026-09-14
docs(ci): land the release-job comment correction that 7cf9f72 missed

yannicklescure:
Both docs nits from @sabiut's review of #845, which was approved and merged before they could be folded in. No behaviour change: the workflow, the tests and the changelog are untouched.

## 1. The workflow comment still described only the npm <=9 outcome

`7cf9f72` was supposed to correct the npm mechanism in three places — the test comment, the changelog entry, and the workflow comment. Only the first two landed.

The miss was procedural, and worth naming because it is the same trap twice in one PR: the `ci.yml` edit was made but **not yet committed** when the mutation re-check in that same turn restored the file with `git checkout -- .github/workflows/ci.yml` between mutations. That reverted the uncommitted comment edit, and the subsequent `git add -A` committed only the two surviving halves. The lesson is the one already in the test-methodology learning about mutation checks — commit the fix *before* mutating, or the restore is indistinguishable from a revert.

The comment now carries both outcomes and why the pin exists:

```
# Node 22, not 20: asar declares engines.node >=22.12.0 from
# 4.0.0 onward, and the claude-code CLI installed further down
# declares >=22.0.0. Neither install fails loudly below that —
# what it does depends on the runner's npm. A bare install is a
# range spec, and npm-pick-manifest >=9 (npm 10.x) filters
# ranges by engines, so Node 20 quietly resolves asar 3.4.1 and
# CLI 2.1.197 instead. npm <=9 does not filter: it resolves
# 4.3.0, whose EBADENGINE is only a warning, leaving an asar
# that `command -v` finds and that dies on first use — ...
```

This matches what you measured independently: Debian 13's npm 9.2.0 on Node 20.19.2 resolving 4.3.0 and dying with `CANNOT RUN WITH NODE 20.19.2`, versus npm 10.9.9 on the same Node resolving 3.4.1 and running.

## 2. The learnings bullet said three tests, and it is two

You are right as the bullet was worded. Verified both variants against the suite:

```
--- comment names @electron/asar ONLY
    not ok 2 asar is installed by exactly one step
    not ok 6 node is set up before asar is installed, and is ungated

--- comment names BOTH packages
    not ok 2 asar is installed by exactly one step
    not ok 5 asar is not co-installed with release-notes-only tooling
    not ok 6 node is set up before asar is installed, and is ungated
```

My "three" came from the original annotation, which named `@anthropic-ai/claude-code` alongside asar and so tripped test 5 as well. The bullet now states both counts and which spelling produces which, since the distinction is the actionable part.

## Verification

`tests/ci-release-job.bats` 10/10 — which also confirms the rewritten comment does not itself trip the region-level greps it warns about. `actionlint` clean on the change; its one repo-wide finding (`if: false` on the manually-managed release-notes step) is pre-existing and untouched. Mutations re-run after committing: reverting the bump and deleting the `node-version` key each red test 7.

---
80% AI / 20% Human
Claude: diagnosed why the comment correction was lost, wrote the corrected comment, verified the two-versus-three test count empirically, and re-ran the mutations.
Human: reviewed; @sabiut reported both nits.


sabiut (approved):
Approving. Verified on 3da1884.

ci.yml changes are comment-only, the rest is wording. tests/ci-release-job.bats 10/10, full suite 522/0, actionlint only the pre-existing `if: false`, clean test-merge onto main. The comment keeps the package names bare, so the region greps stay quiet.

The 9.1.0 boundary is right. Measured under Node 20.19.2 against the live packument: npm-pick-manifest 9.0.0 still resolves asar 4.3.0, 9.1.0 resolves 3.4.1. And npm 10.8.2 pins `^9.1.0` where 10.8.1 pins `^9.0.1`, so "npm >=10.8.2" is the correct floor for the filtering behaviour.

The learnings bullet now matches what I measured, and the "commit before mutating" lesson is a good one to have in that file.

PR #851, opened 2026-09-14
fix(asar): fall back to the pinned install when a PATH asar is dead (#849)

### LINKED ISSUE #849: _resolve_asar hard-fails on a stale global asar instead of falling back to the pinned install
Follow-up to the non-blocking review note on #846.

## Current behaviour

`_resolve_asar` (`scripts/_common.sh`) prefers an `asar` on `PATH`, and if that binary fails the run-check it stops:

```
asar at '/usr/bin/asar' will not run:
CANNOT RUN WITH NODE 20.19.2
@electron/asar 4.x needs Node.js v22.12.0+; this host has v20.19.2. Use @electron/asar@3 on older Node, or a newer Node (#839).
```

The message is accurate and actionable. But on a host carrying a stale global 4.x — installed by an npm 9 that took `latest` — the fix is already sitting in the function: the `@electron/asar@3` install branch two lines below would produce a working tool. The resolver just never reaches it, because the `PATH` lookup succeeded and a failed run-check is terminal.

That host shape is not exotic. It is the #839 host (Debian 13 stable, Node 20.19.2, npm 9.2.0) plus one `npm i -g @electron/asar` at any point in the past.

## Proposed shape

```
PATH asar fails the run-check
  → warn, naming the refusal
  → fall through to the @3 install
  → if THAT also fails, report both:
      the PATH asar's refusal FIRST (the actual root cause)
      the install failure second
```

## The part that needs care

The both-fail path is the whole design risk, and it is why this is not a two-line change.

The entire point of #839 is that the operator learns it is a **Node** problem at the point of failure, rather than three stages downstream at a phantom `desktopName is missing/empty` WM_CLASS tripwire. A naive fallback regresses that for every host that is offline, air-gapped, or behind a proxy: the accurate `CANNOT RUN WITH NODE 20.19.2` gets replaced by `Failed to install @electron/asar@3`, which points at the network instead of at the Node floor — strictly worse diagnosis for the same underlying cause.

So the both-fail message must still carry the Node floor, and that needs its own mutation row (delete the floor from the both-fail path; a test should go red).

## Test churn

Seven tests in `tests/asar-resolve.bats` stub a broken `asar` on `PATH` and assert a refusal, with **no** `npm` stub:

```
resolve asar: an asar that exits non-zero is refused
resolve asar: a version-shaped reply on a non-zero exit is refused
resolve asar: a refusal that exits zero is refused
resolve asar: the same refusal on stderr at exit zero is refused
resolve asar: the refusal names the Node floor, not just the tool
artifact tests: a failed extract fails instead of passing
artifact tests: an unresolvable asar fails instead of skipping
```

Under a fallback, every one of those reaches a **real** `npm install` against the registry. They go red, and they go red by hitting the network from a unit test — which is its own problem regardless of the assertion outcome.

Each needs a `_stub_npm_install` variant that *fails*, so the refusal path stays reachable and stays asserted. Mechanical, but it is the bulk of the work.

`resolve asar: an asar on PATH wins over an install` is unaffected — it stubs a *working* `PATH` asar, which must still win without an install attempt.

## Acceptance

- [ ] A dead `PATH` asar warns and falls through to the `@3` install; the resolver succeeds
- [ ] A working `PATH` asar still wins with no install attempted (existing test stays green unchanged)
- [ ] Both-fail reports the `PATH` refusal first and still names the Node floor
- [ ] No test in `tests/asar-resolve.bats` touches the network
- [ ] Each new guard is mutation-checked — revert it, something goes red

---

yannicklescure:
Closes #849. Follow-up to the non-blocking review note @sabiut left on #846.

## The gap

#846 resolves an `asar` on `PATH` first and treats a failed run-check as terminal:

```
asar at '/usr/bin/asar' will not run:
CANNOT RUN WITH NODE 20.19.2
@electron/asar 4.x needs Node.js v22.12.0+; this host has v20.19.2. …
```

Accurate, but it leaves the machine broken when the `@electron/asar@3` install two lines below is the fix for that exact binary. That host is not exotic — it is the #839 host (Debian 13, Node 20.19.2) plus one `npm i -g @electron/asar` at any point in the past, since an npm 9 takes `latest` and `latest` is 4.x.

## The change

Fall through to the pinned install instead of stopping. This is safe *precisely because* it is gated on the run-check: a tool that cannot run is not a staged tool being overridden, it is a dead one. A **working** `PATH` asar still wins outright, with no install attempted — pinned by a test that also asserts npm was never invoked.

```
PATH asar runs        → use it, return (no network)
PATH asar does NOT run → notice, fall through to @<major> install
install fails          → PATH refusal FIRST, then install failure, then Node floor
installed asar dead    → PATH refusal (if any), then its refusal, then Node floor
```

### The part that needed the care

A naive fallback regresses the diagnosis it was built to protect. On an offline or proxied host it replaces the accurate `CANNOT RUN WITH NODE 20.19.2` with `Failed to install @electron/asar@3` — pointing at the network instead of at the Node floor, which is exactly what #839 exists to prevent.

So the `PATH` refusal is **held** rather than printed at the fallback point, and re-emitted to *lead* the final report — below npm's own error spew, where the operator is actually looking. Real output on a simulated offline host with a dead `PATH` asar:

```
asar at '/tmp/…/asar' will not run.
Falling back to @electron/asar@3...
npm error code ENOTFOUND
npm error syscall getaddrinfo
npm error errno ENOTFOUND
npm error request to https://registry.npmjs.org/@electron%2fasar failed
asar at '/tmp/…/asar' will not run:
CANNOT RUN WITH NODE 20.19.2
Failed to install @electron/asar@3 into /tmp/…/inst.
@electron/asar 4.x needs Node.js v22.12.0+; this host has no node. Use @electron/asar@3 on older Node, or a newer Node (#839).
```

Root cause leads the report; the Node floor survives every exit path.

> Worth noting an intermediate draft printed the refusal eagerly *and* in the report. Two mutations survived against it — deleting either re-print changed nothing, because the eager copy still satisfied the ordering assertion. That was the code telling me the re-print was redundant, so the eager print became a two-line notice and the detail moved to the report. Both mutations go red now.

### Refactor

The run-check is extracted to `_asar_probe` and the install to `_asar_install`, so the resolver reaches each from two places without duplicating them. Both probe arms are unchanged: exit code and stdout shape judged separately, merged streams reported, the match anchored at the start of the reply.

## Testing

`tests/asar-resolve.bats` → 30 tests. Full suite **552/552**. `shellcheck -x` clean over CI's selection; no added line over 80 columns.

`setup()` now installs a **failing** npm stub by default, so no unit test can reach the real registry. That is not decorative — removing it is mutation-checked with a sentinel-file probe, and four tests would otherwise hit `registry.npmjs.org`.

Seventeen mutations, **zero survivors**:

| Mutation | Red |
|---|---|
| Fallback removed (dead `PATH` asar terminal again) | 5 |
| `PATH` refusal dropped from install-failure report | 1 |
| `PATH` refusal dropped from probe-failure report | 1 |
| Node floor dropped from install-failure path | 2 |
| Node floor dropped from installed-probe-failure path | 2 |
| Working `PATH` asar no longer wins (always installs) | 5 |
| `$asar_exec` left on the dead `PATH` asar after fallback | 4 |
| Probe exit-code arm deleted | 1 |
| Probe stdout-shape arm deleted | 2 |
| Probe version regex unanchored | 1 |
| Probe `v?` dropped | 1 |
| Probe `-x` gate deleted | 2 |
| Probe report loses stderr | 3 |
| Empty-install-dir guard deleted | 1 |
| Missing-install-dir guard deleted | 1 |
| Install pin dropped | 3 |
| Network-proofing removed from `setup()` | 4 tests reach real npm |

The `v?` row is worth calling out: the shipped 3.4.1 wrapper answers `v3.4.1`, so dropping the optional `v` rejects the exact version we pin. That was untested before this PR.

Also verified against the real registry, not just stubs: a dead `PATH` asar self-heals to `v3.4.1` and returns 0 with `$asar_exec` pointing at the installed wrapper.

## Acceptance (from #849)

- [x] A dead `PATH` asar warns and falls through to the `@3` install; the resolver succeeds
- [x] A working `PATH` asar still wins with no install attempted
- [x] Both-fail reports the `PATH` refusal first and still names the Node floor
- [x] No test in `tests/asar-resolve.bats` touches the network
- [x] Each new guard is mutation-checked

---
90% AI / 10% Human
Claude: designed the fallback ordering and the held-refusal reporting, extracted `_asar_probe`/`_asar_install`, restructured the suite around the probe, network-proofed `setup()`, ran the 17-mutation sweep, caught and fixed the redundant-re-print the two surviving mutations exposed, verified both paths against the real registry and a simulated offline host.
Human: scoped the follow-up from the review note and set the mutation-check and no-network requirements.


sabiut (approved):
Approving. Verified on 1bb305b.

tests/asar-resolve.bats 30/30, full suite 552/0, CI-selection shellcheck clean. All seventeen mutation rows go red here, most at the counts in your table or wider. The suite also passes 30/30 with `npm_config_registry` pointed at a dead port, which is the cleanest statement that nothing in it reaches the network.

Real registry on the #839 host shape, Node 20.19.2 with npm 9.2.0: a dead 4.3.0 asar first on PATH self-heals to 3.4.1, rc 0, `$asar_exec` on the installed wrapper. A working 3.4.1 on PATH wins with no install directory created. Dead PATH asar plus an unreachable registry gives the report in the order you designed: the PATH refusal with `CANNOT RUN WITH NODE 20.19.2` leads, then the install failure, then the Node floor hint. `validate_app_contents` on that host now passes all eleven assertions where #846 reported a FAIL, and `tests/test-patch-stage.sh` against the pinned 1.52386.3 resources passes through the new resolver.

Holding the PATH refusal for the report rather than printing it eagerly was the right call. The eager print would have hidden the exact mutations you found.

CHANGELOG.md conflicts with main now that #852 landed under Unreleased. Both bullets should stay, so I'll resolve that in the merge commit rather than ask for another rebase.

PR #852, opened 2026-09-14
fix(audit): probe the whole bundle instead of one resolved chunk (#850)

### LINKED ISSUE #850: patch-necessity-audit.sh still assumes the pre-#820 single-main-chunk model and aborts on multi-chunk trees
Reported by @sabiut while reviewing #846. Pre-existing on `main`, unrelated to that PR — filing separately as agreed.

## Symptom

On a 1.49585.0 tree, `tools/patch-necessity-audit.sh` aborts:

```
index.js requires ... but it is missing
```

`index.js` in that bundle `require()`s more than one `index.chunk-*.js`, so the `grep -oP` at `tools/patch-necessity-audit.sh:103` captures a multi-line string, the `-f` test on it fails, and the tool exits 1.

## The trap: do not just take the first match

The two-line capture is the proximate cause, but `| head -1` would be a worse outcome than the abort.

The tool still assumes the 1.19367.0 model — a stub `index.js` behind exactly **one** content-hashed main chunk — and collapses the whole bundle to a single resolved `$index_js` that every subsequent anchor count greps:

```bash
# tools/patch-necessity-audit.sh:103
main_chunk=$(grep -oP 'require\("\./\Kindex\.chunk-[^"]+\.js(?="\))' "$index_js")
```

The build stopped believing that at #820. From `_resolve_anchor_file` (`scripts/patches/app-asar.sh:174`):

> There is deliberately no "the main JS file" any more. Pre-3.x bundles kept the whole main process in `.vite/build/index.js`; 1.19367.0 split it into a stub plus one content-hashed main chunk; 1.26832.0 dissolved that core entirely — `index.js` became a 190 KB file require()ing 83 chunks, across two chunk families (`index.chunk-*` and `index2.chunk-*`), with the tray anchor left behind in `index.js` itself. A single resolved path cannot serve every patch, so each anchor resolves its own file.

So on a multi-chunk tree the audit tool currently **aborts loudly**. Patched to take the first chunk, it would instead **silently report `not-needed`** for every anchor that lives in any other chunk — an anchor that is present and unmatched is indistinguishable from an anchor that upstream already fixed, which is precisely the `not-needed` verdict.

Those verdicts feed the patch-necessity matrix in `docs/learnings/official-deb-rebase-verification.md`, which is what decides whether a patch gets deleted. A false `not-needed` there argues for deleting a patch that is still load-bearing. Aborting is the safe failure; a first-match fix converts it into the unsafe one.

## Suggested direction

Resolve per-anchor the way the patch stage does, rather than resolving one main file up front — i.e. give the audit tool the equivalent of `_resolve_anchor_file`, searching all of `.vite/build/*.js` and asserting match counts, so a zero-match and a many-match are distinguishable from a genuine absence.

Worth noting `_resolve_anchor_file` already solved the adjacent traps the audit tool would hit on the same trees: `grep -rlPz` so an anchor's `\s*` can span a beautified newline, a quote class `[`"']` rather than a bare double quote (1.26832.0 swapped the minifier and re-emitted most string literals as backtick templates), and exactly-one-match assertions so a decoy chunk carrying the same literal cannot be picked silently.

## Scope note

Report-only tool; mutates nothing and is not on the build path, so this does not affect shipped packages. It blocks running a necessity audit against any bundle from 1.26832.0 onward, which is the current upstream shape.

---

yannicklescure:
Closes #850. Reported by @sabiut while reviewing #846.

## The abort

`tools/patch-necessity-audit.sh` followed the single `require()` out of `index.js` and grepped whatever it resolved to — the 1.19367.0 shape, a stub plus exactly one content-hashed chunk. 1.26832.0 dissolved that core into ~83 chunks across two families, so the `grep -oP` captured several filenames into one string and the `-f` test on it failed. On `main`, against a multi-chunk tree:

```
index.js requires index.chunk-aaa.js
index.chunk-bbb.js but it is missing
```

The multi-line capture is visible in the message itself.

## Why the obvious fix is the dangerous one

`| head -1` would be **worse than the abort**. Every anchor living in another chunk would read zero, and a zero here renders as `not-needed` — indistinguishable from an anchor upstream genuinely fixed. Those verdicts feed the deletion matrix in `docs/learnings/official-deb-rebase-verification.md`, so a false `not-needed` argues for deleting a patch that is still carrying load. Aborting is the safe direction; first-match converts it into the unsafe one.

## The change

Probe the whole `.vite/build` set rather than resolving one file. That sidesteps resolution entirely, because this tool only ever **counts** anchors and never edits one — so unlike the patch stage, it never has to decide which chunk an anchor belongs to. Same "there is deliberately no main JS file any more" move `_resolve_anchor_file` made in #820.

### Three latent bugs surfaced on the way

- **`grep -c` undercounts.** It answers "files with a match", so a minified chunk carrying twelve occurrences on its single line reported `1`. Invisible while the tool read one file; actively misleading across a set. Now `grep -oh | wc -l`.
- **The sort was locale-dependent.** A locale-aware collation ignores punctuation, so `index2.chunk-*` sorts *ahead* of `index.chunk-*` in `en_US` and *after* it in `C`. `extract()` answers with the first match in that order, so two operators got different output. Now `LC_ALL=C sort`.
- **A BRE alternation was about to become a literal pipe.** `probe_auto_updater` used plain `grep` with `apt_channel_pending\|apt channel not yet live`. Carried into `grep -P` unchanged, `\|` is a *literal* pipe and the probe silently stops matching — flipping a `not-needed` patch to `check`.

> The locale bug is the one I'd flag for review. I found it because two mutations survived the first sweep: narrowing `has()` and `extract()` to `bundle_js[0]` changed nothing. The cause wasn't a weak test — it was that a bare `sort` put the anchor-bearing chunk first by accident, so "first file" and "whole bundle" happened to agree on my fixtures. Pinning the locale fixed the tool and made both mutations catchable.

### A fifth, found by the new tests

`usage()` was `sed -n '2,20p'` — correct when written, silently stale once the header block grew to 21 lines. `--help` stopped after the `needed?` verdict and dropped `check`, which is the verdict most rows in the matrix actually carry. The range is now derived from the closing banner so it cannot drift again, with tests in both directions: the listing must reach `check`, and it must not run past the banner into the code. Pre-existing (lines 1–29 are untouched by the first commit), but it is a user-facing break in the exact tool this PR exists to fix.

### A fourth, found by running it

With the tool able to reach that row for the first time, the autoupdater probe turned out to be anchored on a string upstream retired. `apt_channel_pending` was renamed to `managed_by_package_manager` in the 1.18286.2 → 1.19367.0 window when the APT channel went live. The build's AU-1 tripwire (`_check_upstream_tripwires`) moved with it; this probe had not, so it reported a false `check` against every bundle since. Both names are accepted now and the audit agrees with the tripwire again.

## Testability

The run section is wrapped in `main` behind a `${BASH_SOURCE[0]} == "$0"` guard, so the probes can be unit-tested against fixture trees without fetching a 170 MB `.deb`. Argument parsing and tree staging moved into `_stage_tree`.

## Testing

New `tests/patch-necessity-audit.bats` — 25 tests across all three bundle shapes (single-file, stub-plus-one-chunk, multi-chunk across two families). Full suite **568/568**, `shellcheck -x` clean, no new line over 80 columns.

Fourteen mutations, **zero survivors**:

| Mutation | Red |
|---|---|
| **The unsafe fix**: resolve only the first chunk | 10 |
| Resolve only `index.js` (pre-1.19367.0 model) | 10 |
| `count` reverts to `grep -c` | 6 |
| `has()` narrowed to the first file | 1 |
| `extract()` narrowed to the first file | 1 |
| Updater probe back to a literal pipe | 3 |
| Updater anchor back to the old name only | 1 |
| Empty-bundle guard deleted | 1 |
| Preload probe widened off `mainView.js` | 1 |
| Run guard removed (`main` runs on source) | 23 |
| argv dropped from the run guard | 1 |
| Sort no longer locale-pinned | 3 |
| `usage()` range hardcoded stale again | 1 |
| `usage()` range runs past the banner | 1 |

**End to end against the pinned official 1.52386.3 `.deb`**, where `main` aborts:

```
Probing 173 JS file(s) under .vite/build/
frame-fix-wrapper            check        frame:!1 occurs 7x — confirm Linux reachability
tray.sh (race + icons)       not-needed   TrayIconLinux refs: 2, setImage refs: 2 (in-place native)
menuBarEnabled default       not-needed   defaults map ships menuBarEnabled:!0
quick-window.sh              needed?      var TW: ||hide() anchor present, no blur()
claude-code.sh               not-needed   getHostPlatform has native linux-x64/linux-arm64 branch
autoUpdater neutering        not-needed   updater disabled at source (managed_by_package_manager)
cowork.sh reroute            diverges     official coworkd refs: 4, hardcoded OVMF paths: 2
```

Real evidence in every row rather than an abort. Note these verdicts have not been reviewed against the current matrix in `official-deb-rebase-verification.md` — that is a separate exercise, and this PR only makes the tool able to produce them.

---
90% AI / 10% Human
Claude: diagnosed the chunk-model staleness, chose bundle-wide probing over per-anchor resolution and argued why, found the grep -c undercount, the locale-dependent sort (via two surviving mutations) and the BRE-to-PCRE pipe trap, spotted the retired updater anchor by running the fixed tool, made the script sourceable, wrote the suite and ran the 12-mutation sweep, verified against the real .deb.
Human: triaged the report from the #846 review and filed it with the first-match warning.


sabiut (approved):
Approving. Verified on 3e6f828.

tests/patch-necessity-audit.bats 25/25, full suite 568/0, CI-selection shellcheck clean, clean test-merge onto main. All fourteen mutations go red; I narrowed `count`/`has`/`extract` to the first file in one go and got 7 rather than your 10, so the table holds either way.

The claims check out here. en_US sorts `index2.chunk-*` ahead of `index.chunk-*`, C sorts it behind. The pinned 1.52386.3 bundle carries `managed_by_package_manager` once and `apt_channel_pending` nowhere, which is what the AU-1 tripwire in app-asar.sh already keys on, so the probe and the tripwire agree again. On main the help range stops at line 20 with the banner closing at 22, so `--help` was dropping the `check` verdict.

End to end with `--deb` against the real pinned 1.52386.3 official deb and against 1.49585.0: full matrix on both, where main aborts on both with the two-chunk capture. Verdicts match your table.

One nit, not blocking: the "a locale that reorders the chunks" test sets `LC_ALL=en_US.UTF-8` without checking the locale exists. Where it doesn't, sort falls back to C and the test passes vacuously. A `locale -a` check with a skip would make that visible. Fine as a follow-up.

Agree with the direction: the abort was the safe failure and probing the whole bundle keeps it safe. Since this tool only counts, it never needs to pick a chunk.

PR #856, opened 2026-09-16
Clean up stale rootfs.vhdx left behind by win32->unix manifest switch

### LINKED ISSUE #855: Stale rootfs.vhdx / rootfs.vhdx.zst never cleaned up after migration to native unix rootfs.img
## Summary

A `vm_bundles/claudevm.bundle/` created before the official app started
shipping a native `unix`-platform VM image keeps its old
`rootfs.vhdx` + `rootfs.vhdx.zst` pair forever after the bundle
migrates to the new `rootfs.img` format. On my machine this was
~10.9 GB of dead weight (8.7 GB + 2.2 GB) sitting alongside a fully
functional 10 GB `rootfs.img`.

## Root cause

Before the v3.0.0 official-deb rebase, this project's Linux Cowork
support worked by repurposing the upstream **win32** manifest's VHDX
file entries (there was no native Linux entry yet) and converting
`rootfs.vhdx` -> `rootfs.qcow2` on first use (see
`docs/learnings/cowork-vm-daemon.md`, and the old
`scripts/cowork-fallback/cowork-vm-service.js` KVM backend).

Anthropic's manifest has since grown a real `unix` platform key
(`files.unix.{arch}` entries, e.g. `publishedAt: 2026-08-13` in the
copy currently shipping) serving `rootfs.img` directly, downloaded via
`https://downloads.claude.ai/vms/linux/{arch}/{sha}` — confirmed by
`strings`-ing the installed `app.asar`. The official `coworkd` uses
this format natively; no `rootfs.qcow2` is ever produced from it.

Nothing deletes the superseded VHDX pair when a bundle created before
this switch migrates to the new format:

- Confirmed no `rootfs.qcow2` file exists anywhere in the bundle (the
  old KVM-backend conversion never ran against these particular
  files).
- The old daemon's own log (`~/.config/Claude/logs/cowork_vm_daemon.log`)
  has been silent since 2026-07-04, while `rootfs.img` /
  `sessiondata.img` continued to be touched into August/September —
  consistent with the *official* `coworkd` (not our bwrap fallback)
  now owning the bundle.
- `rootfs.vhdx` (birth 2026-03-02) predates `rootfs.img` (first seen
  2026-08-24) on my install by ~5 months.

## Suggested fix

A launcher-side cleanup (in the spirit of `backup_user_config` /
`cleanup_stale_cowork_socket` in `scripts/launcher-common.sh`, not an
asar patch — this is local state, not application logic): for each
`~/.config/Claude/vm_bundles/*/`, if `rootfs.img` is present (proof
the bundle already migrated to the native format), delete any
leftover `rootfs.vhdx` / `rootfs.vhdx.zst` in the same directory.
Fail-safe, never blocks launch.

I have a PR ready implementing this.

## Environment

- `claude-desktop-unofficial` 1.40609.1-3.2.3 (amd64, .deb)
- Bundle birth: 2026-03-02; `rootfs.img` first present: 2026-08-24

lizthegrey:
## Summary

- A `vm_bundles/*/` bundle created before Anthropic's manifest had a native `unix` platform key can carry a `rootfs.vhdx` + `rootfs.vhdx.zst` pair (~11 GB combined on the reporting machine) that's fully superseded once the bundle picks up the current `rootfs.img` format. Nothing — upstream or our own bwrap fallback — deletes the old pair once the new one exists.
- Adds `cleanup_stale_vm_bundle_images` to `scripts/launcher-common.sh`: for each bundle directory, if `rootfs.img` is present (proof the migration completed), remove any leftover `rootfs.vhdx` / `rootfs.vhdx.zst`. A bundle still on the old format alone (no `rootfs.img` yet) is left untouched. Fail-safe, never blocks launch.
- Skips a bundle while `COWORK_VM_BACKEND=kvm` is set and that backend hasn't produced its own `rootfs.qcow2` yet — the KVM fallback (`scripts/cowork-fallback/cowork-vm-service.js`) still sources `rootfs.vhdx` directly until then and never reads `rootfs.img`, so deleting vhdx out from under it would leave it with no rootfs it can use.
- Wired into all three launchers (deb/rpm/AppImage) next to the existing `cleanup_stale_*` calls.
- 8 bats tests; full `launcher-common.bats` suite 132/133 (one pre-existing unrelated failure, confirmed local-only both here and by @sabiut's review).
- `shellcheck -x` clean on all touched shell files.

See `docs/learnings/cowork-vm-daemon.md` for the full root-cause writeup (traced through the installed `app.asar`'s manifest strings and the old KVM-backend conversion code) and the CHANGELOG entry for more detail.

Fixes #855

## Test plan

- [x] `npx bats tests/launcher-common.bats --filter cleanup_stale_vm_bundle_images` — 8/8 pass
- [x] Full `launcher-common.bats` suite — 132/133, no new failures vs. `main`
- [x] `shellcheck -x` on `scripts/launcher-common.sh`, `scripts/packaging/{deb,rpm,appimage}.sh`
- [x] Verified against my own real `~/.config/Claude/vm_bundles/claudevm.bundle/`: had both formats present, deleting the vhdx pair reclaimed 10.9 GB with the bundle still functional afterward
- [x] Verified against the current pinned upstream `2.110.0` (not just the `1.40609.1` this was originally diagnosed on) — same `unix`/`rootfs.img` vs `win32`/`rootfs.vhdx` manifest split, same bundle constants

---
90% AI / 10% Human
Claude: traced the root cause end-to-end (app.asar manifest strings, old KVM-backend conversion code, daemon logs), wrote the implementation, tests, docs, and PR, and addressed review feedback.
Human: identified the disk-space symptom, confirmed the fix approach and scope, reviewed before submission.


lizthegrey:
Re: the version-drift warning — the original investigation was against my installed `1.40609.1-3.2.3`. Verified against the current pinned upstream `2.110.0` (downloaded and sha256-checked against `scripts/setup/official-deb.sh`'s pinned hash) before leaving this open:

```
publishedAt:"2026-09-11",files:{unix:{arm64:[{name:"rootfs.img",...
...win32:{arm64:[{name:"rootfs.vhdx",...
```

Same `unix`/`rootfs.img` vs `win32`/`rootfs.vhdx` manifest split, just a newer `publishedAt`. The bundle constants are also unchanged: `"vm_bundles"`, `"claudevm.bundle"`.

`cleanup_stale_vm_bundle_images` doesn't touch or reference the app.asar at all — it's a plain filesystem check in the launcher (`rootfs.img` present in a bundle dir → any leftover `rootfs.vhdx`/`rootfs.vhdx.zst` is safe to remove), so it isn't sensitive to line numbers or minified identifiers drifting between releases the way an asar patch would be. Confirmed the two file names it keys on are still exactly what 2.110.0 ships.

---
Written by Claude Sonnet 5 via [Claude Code](https://claude.ai/code)

sabiut (commented):
Thanks for this. The root-cause writeup in #855 made it easy to review, and re-checking against the pinned 2.110.0 before I got here was appreciated.

I tested `b4a31c8` locally:

- New tests 5/5, `launcher-common.bats` 130/130, full suite 425/425, `shellcheck -x` clean on all four touched scripts. The pre-existing failure you hit doesn't reproduce here or in CI, so that one is local to your machine.
- Mutation-checked the tests: dropping the `rootfs.img` guard, neutering the `rm`, skipping the `.zst`, and dropping the log line each turn a test red. Good coverage.
- Ran the function against a sandboxed `XDG_CONFIG_HOME`: bundle names with spaces, a stray file in `vm_bundles/`, a read-only bundle dir (fails quietly, exits 0, no false log line, cleans up on the next run), a symlinked `rootfs.vhdx` (only the link goes), and a second run (no-op). All behave.

One thing I'd like settled before merge:

**"No code path converts back to vhdx once img exists" isn't quite the whole picture.** The fallback daemon's `KvmBackend` still treats `rootfs.vhdx` as its source. `scripts/cowork-fallback/cowork-vm-service.js` (~L1571) converts `rootfs.vhdx` → `rootfs.qcow2` when the qcow2 is missing, and never reads `rootfs.img`. That backend is reachable via `COWORK_VM_BACKEND=kvm`, or auto-selected when bwrap isn't installed. So a bundle with `rootfs.img` + `rootfs.vhdx` and no `rootfs.qcow2` yet loses the only rootfs that path can use, and since the manifest no longer serves vhdx to Linux, it can't come back. I reproduced that in the sandbox.

It's narrow: anyone who has started the KVM backend even once already has the qcow2 and is unaffected. I don't want a blanket "skip when qcow2 is absent" guard either, since most pre-3.0 users were on bwrap, never converted, and are exactly who this cleanup is for. Two options:

1. Skip only on the explicit opt-in:
   ```bash
   # KVM fallback still sources rootfs.vhdx until it has
   # produced rootfs.qcow2 (cowork-vm-service.js).
   if [[ ${COWORK_VM_BACKEND,,} == kvm \
   	&& ! -f ${bundle}rootfs.qcow2 ]]; then
   	continue
   fi
   ```
   plus a test for it.
2. If the KVM fallback is effectively dead for post-3.0 bundles anyway, leave the code as is and just correct the comment/doc/CHANGELOG wording so it doesn't claim nothing reads the vhdx.

@RayCharlizard, that's your call, since it's your backend. Is the vhdx → qcow2 path still something we support?

Smaller things, none blocking:

- The header comment, the learnings section, and the CHANGELOG say the bwrap fallback uses `rootfs.img` natively. As far as I can tell the bwrap backend doesn't use a rootfs image at all. Worth tightening to "the official `coworkd` uses `rootfs.img`".
- `$bundle` comes out of the glob with a trailing `/`, so the paths are `…bundle//rootfs.img` and the log line prints the trailing slash. `bundle=${bundle%/}` at the top of the loop tidies both.
- Nothing pins the three launcher call sites: deleting the call from `deb.sh` leaves the suite green. The sibling `cleanup_stale_*` calls aren't pinned either, so this is a pre-existing gap and not yours to fix here. I'll track it separately.
- Placement of the new section at the top of `cowork-vm-daemon.md` is Ray's to weigh in on.

Happy to approve once the KVM question lands either way.

sabiut:
Filed the unpinned launcher call sites as #857. It predates this PR and does not block it.

lizthegrey:
Thanks for the thorough review and for reproducing the KVM edge case — appreciated.

Pushed `68df79c` addressing all of it:

- **KVM fallback guard**: went with option 1 (skip on the explicit `COWORK_VM_BACKEND=kvm` opt-in when `rootfs.qcow2` doesn't exist yet), matching the existing `[[ ${COWORK_VM_BACKEND:-} == 'bwrap' ]]` exact-match style already in the launcher rather than the `${,,}`-normalized form in my sketch. Didn't try to detect auto-selected KVM (no bwrap installed) from the launcher side — that's the app's own runtime decision, and guessing at it felt more likely to be wrong than useful. Still tagging @RayCharlizard in case the answer is actually "KVM/vhdx is fully retired, drop the guard" — happy to simplify if so, but figured shipping the safe version now beats leaving a known bug open while that gets settled.
- **Trailing slash**: fixed with `bundle=${bundle%/}` at the top of the loop, plus a regression test asserting no `bundle//` in the log.
- **Wording**: corrected the header comment, the learnings doc, and the CHANGELOG entry — it's the official `coworkd` that uses `rootfs.img`, not bwrap (bwrap doesn't use a rootfs image at all).
- **Tests**: added 3 — guard active without qcow2, guard inactive once qcow2 exists (normal cleanup applies), and the trailing-slash regression. Also re-ran the full suite: 132/133 (same pre-existing `COWORK_NODE_PATH` failure you confirmed is local-only), `shellcheck -x` clean on all four touched scripts.

Left the test-pinning gap and the doc-placement call as you scoped them — not touching those here.

---
Written by Claude Sonnet 5 via [Claude Code](https://claude.ai/code)

sabiut (commented):
Thanks for the quick turnaround. I verified `68df79c` locally:

* 8/8 new tests pass
* 428/428 full suite passes
* `shellcheck -x` is clean
* Dropping the guard, ignoring the qcow2 check, or matching `bwrap` instead of `kvm` each fails a test
* The wording fixes now read correctly

Two remaining points—the first is on me.

1. The KVM guard is dead code in 3.x

I should have checked whether this path was reachable before suggesting the guard. Sorry for the detour.

The in-ASAR gate requires `process.env.COWORK_VM_BACKEND === "bwrap"` (`scripts/patches/cowork-bwrap.sh`, `GATE`), and the launcher's `setup_cowork_bwrap_env` only runs on that same value. If `COWORK_VM_BACKEND=kvm`, the official `coworkd` runs and uses `rootfs.img`; `cowork-vm-service.js` never starts. When the service does start, the backend is already `bwrap`, so `detectBackend()` cannot reach `KvmBackend`.

`docs/configuration.md` already documents this behaviour: values other than `bwrap` are ignored by the official client. Therefore, there is no 3.x launch path in which the VHDX-to-QCOW2 conversion runs, so the new guard does not protect a reachable path.

I would prefer not to ship a guard and two tests that pin unreachable behaviour. The new learnings text also describes KVM as "the one holdout," which makes it sound like a live path.

My suggestion is to:

* Remove the KVM guard and its two tests.
* Keep the wording fixes.
* Rewrite the KVM sentence as historical context: the code still exists in `cowork-vm-service.js`, but it is unreachable behind the bwrap-only gate.

@RayCharlizard, please shout if I have missed a supported entry point that can still reach this path.

2. The trailing-slash test cannot currently fail

The test asserts that the log does not contain `bundle//`, but the log line prints `${bundle}: …`. With a trailing slash, the output is:

`claudevm.bundle/: rootfs.vhdx`

There is never a double slash for the current assertion to catch. I confirmed this by removing the trailing-slash stripping logic; all eight tests still passed.

Please change the test to match the exact expected line:

`claudevm.bundle: rootfs.vhdx`

That will make the stripping behaviour load-bearing.

With those two changes, this is approved from me. I'm also happy to push the fixup myself if you prefer—`maintainerCanModify` is enabled.

lizthegrey:
Maintainer push is on. Go ahead.

sabiut:
Pushed `8d63090` with both changes: the KVM guard and its two tests are gone, the comment/learnings/CHANGELOG state the KVM history as unreachable in 3.x, and the slash test now pins the exact `in <bundle>: rootfs.vhdx (#855)` log tail (removing the strip turns it red). Locally 6/6 + 426/426, `shellcheck -x` clean. Will approve and merge once CI is green.

sabiut (approved):
CI green on `8d63090` (15/15). Verified locally: 6/6 function tests, 426/426 full suite, `shellcheck -x` clean, removing the slash strip turns the new assertion red. Merging.

PR #858, opened 2026-09-18
docs(acknowledgments): consolidate repeat contributors into one bulleted entry

lizthegrey:
## Summary
- lizthegrey, sabiut, JustinJLeopard, diarized, colonelpanic8, and CochainComplex each had two separate entries in `ACKNOWLEDGMENTS.md` (most recently #855/#856 for lizthegrey, appended as a new bottom entry instead of joining the existing one).
- Merged each into a single entry with one sub-bullet per contribution, kept at the position of that contributor's first credited contribution, matching the existing multi-bullet style used elsewhere in the file.
- Added a rule to CLAUDE.md's Contributor Credits section: a returning contributor gets a new sub-bullet under their existing entry, not a second top-level entry.

## Test plan
- [x] Diffed `ACKNOWLEDGMENTS.md` to confirm no content was dropped, only reorganized.
- [x] Re-ran the duplicate-name check (`grep` for bolded names + `uniq -c`) to confirm no contributor appears more than once.
- N/A: docs-only change, no code paths affected.


sabiut:
Thanks for the sweep. Verified the consolidation mechanically: the multiset of every `#NNN` reference is identical to `main`, no name appears twice any more (six did), and the only wording changes are the sub-bullet rewrites plus the two→three correction, which is right.

Pushed `fc9890a` on top with two additions: the new rule mirrored into `AGENTS.md` (it carries the same Contributor Credits section and is documented as the vendor-neutral mirror of `CLAUDE.md`, so they would otherwise drift), and the yannicklescure entry split into sub-bullets — it was already a single entry so it escaped the dedupe, but as a seven-contribution `; and for` chain it is exactly the shape the rule describes converting. Same text, all eleven refs preserved. Will approve and merge once CI is green.

sabiut (approved):
CI green on `fc9890a` (15/15). Content verified mechanically against `main` (identical `#NNN` multiset, no duplicate names, `CLAUDE.md`/`AGENTS.md` credits sections identical). Merging.

PR #866, opened 2026-09-19
test(ci): require a node-version on every setup-node step

yannicklescure:
Closes the gap @sabiut flagged in review of #847:

> delete the `node-version` key at a single site, leaving the other three, and both tests stay green. Test 1 only asks for at least one version per file and test 2 only checks the versions present, so a setup-node step with no version at all goes unnoticed and takes the runner default.

Reproduced first, on merged `main`: deleting `issue-triage.yml:111` while the other three keys remain leaves `workflow-node-floor.bats` fully green. My original mutation only ever deleted *all four* keys at once, which is why it never surfaced.

## The fix

A third case asserting **`setup-node` step count == `node-version` key count, per file**. A count rather than another value, because the defect is a missing declaration, not a wrong one — and because the runner's default happens to clear the floor today (22.23.2 on `ubuntu-latest`), so any value-based assertion stays green through it. Whether the runtime is this repo's or the runner's is the thing the suite exists to pin.

It catches the inverse for free: a `setup-node` step removed with its `node-version` key left behind.

`setup_node_steps()` skips comment lines, matching `node_majors()`, so a commented-out step cannot inflate the count and a commented-out key cannot satisfy it.

## Mutations

Each applied and reverted independently against a clean tree:

| Mutation | Reds |
|---|---|
| `issue-triage.yml`: 1 of 4 keys **deleted** | **test 3** (was all-green) |
| `issue-triage.yml`: 1 of 4 keys **commented out** | **test 3** (was all-green) |
| `ci.yml`: its only key deleted | tests 1, 3 |
| `issue-triage-v2.yml`: its only key deleted | tests 1, 3 |
| `setup-node` step removed, key left behind | test 3 |
| `issue-triage.yml:111` -> 20 | test 2 |
| `ci.yml:157` -> 20 | test 2 |
| `issue-triage-v2.yml:124` -> 20 | test 2 |
| `issue-triage.yml`: all 4 keys deleted | tests 1, 3 |

Full suite 457/0, `shellcheck` clean, no line over 79 columns.

One process note, since it nearly put a wrong number in this description: my first pass at the matrix ran while an earlier backgrounded mutation run was still cycling over the same workflow files, and the interleaving produced a bogus `not ok 3` on a pure value mutation. Re-run serially once nothing else was in flight, that mutation reds only test 2, as it should. The table above is from the clean run.

## Not included

The unpinned `@anthropic-ai/claude-code` in `issue-triage.yml` — agreed it is a separate question, and worth its own issue if v1 stays the documented fallback. Happy to open that separately.

---
Generated with [Claude Code](https://claude.com/claude-code)
85% AI / 15% Human
Claude: reproduced the gap on merged main, wrote the count-equality case and its helper, ran the nine-mutation matrix, and caught that the first matrix run was invalidated by a concurrent background job.
Human: directed the fix; @sabiut found the gap in review.


yannicklescure:
@sabiut — the gap you flagged is closed here, thanks for catching it. Reproduced on merged `main` first: deleting `issue-triage.yml:111` while the other three keys remain leaves the suite fully green. My original mutation only ever deleted all four keys at once, which is exactly why it never surfaced.

The review pass raised one thing worth your input, since it is a false positive waiting to happen rather than anything wrong today:

**`node-version-file:` would red the new test.** `node_majors` matches `node-version:` literally, so a step configured with `node-version-file: .nvmrc` — a legitimate setup-node option — would look like a step with no version, and the count-equality case would fail it. Same for an expression value like `node-version: ${{ matrix.node }}`, where the regex wants literal digits after the optional quote.

The expression case I think is correct to red: the suite exists to pin a literal major, and routing that through a matrix is a decision someone should have to make deliberately. `node-version-file` is the one I would call a genuine false positive — it pins the version just as firmly, only in a different file. I left both alone rather than pre-emptively widening the matcher for a shape this repo does not use, on the grounds that a red which makes someone read this test is cheaper than a matcher that quietly accepts something it never verified. Happy to handle `node-version-file` explicitly if you would rather it be accepted, though it would mean reading the referenced file to get the major.

Also noted: the `*uses:*actions/setup-node*` glob is unanchored, so a `run:` body line containing that literal text outside a comment would inflate the count. Contrived, and `ci-release-job.bats` matches just as loosely with `index($0, pat)`, so I matched the existing idiom rather than tightening it and taking on fragility around `- uses:` versus `uses:`.

Mutation table and full verification are in the description. `ci-release-job.bats` still 9/9, which is the tell that the half of this invariant moved here in #847 has not regressed.

---

sabiut (approved):
Verified `ad9608f`: 3/3, full suite 457/457, shellcheck clean. Reproduced the gap on `main` exactly as described — one of `issue-triage.yml`'s four keys deleted leaves the old suite green — and the new case reds it, the inverse (step removed, key kept), and a commented-out key.

On `node-version-file:` — leave it red. Nothing here uses it, and a red that makes the next person read this test before adopting a different pinning style is the right default; accepting it would mean reading the referenced file to get the major, which is more than the case is worth today. The unanchored `uses:` match is consistent with `ci-release-job.bats`, fine. Merging.

PR #868, opened 2026-09-19
fix(ci): delete the v1 triage fallback instead of hardening it

### LINKED ISSUE #867: Triage v1 fallback missing the v2 hardening: unpinned CLI, --dangerously-skip-permissions, persisted credentials
The v1 triage pipeline (`.github/workflows/issue-triage.yml`) did not receive the hardening that landed for v2. It is `workflow_dispatch`-only since v2 took over production routing, so nothing fires today — but [`docs/issue-triage/README.md`](../blob/main/docs/issue-triage/README.md) keeps it as the documented manual fallback, and states that "rollback to v1-as-primary is a one-file change in either workflow." Whoever performs that rollback, probably while something else is already going wrong, inherits the pre-hardening posture silently.

Raised by @sabiut in review of #847:

> The unpinned claude-code CLI in v1 is a separate question, agreed; worth its own issue if v1 is going to stay as the documented fallback.

## What v1 is missing

Both jobs run with `ANTHROPIC_API_KEY` and a workflow-level `permissions:` of `issues: write`, `contents: read`, `actions: read`.

**1. The CLI install is unpinned — three sites.**

```
.github/workflows/issue-triage.yml:114   run: npm install -g @anthropic-ai/claude-code
.github/workflows/issue-triage.yml:301   run: npm install -g @anthropic-ai/claude-code
.github/workflows/issue-triage.yml:490   run: npm install -g @anthropic-ai/claude-code
```

v2 pins, with the rationale already written down at `issue-triage-v2.yml:126-130`:

```yaml
      - name: Install Claude CLI
        # Version-pinned like the SHA-pinned actions above: this CLI
        # runs with the Anthropic API key and repo tool access, so a
        # floating `latest` is a supply-chain hole. Bump deliberately.
        run: npm install -g @anthropic-ai/claude-code@2.1.214
```

**2. The investigator runs on `--dangerously-skip-permissions` — two sites.**

```
.github/workflows/issue-triage.yml:413   --dangerously-skip-permissions \
.github/workflows/issue-triage.yml:575   --dangerously-skip-permissions \
```

v2 replaced this with an explicit allowlist (`issue-triage-v2.yml:684`):

```
  --allowedTools "Read,Grep,Glob,Bash(strings:*),Bash(ls:*)" \
```

and says why at line 653: "an explicit allowlist, NOT `--dangerously-skip-permissions`."

This is the larger of the two. The investigator is the one LLM call that takes attacker-controlled issue text into its prompt *and* has tool access.

**3. The checkout does not set `persist-credentials: false`.**

v2 does, at `issue-triage-v2.yml:119`, with the reasoning in the comment above it: the investigator gets tool access to that workspace, so the job token should not be sitting in `.git/config` for it to read.

## Options

1. **Backport the three v2 changes to v1** — pin the CLI, swap `--dangerously-skip-permissions` for the same allowlist, set `persist-credentials: false`. Keeps the fallback real.
2. **Delete v1** and drop the fallback claim from `docs/issue-triage/README.md`. v2 has been production since it took over, and a fallback nobody has exercised is a liability of its own.
3. **Keep v1 but gate it** — a leading step that fails unless an explicit "I accept the v1 posture" input is set, so a rollback under pressure is a deliberate act rather than a default.

My read is (1) if the fallback is genuinely wanted, (2) otherwise — the worst outcome is the current one, where the fallback exists, is documented as a one-file rollback, and quietly differs from production on three security-relevant axes. But that is a maintainer call about how much v1 is worth.

## Scope note

@sabiut's review raised only the unpinned CLI. Items 2 and 3 turned up while confirming item 1 and share its root cause — v1 predates the hardening pass — so they are filed together rather than as three issues. Splitting them is fine if that suits the milestone better.

Not urgent: v1 has no automatic trigger, so none of this is live. It becomes live the moment someone takes the documented rollback.

---

yannicklescure:
Option 2 from #867, per @sabiut's decision there. This PR previously implemented option 1 (backport the v2 hardening into v1); it has been force-pushed to the deletion. The option-1 history is at `352601d`, still reachable from the comments below — [the reasoning for the switch is here](https://github.com/aaddrick/claude-desktop-debian/pull/868#issuecomment-5747402127), in short: hardening v1 removed the only thing it did that v2 doesn't, so it stopped being a fallback worth keeping.

Closes #867.

## What goes

`.github/workflows/issue-triage.yml` — 687 lines, `workflow_dispatch`-only since v2 took over production routing, and not run since.

## What had to move with it

| Site | Change |
|---|---|
| `docs/issue-triage/README.md` | lines 3, 769, 780 — the fallback claim and the layout-table row. §Rollout posture now states there is no fallback pipeline, and why. |
| `tests/workflow-node-floor.bats` | dropped from `FLOORED_WORKFLOWS`; the header comment said "all three". |
| `.github/workflows/tests.yml` | dropped from the `paths:` trigger; the comment said "all three". |
| `.github/workflows/issue-triage-v2.yml` | header comment no longer promises a `workflow_dispatch`-only fallback. Comment-only. |
| `.claude/skills/triage/SKILL.md` | `gh run list --workflow issue-triage.yml` pointed at the deleted file. |

**One of those is a pre-existing fix, flagging it rather than burying it:** the same skill ran `gh workflow run "Issue Triage"`, and no workflow carries that name — v1 was `Issue Triage (v1 — manual fallback only)`, v2 is `Issue Triage v2`. `/triage` has been dispatching nothing. Both lines now use `issue-triage-v2.yml` by filename, which survives a `name:` change.

## What stays: the hardening test, re-pointed at v2

`tests/triage-workflow-hardening.bats` was written for this PR's option-1 form and is the one piece worth keeping either way. Its `TRIAGE_WORKFLOWS` array now holds v2 alone, and it asserts file-wide that:

1. every `@anthropic-ai/claude-code` install is version-pinned, not a floating `latest`
2. no call runs on `--dangerously-skip-permissions`
3. `persist-credentials: false` appears once per `actions/checkout@`

Nothing asserted any of this before. All three were argued for in `issue-triage-v2.yml`'s own comments and held in place by nothing else — which is exactly how v1 came to differ from production on all three at once while reading as deliberate. A comment explaining a posture does not keep it.

Asserted file-wide rather than at the sites that established it, so a newly added job inherits the requirement instead of bypassing it. Comment lines are stripped before matching, since v2 deliberately names `--dangerously-skip-permissions` in a comment explaining what it uses instead.

**Mutation-checked**, per [`docs/learnings/test-methodology-and-coverage.md`](../blob/main/docs/learnings/test-methodology-and-coverage.md):

| Mutation on `issue-triage-v2.yml` | Result |
|---|---|
| `claude-code@2.1.214` → `claude-code` | reds case 2 only |
| allowlist → `--dangerously-skip-permissions` | reds case 3 only |
| one `persist-credentials: false` deleted | reds case 4 only |
| workflow renamed out from under the array | reds case 1; **cases 2–4 pass vacuously**, which is what case 1 exists to catch |

## Verification

- `bats tests/*.bats` — 461 tests, 0 failures
- `shellcheck -s bash` clean on both touched `.bats` files (CI's shellcheck job globs `*.sh`, so they are not otherwise covered)
- both edited workflows parse as YAML
- no `issue-triage.yml` reference remains outside CHANGELOG history
- 80-column rule holds on both `.bats` files with tabs expanded

`actionlint` was not run — not installed here, and I did not want to fetch a binary to run it. Both workflow edits are a comment block and one `paths:` entry, so CI's actionlint job is the check that matters.

## Not in scope

- **#869** pins `ci.yml:190` and touches no triage workflow. Independent of this PR either way.
- **v1's stale prompt file map** (`tray.sh`, `cowork.sh`, `titlebar.sh`, `frame-fix-wrapper.js`, `frame-fix-entry.js`, all gone in the v3.0.0 rebase) is deleted along with the file that carried it, so it needs no separate cleanup.
- **The classifier's phantom `build` label**, raised on #867, is a separate defect in `.claude/scripts/prompts/classify.txt` and unaffected by this.

---
75% AI / 25% Human
Claude: traced the reference sites, wrote the deletion and the doc rewrites, re-pointed and mutation-checked the hardening test, found the broken `/triage` dispatch name.
Human: took the option-2 call, directed the conversion, reviewed.


yannicklescure:
Review pass on this PR caught a real defect I introduced, fixed in `352601d`. Flagging it here because it would have degraded the fallback quietly rather than failing loudly.

**The allowlist contradicted the prompt.** v1's investigate prompt has a section v2 does not, telling the model to verify URLs with a network fetch for download and CDN issues. The new `--allowedTools "Read,Grep,Glob,Bash(strings:*),Bash(ls:*)"` denies that. So the step would have failed closed on every network check while still being instructed to perform one — producing wasted turns, or speculation presented as verified, which is exactly what the "Verify before stating" rule four lines above exists to prevent.

Not a hard failure, which is why it would have gone unnoticed: v1 still emits findings, just worse ones, on a recurring issue class in this repo.

Confirmed v2 has no equivalent instruction — none of the eight prompt bodies in `.claude/scripts/prompts/` carry it. So this is v1-specific text the backport left stranded, not a capability v2 kept.

**Fixed by replacing, not deleting**, following how `investigate.txt` handles an unavailable check (say so, mark it not-runnable, cap confidence at `low`). The section now states that network claims cannot be verified here, that neither a 404 nor a success may be assumed, and that any network-dependent conclusion is low confidence.

**Deliberately not fixed by allowlisting a network tool.** That reopens precisely the channel `issue-triage-v2.yml:653-657` names: an arbitrary outbound request from a step whose env holds `ANTHROPIC_API_KEY`, driven by a prompt carrying attacker-controlled issue text. If you would rather v1 keep real network verification, that is a posture decision and should be taken as one, not smuggled in through this PR.

Also in `352601d`: the install test now matches on the package name rather than `npm install -g <pkg>`, so `npm i -g` and `npm install --global` cannot pass vacuously by presenting no lines to judge — all three spellings are mutation-checked. Two lines rewrapped for the 80-column rule once tabs expand.

## Two things left alone, both out of scope

**`ci.yml:190` still installs the CLI unpinned**, for release-notes generation. Different risk profile — tag-triggered, no attacker-controlled input, no skip-permissions flag, no `claude -p` call — so it is not the hole this PR closes, and `TRIAGE_WORKFLOWS` excludes it on purpose. But it is the last floating install of that CLI in the repo. Happy to pin it separately if you want them uniform.

**v1's investigate prompt still names files that did not survive the v3.0.0 rebase** — `tray.sh`, `cowork.sh`, `titlebar.sh`, `frame-fix-wrapper.js`, `frame-fix-entry.js` around lines 373-386. Pre-existing and unrelated to hardening, so not touched here; worth its own cleanup if v1 is being kept.

---

yannicklescure:
Your option-2 decision in #867 (2026-09-19T21:00:03Z) landed before I opened this PR (23:20:40) and I didn't reference it. That was mine to catch; this went up as option 1 without engaging the call you'd already made.

Having reviewed my own work since: **I now think option 2 is right, and this PR is the evidence for it.**

## Option 1's value collapsed during this PR's review

The one argument for keeping v1 was that it isn't a strict subset of v2 — v1's investigate prompt instructs a network fetch to verify URLs on download and CDN issues, and v2 has no equivalent in any of the eight prompt bodies. Under `--dangerously-skip-permissions` that instruction was live, so v1 could do something v2 can't.

`352601d` removed it. Hardening the allowlist and keeping real network verification are mutually exclusive without reopening the outbound channel that `issue-triage-v2.yml:653-657` exists to close, so the instruction had to go.

Which means that as this PR now stands, v1 is v2's posture minus v2's pipeline, in a 31 KB file with no trigger. The differentiator was the reason to keep it, and hardening it is what destroyed the differentiator. That's a fairly complete argument that the fallback can't be both safe and worth having.

## And it would leave a second cost in place

`issue-triage.yml:355-406` hands the investigator a map of the patch suite that no longer describes this repo:

```
355:  Each subsystem owns its own file — tray.sh, cowork.sh, claude-code.sh, quick-window.sh,
356:  titlebar.sh, app-asar.sh — with shared helpers in scripts/patches/_common.sh.
362:  The wrapper files (frame-fix-wrapper.js, frame-fix-entry.js) intercept require('electron')
406:  - Whether the fix belongs in scripts/patches/*.sh (sed patch) or frame-fix-wrapper.js (Electron intercept)
```

`tray.sh`, `cowork.sh`, `titlebar.sh`, `frame-fix-wrapper.js` and `frame-fix-entry.js` all went in the v3.0.0 rebase. So a rollback under pressure gets an investigator pointed at five files that don't exist, and told that an Electron intercept is a valid place to put a fix — the exact thing [D-002](../blob/main/docs/decisions.md) rules out. Pre-existing, out of scope here, and it's a third divergence on top of the three this PR set out to close. That's #847's maintenance-cost point in concrete form: keeping v1 means keeping its prompts current too, and nobody has been.

(I can't corroborate "v1 has not run once" from the API — `cleanup-runs.yml` deletes non-release runs after 3 days, so a zero-run query proves only the last 3 days. Taking your 2026-04-21 date.)

## Proposal

Happy to convert this PR to the deletion rather than have you redo it — the four sites are already mapped:

- remove `.github/workflows/issue-triage.yml`
- drop it from `FLOORED_WORKFLOWS`, `tests/workflow-node-floor.bats:56` (and the comment at `:7`)
- rewrite the fallback wording in `docs/issue-triage/README.md` lines 3, 769, 780
- rewrite the CHANGELOG entry this PR added

**One piece worth salvaging either way:** `tests/triage-workflow-hardening.bats`, added here, asserts the pin / no-skip-permissions / `persist-credentials: false` posture across a `TRIAGE_WORKFLOWS` array. Re-point that array at `issue-triage-v2.yml` alone and it becomes a regression guard on v2 — which nothing currently is. Today v2's hardening is held in place only by the comments explaining it, so a future edit that drops the pin or the allowlist goes green.

Also unaffected: #869 pins `ci.yml:190` and touches no triage workflow, so it stands whichever way this goes.

Say the word and I'll force-push the conversion, or close this and leave it to you — no preference, and no objection if you'd rather do the deletion yourself.

---

sabiut (approved):
Verified `807f17e`: 4/4 hardening, 3/3 node-floor, full suite 461/461 on the branch and 462/462 on a local merge with current `main` (the CHANGELOG auto-merges cleanly now that #869 is in), shellcheck and actionlint clean, no reference to the deleted workflow left anywhere. The three v2 mutations red on the intended test each. Good catch on `/triage` dispatching a name no workflow has carried since April; by-filename is right. Merging.

Issue #864, opened 2026-09-19
`launcher.log` grows without bound during a session: Electron's stdout/stderr is redirected into it with no cap (32 GB reported)


sabiut:
Reported on #709 by @luohoa97 with a screenshot: `~/.cache/claude-desktop-debian/launcher.log` at **32.2 GB** (written since 10:11 that morning) and `launcher.log.1` at **20.6 GB**.

**Mechanism.** `run_electron_and_cleanup` (`scripts/launcher-common.sh:952`) launches the official binary as `"$@" >> "$log_file" 2>&1 &`, so Electron's entire stdout/stderr for the whole session lands in `launcher.log` with no cap. The 5 MiB rotation from #747 runs only at launch start, so once a running session starts looping on a Chromium/Electron message it writes until the disk is full; at the next launch, rotation just moves the multi-GB file to `.1` and starts a fresh one, which is exactly the pair in the screenshot. 32 GB in ~12 h is ~0.75 MB/s, i.e. thousands of lines per second from a repeating error. Which message is looping on the reporter's machine is unknown yet (asked on #709); it may be a separate bug, but the launcher must not let *any* loop fill the disk.

This is the #582 class (unbounded Electron output) landing in our own file instead of syslog. #582 was closed today as "our half done" on the strength of the launch-time rotation; that assessment was wrong for the in-session case.

**Fix shape.**
- Route Electron's output through a bounded filter instead of a bare redirect: `"$@" > >(filter >> "$log_file") 2>&1 &`. Process substitution keeps `$!` as Electron's pid, so signal forwarding and `wait` are unchanged.
- The filter (awk) collapses runs of identical lines into `<line>` + `[launcher] last line repeated N times`, and after a per-session cap (20 MiB) stops **writing** but keeps **reading**, so the child never sees SIGPIPE. One marker line records that the cap was hit, so `--doctor` and bug reports can see it.
- Tests: a chatty stub run through `run_electron_and_cleanup` shows the file stays under the cap and the child runs to completion; mutation checks on the cap, the dedupe, and the "never close the pipe" property.

**Workaround** until the fix ships: close the app and `rm ~/.cache/claude-desktop-debian/launcher.log*`. Nothing else depends on those files.

sabiut:
Fix is up in #865.

PR #869, opened 2026-09-20
fix(ci): pin the claude-code CLI the release job installs

yannicklescure:
Closes the last floating install of `@anthropic-ai/claude-code` in the repo. Raised as out-of-scope in review of #868; independent of that PR — this touches `ci.yml`, #868 touches `issue-triage.yml`.

The CLI runs with `ANTHROPIC_API_KEY`, so a `latest` tag is a supply-chain hole for the same reason it was one in the triage workflows. Pinned to `@2.1.214`, matching `issue-triage-v2.yml` and the v1 backport in #868.

## Worth knowing before reviewing

**The install is currently dead.** Its only consumer is the `compare-releases` step, which carries `if: false # disabled — release notes are managed manually`. Nothing in `ci.yml` invokes `claude` today — no `claude -p`, no bare `claude` call anywhere in the file. So this changes no behaviour at all.

**Pinning rather than deleting the step is a deliberate call.** Deleting the install would be the stronger security posture in the narrow sense — no install at all beats a pinned one. But that step reads as re-enablable (`# was: steps.prev.outcome == 'success' && ...` is preserved right above it), and removing the install would make switching it back on quietly depend on someone remembering to re-add a *pinned* one. Pinning keeps the capability and makes the re-enable safe by default.

If the intent is that `compare-releases` is gone for good rather than paused, then deleting both it and this install is the better change and I am happy to do that instead — but that is a call about the step's future, not about the pin.

## Test

`tests/ci-release-job.bats` gains the assertion, since the release job's invariants are that file's remit and it already parses this job. It matches on the package name rather than on `npm install -g <pkg>`, so `npm i -g` and `npm install --global` cannot pass vacuously by presenting no line to judge — the same widening applied to #868's equivalent test after review.

Mutations, each applied and reverted independently against the committed fix:

| Mutation | Reds |
|---|---|
| unpin the install | test 7 |
| unpinned via `npm i -g` | test 7 |
| unpinned via `npm install --global` | test 7 |
| the install's `run:` line deleted | test 7 |

Full suite 458/0, `shellcheck` clean. `actionlint` reports one finding on `ci.yml` — the `if: false` on `compare-releases` — which is pre-existing on `main` and is the very step discussed above; untouched here.

## Overlap

No conflict with #868 on workflow files. Both add a `[Unreleased]` → Fixed bullet, so whichever merges second will want a trivial CHANGELOG resolution.

---
80% AI / 20% Human
Claude: traced the CLI's only consumer to the disabled compare-releases step, pinned the install, added the assertion to the release-job suite, and ran the four-mutation matrix.
Human: asked for the pin; it surfaced from @sabiut's review of #868.


yannicklescure:
Review pass found a genuine vacuity hole in the assertion I added, fixed in `d4f5b93`. Worth spelling out because the test's own comment claimed to cover exactly the case it did not.

**The hole.** `step_blocks` selects a step block on its **raw** text, but `uncommented` strips comments only afterwards. So replacing the install with a comment that names the package:

```yaml
        # TODO: reinstate `npm install -g @anthropic-ai/claude-code`
```

kept `$block` non-empty — the `name:`, `if:` and `continue-on-error:` lines survive the strip — while the `grep -F` then yielded zero lines, the `while` loop never executed, and the test passed **green with no install in the workflow at all**.

I reproduced that before accepting the fix: pre-fix, that mutation is `ok 7`. Post-fix it reds. Counting the matches before the loop is what closes it.

This is the same raw-match/strip-after asymmetry that `docs/learnings/test-methodology-and-coverage.md:144` already documents from the other direction — there, a comment naming `@electron/asar` in this region *reds* tests 2 and 6. Same mechanism, opposite sign. Placing my `ci.yml` comment so it names neither specifier was deliberate for that reason; the vacuity direction I simply missed.

Full mutation set after the fix, each applied and reverted independently against the committed state:

| Mutation | Reds |
|---|---|
| run line → comment naming the package | test 7 — **was green before `d4f5b93`** |
| `claude-code@2.1.214` → `claude-code` | test 7 |
| rewritten as `npm i -g` | test 7 |
| rewritten as `npm install --global` | test 7 |
| the `run:` line deleted | test 7 |

Nothing else in the file moves under any of them, and `tests/workflow-node-floor.bats` — the other consumer of `ci.yml` — stays 3/3.

**Also in `c8295f3`, unrelated to the pin:** `CHANGELOG.md` had two `[Unreleased]` → Fixed bullets run together into one paragraph, `...pull/866))- \`launcher.log\` can no longer...`. The missing newline is on my own #866 bullet, and it made the #864 entry render as a continuation of it. One-character fix, kept as its own commit so it can be dropped if you would rather it went separately.

**Confirmed for the record**, since the PR description rests on it: there is no invocation of the `claude` binary anywhere in `ci.yml`. Every other occurrence of the string is shell parameter expansion on the version tag (`${tag#*+claude}`, `claudeVer`) or a release-asset URL. The `compare-releases` step at `:251` is the sole consumer and is `if: false`.

---

sabiut (approved):
Verified `c8295f3`: 10/10 + 3/3, full suite 458/458, shellcheck clean, actionlint only the pre-existing `if: false`. All five mutations red on test 7, including the comment-naming-the-package hole; good catch on that one, the same asymmetry the learnings doc records from the other direction. Pin over delete is the right call for now: whether `compare-releases` returns is a release-notes decision, not a security one, and a pinned dead install costs nothing while that's decided. The CHANGELOG newline was my #865 bullet's fault, thanks. Merging.

PR #870, opened 2026-09-21
fix(triage): feed the classifier the repo's live label set

yannicklescure:
Follow-up to #867, kept separate from #868 as asked. Takes option (1) from the thread — interpolate rather than hand-fix — plus the `security` criterion and a test over the names that have to stay hardcoded.

## The drift

`classify.txt:106-110` handed the model a list of "safe choices" that had drifted from the repo's labels:

| Suggested in the prompt | In the label set? |
|---|---|
| `format: aur` | no |
| `tray` | no |
| `nix` | no — the label is `format: nix`, which the same line also lists |
| `build` | no |

The sentence ended with "Never invent labels" while handing over four invented ones. `classify.json:27` repeated `'tray'` in its own examples. The phantom `build` label on #867's first pass was the prompt being obeyed, not classifier drift.

Nothing wrong was applied — Stage 9's `apply_if_valid` already filters `suggested_labels` through a cached `gh label list` — but that gate is also why it stayed invisible. Its only symptom is `::notice::Label 'build' not in repo label set` on a run nobody re-reads, so the failure recurs silently and per issue.

## The fix

The cache is written to `/tmp/triage/repo-labels.json` two steps before the classify call and, until now, read in exactly one place. It is now interpolated into the prompt as a `<repo_labels source="gh label list, trusted">` block, placed ahead of the reporter-controlled wrappers so the two provenances stay ordered. The vocabulary the classifier is offered and the vocabulary Stage 9 will accept are one list; a rename propagates on the next run. An empty cache fails the step rather than handing the model an empty vocabulary.

`classify.txt` and `classify.json` now point at that block instead of enumerating anything.

## The `security` criterion

Stated in the prompt's label section in roughly the words from the thread: exposure in what we ship, or in what runs with our credentials — the packaged artifacts and the paths they install into on one side, the workflows holding the API key and repo write access on the other. With the distinguishing case kept: a hardened default rolled back is exposure, an opt-in recipe a user has to choose for themselves is not. Also recorded in the triage README's label taxonomy, since the rule is more reusable than this issue.

## The test

`tests/triage-classify-vocabulary.bats`, 8 cases. The live one intersects every name the pipeline still states by hand against `gh label list` — `priority: critical` (the never-emit rule and Stage 9's demotion), `priority: medium` (Stage 9's default fill), `security`, the four class labels, the five `triage: *` states — so a rename reds the suite instead of degrading a prompt. It skips when `gh` is missing or unauthenticated locally, and hard-fails rather than skipping under `CI`. The job takes `issues: read` and a step-scoped `GH_TOKEN` for that one call.

The offline cases pin the interpolation itself: the cache step precedes the classify step, the block is built from `repo-labels.json`, the empty-set guard is present, the prompt carries the `security` criterion, and — the one that would have caught this in the first place — the prompt and schema enumerate no namespaced (`priority:` / `format:` / `platform:`) labels of their own beyond the pinned set. That is the shape the deleted list would grow back in.

## Verification

- Full suite 628/628, no skips (`gh` authenticated locally, so the live case ran and all 12 pinned names resolved).
- `actionlint` clean on both touched workflows; `shellcheck` unaffected (it globs `*.sh`).
- Mutation-checked, one mutation at a time: restoring the old hardcoded list reds the two vocabulary cases (both intended — it reintroduces the list *and* drops the block reference); dropping the `<repo_labels>` block, renaming the cache step, removing the empty-set guard, deleting the `security` paragraph, and letting a pinned name go stale each red exactly one case and nothing else.

One judgement call worth flagging: the pinned list lives in the test file rather than being derived from the workflow, so a *newly* hardcoded label name would not be covered until someone adds it there. Deriving it looked more brittle than the thing it protects — the names appear across a `case` map, a `for` loop and two prompt sentences — and the negative namespaced-label assertion covers the family that actually drifted. Happy to revisit if you'd rather have the derivation.

Refs #867

---
85% AI / 15% Human
Claude: traced the drift to the prompt's own suggested-labels line, wrote the interpolation, the prompt and schema edits, the test suite, docs and changelog, ran the suite and the mutation checks
Human: relayed the maintainer's direction on scope (option 1, security criterion, keep separate from #868) and reviewed


sabiut (approved):
Verified `0e645a7`: 8/8 with the live intersection running both here and in CI, 470/470 locally (628 in CI with the cowork suites), lint clean, all five mutations red on the intended cases. Interpolating names only, ahead of the untrusted wrappers, is the right shape, and the empty-set guard adds no failure mode the cache step didn't already have. Keeping the pinned list in the test is fine; the namespaced-label negative case covers the family that drifted. Merging once the build legs finish.

sabiut:
Re-checked `cb2f878` after the refactor: 8/8 and 470/470 here, `skip` still fires from inside `require_gh` with `gh` hidden and the `CI=1` path hard-fails with the reason, stale-pinned-name mutation still reds test 8. Approval stands.

Issue #857, opened 2026-09-17
No test pins the launcher's pre-launch call sites (`cleanup_stale_*`, `heal_autostart_entry`, `backup_user_config`)


sabiut:
The launcher's pre-launch hygiene functions are unit-tested in `tests/launcher-common.bats`, but nothing checks that the generated launchers actually call them. A refactor or a bad conflict resolution in `scripts/packaging/{deb,rpm,appimage}.sh` can drop a call and ship with a fully green run.

Found while reviewing #856, which adds a seventh call to the same block.

### Reproduce

On `main` (a6c3b2d), delete these six calls from all three launchers (18 lines in total):

```bash
sed -i -E '/^(cleanup_orphaned_cowork_daemon|cleanup_stale_desktop_helpers|cleanup_stale_lock|cleanup_stale_cowork_socket|backup_user_config|heal_autostart_entry .*)$/d' \
	scripts/packaging/{deb,rpm,appimage}.sh
bats tests/*.bats    # 420/420, nothing red
```

Every function's own tests keep passing, because they call the function directly and never go through a launcher.

### What is and isn't covered today

| Call | deb | rpm | AppImage |
|---|---|---|---|
| `cleanup_replaced_desktop_ui` | pinned via `run_replaced_ui_cleanup_test` (installed launcher) | pinned, same | no leg, by design (no-op under the FUSE mount) |
| `cleanup_orphaned_cowork_daemon` | not pinned | not pinned | not pinned |
| `cleanup_stale_desktop_helpers` | not pinned | not pinned | not pinned |
| `cleanup_stale_lock` | not pinned | not pinned | not pinned |
| `cleanup_stale_cowork_socket` | not pinned | not pinned | not pinned |
| `heal_autostart_entry` | not pinned | not pinned | not pinned |
| `backup_user_config` | not pinned | not pinned | not pinned |

`backup_user_config` is the one that matters most. It is the primary fix for the config-wipe class (see `docs/learnings/config-wipe-guard.md`), and losing the call is silent until a user needs a backup that was never taken.

### Proposed shape

A static bats test over the three packaging scripts is enough. The launcher body is a heredoc with each call on its own line, so there's no need to build a package:

- For each of `deb.sh`, `rpm.sh`, `appimage.sh`: assert that each expected call appears as an uncommented, line-anchored statement (`^name( |$)`), so a commented-out call or a mention in prose doesn't satisfy it.
[...]

PR #872, opened 2026-09-21
test(launcher): pin the pre-launch call sites in all three launchers

yannicklescure:
Refs #857

Every pre-launch hygiene helper is unit-tested against its **definition** in `scripts/launcher-common.sh`, and nothing asserted that the three generated launchers **call** it. On `main`, deleting the whole block from all three of `scripts/packaging/{deb,rpm,appimage}.sh` leaves `bats tests/*.bats` fully green. Nothing in `tests/` reads `scripts/packaging/*.sh` at all today. `backup_user_config` — the patch-zero-clean primary fix for the config-wipe class ([`docs/learnings/config-wipe-guard.md`](docs/learnings/config-wipe-guard.md)) — had no test at any level, and losing its call is silent until a user needs a backup that was never taken.

This adds `tests/launcher-prelaunch-calls.bats`, a structural suite over the three packaging scripts. No production change.

## What it asserts

An awk pass splits each packaging script on its `cat > … << EOF` / bare `EOF` heredoc boundaries and keeps the one region containing the launcher's own `setup_logging` — the `.desktop` entry, the AppStream metainfo and deb's `postinst`/`postrm` heredocs are discarded, and rpm's `<< SPECEOF` / deb's nested `<<'APPARMOR_EOF'` are neither openers nor terminators here. Rows come out in `grep -n` shape, so line numbers survive for the ordering and containment assertions.

From **one shared array** of the expected calls (not a per-format copy, so a call added to one launcher and missed in the others reds):

| Assertion | Test |
|---|---|
| the parse has the expected shape before anything reads it | `the launcher-heredoc parser sees the expected shape` |
| the anchor rejects comments, prefixed names, prose mentions, indentation | `the call anchor rejects comments, prefixes and prose mentions` |
| each call present exactly once, column-anchored, uncommented | `every pre-launch call appears exactly once in every launcher` |
| each call after `setup_logging \|\| exit 1`, before `run_electron_and_cleanup` | `every pre-launch call runs after setup_logging, before the exec` |
| no call site outside the launcher heredoc | `no pre-launch call site sits outside the launcher heredoc` |
| each format carries only its own `heal_autostart_entry` argument | `each launcher carries only its own heal_autostart_entry argument` |

The `heal_autostart_entry` argument is the one line of the block that differs per format, and the difference is a quoting one: deb and rpm open their launcher heredoc **unquoted**, so the source literally reads `heal_autostart_entry "/usr/bin/$package_name"` (expanded at build time), while the AppImage heredoc is **quoted** (`<< 'EOF'`), so `"${APPIMAGE:-}"` is literal in the source *and* in the shipped `AppRun`. The foreign-argument negative is derived from the same `heal_call()` the positive assertion uses, so a fourth format cannot leave a stale expectation behind.

## Two corrections to the issue as filed

1. **The array is eight, not the issue's six.** The issue's `sed` reproducer lists six; its table adds `cleanup_replaced_desktop_ui` (which deb and rpm pin through `run_replaced_ui_cleanup_test`, but AppImage has no leg for — so the static pin is its only coverage); and #856 landed `cleanup_stale_vm_bundle_images` in all three launchers after the issue was written. Pinning all eight is a superset of what the issue asks for and costs nothing.
2. **Scope is the three packaging scripts only.** `cleanup_after_electron_exit` at `scripts/launcher-common.sh:934` re-runs four of the same names *after* Electron exits, so a loose repo-wide grep for the names is satisfied with no launcher calling anything. There is no fourth launcher — `nix/claude-desktop.nix` and `nix/fhs.nix` contain none of these calls.

## Verification

- `bats tests/*.bats scripts/cowork-fallback/tests/*.bats` — **634/634, no skips** (628 on `main` + the 6 added).
- `shellcheck` — no `*.sh` touched. The `.bats` file (which CI does not lint) is clean apart from two expected SC2016 infos on the single-quoted `heal_autostart_entry` literals, where the `$` is the text being matched.
- Mutation check, every case below applied to a clean tree and reverted:

| Mutation | Result |
|---|---|
| delete `cleanup_stale_lock` from `rpm.sh` | reds *exactly once* + *ordering* |
| comment out `backup_user_config` in `deb.sh` | reds *exactly once* + *ordering* |
| move the whole block below the exec in `appimage.sh` | reds *ordering* only |
| AppImage `heal_autostart_entry` takes the deb argument | reds *heal argument* only |
| duplicate `cleanup_stale_cowork_socket` in `deb.sh` | reds *exactly once* only |
| move `cleanup_orphaned_cowork_daemon` outside the heredoc | reds *exactly once* + *ordering* + *containment* |
| rename `rpm.sh` | reds all five file-reading tests, parser self-check first |
| demote `cleanup_stale_vm_bundle_images` to a comment | reds *exactly once* + *ordering* |
| rename `cleanup_stale_desktop_helpers` → `…_helpers_extra` | reds *exactly once* + *ordering* |
| anchor loses its trailing-boundary group | reds the near-miss test |
| anchor loses its `^<lineno>:` prefix | reds the near-miss test |
| anchor allows leading indentation | reds the near-miss test |

One mutation deliberately does **not** red, and the file says so inline rather than implying coverage it doesn't have: removing the comment-row strip from `call_lines`. With the `^<lineno>:` anchor in place a `#` already occupies column 0 of any commented-out call, so the strip is redundant today. It stays as the second layer — it is what would make the anchor safe to relax, and `appimage.sh:91-92` carries a real comment directly above its `heal_autostart_entry` while `launcher-common.sh` names these helpers in prose.

## Verification level, stated honestly

This is the **cheap static tier**: it reads the packaging scripts, not a built package, so it pins that the launcher *source* makes the calls in the right region — not that the shipped launcher executes them against real stale state. That end-to-end leg (one fixture per helper: a stale lock, an orphaned daemon, a dead socket, and root for the artifact tests) is the gap #857 concedes in its own Out-of-scope section, and it stays open. Hence `Refs #857`, not `Fixes` — a sibling PR covers the `backup_user_config` definition-level half.

No `./build.sh` run; out of scope for a change that adds no production code.

---
85% AI / 15% Human
Claude: parsed the three launcher heredocs, wrote the suite and its near-miss fixture, ran the full bats suite and the twelve-case mutation harness, wrote the CHANGELOG entry and this description.
Human: identified the coverage gap while reviewing #856, filed the issue with the reproducer and the proposed shape, set the review bar at a recorded mutation check, and caught that the expected list had grown past the issue's six.


sabiut (approved):
Verified `cd6fc96`: 6/6, 476/476 on the branch and on a local merge with current `main`. The #857 reproducer reds tests 3/4/6; single deletion, comment-out, duplicate, and the foreign `heal_autostart_entry` argument each red what the table says. The eight-call correction and the packaging-only scope are both right. This is what #857 asked for; the end-to-end leg was out of scope there by design, so I'm closing #857 with this. Merging.

